PR #25, opened 2026-02-28
fix(presets): fix incorrect corepack command for pnpm

bjoerge:
## Description
Deploying a next js app with pnpm currently fails with the following error:
```
Usage Error: Unsupported package manager specification (pnpxm@latest)
```

I believe this line is wrong:
https://github.com/gotempsh/temps/blob/921ab979735eaa7fe11d44ec61243f7829c98c60/crates/temps-presets/src/nextjs.rs#L111

and should instead be 
```
RUN corepack prepare pnpm@latest --activate
```

## Type of change

- [ ] Documentation update

## Checklist

- [ ] I have written tests that cover the changes
- [ ] All new and existing tests pass (`cargo test --lib`)
- [ ] `cargo check --lib` passes with no warnings
- [ ] I have updated documentation where necessary


dviejokfs:
Thanks for the fix! Can you add it to the CHANGELOG.md?

PR #88, opened 2026-05-11
Add required extra Docker network attachments

adamlevineagent:
## Summary
- add TEMPS_DOCKER_EXTRA_NETWORKS for self-hosted installs that need app containers attached to external Docker dependency networks before startup
- add DeployRequest.extra_networks for future per-request attachments while keeping default behavior empty/backwards compatible
- fail deployment when a configured required network is missing or cannot be attached, avoiding green deploys with dependency DNS failures
- document the operator env var and update deployer sample struct construction

## Verification
- cargo fmt
- cargo check -p temps-config -p temps-deployer
- cargo test -p temps-deployer test_normalize_extra_networks_drops_empty_duplicates_primary_and_overlay -- --nocapture
- cargo test -p temps-deployer test_deploy_request_creation -- --nocapture

Note: cargo check -p temps-config -p temps-deployer -p temps-deployments still fails on Windows in existing docker_cleanup_service code using Docker::connect_with_unix_defaults and inferred cleanup result types; this patch did not touch that path.

PR #136, opened 2026-06-18
Treat deploy cancellations and 4xx health noise correctly

adamlevineagent:
## Summary
- keep workflow cancellation/build-cancellation errors as cancelled job results instead of failed job results
- make deployment-level cancellation detection tolerate `cancelled` and `canceled`
- count only server-side `5xx` responses in project health error-rate rollups for both Timescale and ClickHouse paths, so client 4xx noise no longer drives the `Down` badge

## Verification
- `cargo fmt --package temps-core --package temps-deployments --package temps-proxy`
- `cargo test -p temps-core test_cancellation_errors_are_not_failures`
- `cargo check -p temps-deployments -p temps-proxy` attempted; blocked locally because `openssl-sys` requires `perl`, which is not installed on this Windows host

Work item: playful/167/308
Requested reviewers: agent/playful/claude-deputy-2, mr-friction
Note: pushed from `adamlevineagent: kramer-temps-platform-167308` because this token has READ permission on `gotempsh/temps` upstream.

### DIFF
--- a/crates/temps-core/src/workflow_executor.rs
+++ b/crates/temps-core/src/workflow_executor.rs
@@ -81,6 +81,13 @@ impl WorkflowExecutor {
         Self { job_tracker }
     }
 
+    fn is_cancellation_error(error: &WorkflowError) -> bool {
+        matches!(
+            error,
+            WorkflowError::WorkflowCancelled | WorkflowError::BuildCancelled
+        )
+    }
+
     /// Execute a workflow with proper dependency resolution and parallel execution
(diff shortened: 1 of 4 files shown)


dviejokfs:
Thanks for the PR @adamlevineagent!

What about #88 ?

dviejokfs:
can this be merged? LGTM @adamlevineagent

dviejokfs:
Merging it, thanks for the fix @adamlevineagent !

PR #138, opened 2026-06-19
Add opt-in MariaDB-compatible services

bherila:
## Summary

Adds an opt-in MariaDB-compatible external service for hosted projects while leaving Temps' own internal PostgreSQL database unchanged.

This includes:
- a new MariaDB service implementation with create/import/update/delete, credentials, environment bindings, and backup/restore support via `mariadb-dump` with `mysqldump` fallback
- MariaDB query explorer support, including schema/table discovery and guarded SQL execution helpers
- frontend service presets/forms/icons so users can choose MariaDB from the existing service workflows
- conservative small-host defaults for MariaDB containers, plus standard/dedicated size profiles for larger hosts
- MDX docs covering MariaDB managed services, project linking, Laravel usage, backup coverage, and opt-in service startup behavior

## Notes

The service uses MariaDB images by default while exposing MySQL-compatible connection URLs and environment variables, since most frameworks and drivers connect to MariaDB through MySQL-compatible drivers.

MariaDB remains opt-in. Installing or starting Temps does not pull or run a MariaDB container; a container is created only when a user creates or imports a MariaDB external service. Linked projects receive separate per-project/per-environment databases inside that shared service.

## Validation

Ran on the remote Ubuntu build host as `bwh` from `~/proj/temps`:

- `CARGO_BUILD_JOBS=2 cargo check -p temps-providers --lib`
- `CARGO_BUILD_JOBS=2 cargo test -p temps-providers mariadb --lib`
- `cd web && bun install && TEMPS_VERSION=pr-validation bun run build`

The MariaDB-focused provider test run passed 13 tests.

For the docs-only follow-up commit:

- `git diff --check`


bherila:
FYI I also built + tested this on Amazon Linux

dviejokfs:
Thanks for the contribution @bherila!

Let me review it

bherila:
Thanks. I would really like to make temps my standard devprod stack so I appreciate everything you have built so far, and it's always nice to not have to maintain a disparate fork :)

Open to switching to mysql if that is what you prefer-- Mariadb is really just for nicer licensing.

I have a few more coming too.

dviejokfs:
A few points before this PR is considered ready:

- is there any way to have PITR backup for mariadb? This would be preferred over backing it up daily since for big databases the time it takes for backup completion will only increase
- MariaDB needs more tests in general, especially related to backup and restore (see postgres.rs)

bherila:
No WAL-G equivalent exists for MariaDB. The faithful analog of "base backup + WAL archive" is mariadb-backup (physical base) + binary-log archiving, with binlog replay for PITR. However, the mariadb:lts stock image already bundles mariadb-backup/mysqlbinlog, so we can use that.

bherila:
I'm still testing/debugging this fyi

dviejokfs:
> I'm still testing/debugging this fyi

sounds good, if you can try it in a real instance with a lot of data it would be way better to harden it more before it reaches the main repo

bherila:
I don't actually have such a site to test with real data. I only have a very small database with only a handful of daily users (<10 users) right now, hence why I've been using full snapshots for daily backups. So I'd welcome any volunteer assistance available to test with a larger database.

bherila:
Update from this pass:

- Brought the PR branch up to current `upstream/main` (latest head `ee2b84a7`) and mirrored the same fix/merge chain onto `release/bherila-amzn2023`.
- Fixed the two restore bugs found by the full-chain MariaDB PITR E2E:
  - `mariadb_physical` now pipes `mariadb-backup --stream=mbstream` directly into `gzip`, instead of gzipping the trailing cleanup command and uploading raw mbstream as `.gz`.
  - restore-to-new-service now reads archived binlogs from the source service prefix, not the restored service name, so PITR replay can find the source binlog manifest/objects.
- Also hardened the restore/E2E path while debugging CI: Docker exec polling now stops polling closed stdout/stderr streams and has bounded Docker API calls, PITR binlog upload/replay phases have explicit timeouts/phase logs, and the heavyweight MariaDB PITR Docker E2E now runs as its own GitHub Actions job with explicit setup, image-pull, build, test, diagnostics, and log-artifact steps.

Design summary for review:

- MariaDB support stays engine-per-source for this PR: `mariadb_dump` for logical backups and `mariadb_physical` plus binlogs for PITR. The broader Source/Sink split and Restic integration are intentionally deferred follow-ups rather than part of this change.
- PITR is the daily-base plus short-binlog-cadence model: a default daily physical base schedule is auto-provisioned for managed MariaDB services, and archived binlogs are intended to run on the 5-minute interval path.
- The out-of-box schedule uses the default S3 source. The Docker E2E backs that with MinIO as the S3-compatible target, not Garage.

Verification on GitHub Actions:

- Manually dispatched `Rust Tests` on the updated PR head `ee2b84a7`: https://github.com/bherila/temps/actions/runs/28056752699
- Dedicated job `MariaDB PITR E2E` passed: https://github.com/bherila/temps/actions/runs/28056752699/job/83060922709
- The E2E log shows the expected PITR assertion: `Restored row counts - A=5 B=4 C=0 (expected A=5 B=4 C=0)`, followed by `test result: ok. 1 passed`.

Local checks also passed on both updated branches:

- `cargo check -p temps-providers --lib`
- `cargo test -p temps-backup --features docker-tests mariadb_pitr_full_chain_e2e --no-run`

dviejokfs:
I saw the Postgres updates, let's leave other databases fixes outside of this PR.

Also, the decision to not wait until a backup is completed is on purpose.

Databases with small size have no issues, but when backing up 500-1000 GB we can't do the backups synchronous. Also, imagine if server is restarted, the state will be inconsistent. That's why for backups we need to use docker containers in detach mode and then have a way to check the state of a database

dviejokfs:
Please, keep this PR scoped to MariaDB @bherila

bherila:
I will update again, I don't have a way to reset back to draft status. However, the postgres fix is required 1st as it blocks the mariadb e2e test suite from working correctly.

bherila:
I'm going to open a separate PR for the postgres fix, and then after that's merged it will rebase out of this PR.

bherila:
Moved the Postgres major-upgrade hardening out into its own focused PR: https://github.com/gotempsh/temps/pull/151

That PR contains only the Postgres upgrade fix set:

- real backup/S3 fixture rows for the upgrade tests so the `pre_upgrade_backup_id` FK is valid;
- stronger Postgres readiness that waits for `psql SELECT 1` against the configured database, not just `pg_isready`/Docker health;
- clean live-volume handoff for snapshot/rollback so pg18 cannot reopen stale pg17 `PGDATA`;
- explicit copy-sidecar cleanup and wait logic before deleting volumes.

I also kicked off a fork `Rust Tests` workflow for that branch here as validation proof: https://github.com/bherila/temps/actions/runs/28072935670

Once #151 lands, I’ll rebase this MariaDB PR and drop those Postgres commits from #138 so this PR stays scoped to MariaDB/PITR.

bherila:
After #151 and #152 are reviewed and ideally merged, I will pick this PR back up again.

bherila:
Recreated as #161 — this PR could not be reopened after the branch was rebased onto current upstream `main` and scoped down to MariaDB/PITR only (the Postgres upgrade hardening was split out into #151). Please continue review on #161.

PR #146, opened 2026-06-19
feat(settings): add flat public hostname strategy

### LINKED ISSUE #139: Support flattened public hostname templates for proxied wildcard TLS providers
### Problem description

Temps currently creates environment/service hostnames that can become nested subdomains, for example:

```text
files.staging.example.com
api.preview-123.example.com
```

That works at the DNS layer, but it breaks a common Cloudflare setup where the zone uses proxied records and Universal SSL on a free/basic plan. Cloudflare's default edge certificate covers the apex and one wildcard label, such as:

```text
example.com
*.example.com
```

It does not cover deeper names like:

```text
files.staging.example.com
api.preview-123.example.com
```

For operators who need the Cloudflare proxy to remain enabled, this means Temps-generated hostnames can resolve but fail browser TLS at the Cloudflare edge unless the account buys a higher-tier certificate option or manually customizes hostname generation outside Temps.

### Proposed solution

Add a configurable external hostname strategy that can flatten all public hostnames into a single label under the configured base domain.

For example, with a base domain of `example.com`, Temps could generate:

```text
staging-web.example.com
staging-files.example.com
preview-123-web.example.com
preview-123-files.example.com
```

instead of:

```text
web.staging.example.com
files.staging.example.com
web.preview-123.example.com
files.preview-123.example.com
```

A possible configuration shape could be:

```toml
[dns]
base_domain = "example.com"
hostname_strategy = "flat"
external_hostname_template = "{environment}-{service}.{base_domain}"
preview_hostname_template = "{preview_slug}-{service}.{base_domain}"
```

Useful behavior:

- Keep generated public hostnames to one label beneath `base_domain` when `hostname_strategy = "flat"`.
- Preserve the existing nested hostname behavior as the default if desired.
- Support placeholders for app, environment, service, branch/preview slug, and short hash.
- Enforce DNS label length limits by truncating and appending a short hash when needed.
- Use the flattened hostname consistently for DNS records, certificate issuance, app URLs, object-storage endpoints, webhook URLs, and any browser-facing URLs.
- Optionally validate generated hostnames and warn when they exceed the configured domain depth budget.

This would let Temps work cleanly with Cloudflare proxied DNS and default one-level wildcard edge certificates without requiring users to disable the proxy.

### Alternatives considered

- Disable the Cloudflare proxy: not acceptable for deployments that rely on Cloudflare WAF, edge routing, caching, or access controls.
- Buy deeper-subdomain certificate coverage: works, but it makes basic Temps deployments depend on a paid Cloudflare certificate feature.
- Delegate a child zone such as `apps.example.com`: helps only if Temps still generates one-label names under that zone, such as `staging.apps.example.com`; it does not solve `files.staging.apps.example.com`.
- Manually override each generated hostname: possible for a small setup, but error-prone and hard to keep consistent across apps, previews, object storage, and service endpoints.

### Additional context

This is not Cloudflare-specific in the abstract. Any DNS/CDN/TLS provider with one-level wildcard certificate behavior can hit the same issue. The requested feature is a general-purpose public hostname templating/depth-control mechanism, with Cloudflare Universal SSL as a common example.

bherila:
### Motivation

Cloudflare Free/Pro Universal SSL and similar wildcard-TLS providers cover one label below the apex, not nested generated service hosts. This makes the default generated service URLs fail TLS without an actionable explanation.

Closes #139.

### Description

- Adds per-managed-domain generated_hostname_mode (standard or flat), with provider capability discovery.
- Adds preview and confirmed apply endpoints that show affected hostnames, DNS reconciliation, and provider zone-access status before changing routes.
- Adds opt-in sync_generated_records reconciliation to the configured edge_target; records are created or updated per generated hostname, and cleanup is restricted to records explicitly tagged as Temps-managed.
- Excludes production environments from generated-record sync and never changes custom domains, non-calculated deployment domains, project custom domains, or custom routes.
- Adds Cloudflare zone-access diagnostics and the per-domain UI workflow.
- Uses the generated OpenAPI SDK directly; no hand-written fetch shim remains.

### Migration and compatibility

Migration m20260622_000001_managed_domain_hostnames.rs adds the per-domain columns and carries an existing global flat choice onto current managed domains. The superseded global public_hostnames strategy/templates are removed.

### Scalability

This is control-plane-only work. Hostname enumeration and provider reconciliation run only on explicit preview/apply operations, are bounded by the installations environments and provider zone records, and perform no work on the proxy request path. Provider errors stop the apply with contextual diagnostics rather than creating unbounded retries or queues.

### Validation

- cargo fmt --all -- --check
- cargo check --lib
- cargo test -p temps-core public_hostname --lib (9 passed)
- cargo test -p temps-dns hostname_sync --lib (7 passed)
- web: bun run openapi-ts against an isolated server built from this branch
- web: bunx tsc --noEmit


dviejokfs:
I am against this type of domain structure since it disables the ability to have dynamic environments with automatic HTTPS. This feature will be for advanced users. Can you explain what's the workflow you envision to mitigate this?

bherila:
This will be useful for all users on Cloudflare's free or Pro plan.

Otherwise, if you use Cloudflare's reverse proxy, the free wildcard certificate covers `*.yourdomain.com` ... but nested subdomains like `a.b.yourdomain.com` will only work on the $200+ plans.

However, `a-b.yourdomain.com` is totally fine and works with Cloudflare without requiring any additional plan subscriptions. I'm currently testing this out on a real server.

dviejokfs:
But you could point the wildcard to your server without WAF and then add individual ones with the proxy enabled.

If you don't want to have any non protected CF pointing to Temps you could add individual ones to Temps and then enable the firewall for these ones.

I don't want to make it more complicated for users to decide what url strategy to use, though I think there's value to have a configurable template for the domain generation. 

The issue is consistency

If you have a lot of environments and change the url strategy you end up in a messy situation where some projects/environments end up with one URL strategy and others with another.

Or worse, you change the URL strategy and all the projects/environments are updated, breaking external applications

bherila:
I agree it shouldn't be a required choice- the default is probably ok. But given the current defaults, Cloudflare users will run into this unexpectedly, by default, when they add Cloudflare. I know it sure took me by surprise, and a bit of debugging to figure out what was actually going on there. That wasn't a very user-friendly experience either!

At the end of the day I really just don't want to reveal my server's ip at all, it kind of defeats the point of proxying some but not all subdomains...

bherila:
Would getting rid of the template string settings and just having the drop down menu be more user-friendly in your opinion? Another option is possibly using the Cloudflare integration itself to detect when a user's plan doesn't support nested subdomains and transparently apply the naming policy, but that could also lead to confusion and might require an API key with more permissions than would otherwise be needed. Yet another alternative could be to allow the user to select their Cloudflare plan when they configure that plugin. Many options...

bherila:
Reworked this based on @dviejokfs's feedback — the global URL-strategy setting is gone. Summary of what changed (pushed to this branch):

- **No global strategy or templates.** The Standard↔Flat choice now lives per managed domain (`dns_managed_domains.generated_hostname_mode`), configured where you set up the DNS provider — not in global platform settings. The free-form templates are removed; it's just a Standard/Flat toggle.
- **Default Standard, opt-in per domain.** There's no platform-wide switch that retroactively rewrites every environment's URLs. You opt a specific managed domain into Flat — which directly addresses the consistency concern (no more "change the strategy and everything flips at once").
- **Provider-gated, not a generic knob.** Providers advertise a `flat_hostnames` capability and the UI only surfaces/recommends Flat where it actually helps (Cloudflare Universal SSL). Other providers never see it, so it doesn't become a confusing choice for everyone.
- **Explicit preview → confirm before the breaking change.** Switching a domain to Flat shows an impact preview (which generated hostnames change, plus any DNS records that would be reconciled, plus the token's zone-access state) and requires confirmation — so you can't silently break external apps. Custom domains are never touched.
- **Optional per-hostname Cloudflare DNS sync (opt-in per domain).** With a configured edge target, applying reconciles one proxied record per generated hostname against the zone (A/AAAA for an IP, else CNAME), only ever removing records Temps generated.
- **Token scope flag.** Cloudflare 401/403 now surfaces as a permission error, and verifying a domain records whether the token can manage the zone, so a mis-scoped token is flagged in the UI rather than failing mysteriously.

@dviejokfs on pointing the wildcard at the origin without WAF and adding individual proxied records: that works, but it exposes the origin IP via the unproxied wildcard, which is the thing I'm trying to avoid. The per-domain opt-in plus per-hostname proxied sync keeps every public host behind the proxy without needing the higher-tier plan for nested certs — and because it's opt-in and default-off, it doesn't change anything for users who don't want it.

Tested: `cargo test -p temps-core public_hostname --lib` and `-p temps-dns --lib` (224 pass), and the full `temps` binary builds.

bherila:
## Follow-up: DNS-sync safety fix + CF-free test coverage

Pushed `d4775d24` addressing a safety issue found while testing the Cloudflare zone sync against a live zone, plus test automation so the reconcile logic can be validated without hitting the Cloudflare API.

### The bug
The first cut of the per-hostname sync computed deletes too aggressively: a dry-run against a real zone (`careowner.com`) would have **deleted the production record `app.careowner.com`** (and ~10 other untagged records) and created a `*-prod` record pointing at the staging edge IP. Root causes:
1. Deletion treated *any* untagged single-label record under the base as "stale".
2. Production environments weren't excluded from the generated host set.

### The fix (`crates/temps-dns/src/services/hostname_sync.rs`)
- **Deletes are now tag-scoped**: only records carrying the `temps:managed` tag are ever deleted. Untagged / user-created records are never touched. (Cloudflare records can't carry the tag through the provider abstraction, so on Cloudflare the sync is effectively create/update-only — the safest behavior.)
- **Production environments are excluded** (`is_production_env` matches `prod`/`production` on slug or name); deleted environments are skipped.
- Generated host enumeration now uses `environments.subdomain`.
- Records are upserted via `set_record` (create-or-update), comparing existing content against the configured edge target.

### Test automation (no Cloudflare needed)
Added `#[cfg(test)] mod tests` with an in-memory `MockProvider` implementing `DnsProvider`, covering 7 cases incl. the regressions above:
- `reconcile_never_deletes_untagged_records`
- `production_environments_are_excluded`
- `reconcile_deletes_only_tagged_stale_records`
- `reconcile_creates_missing_and_skips_correct`
- `reconcile_dry_run_writes_nothing`
- `desired_content_picks_record_type`, `relative_name_strips_suffix`

All 7 pass (`cargo test --lib -p temps-dns hostname_sync`).

### Live verification (against the real `careowner.com` Cloudflare zone)
- Dry-run preview now shows **only** the two staging creates (`careowner-staging`, `careowner-preview` → A `35.163.83.53`), **zero deletes**.
- Applied the sync; re-running the preview returns 0 changes (idempotent — records confirmed present in the live zone via `list_records`).
- `app.careowner.com` (production) unchanged before/after; `careowner-prod` (a pre-existing health-check record) correctly excluded and untouched.

PR #147, opened 2026-06-20
feat(tls): on-demand certs for the console host + sslip.io auto-enable

dviejokfs:
## Summary

Makes the **console host** eligible for on-demand HTTP-01 TLS (ADR-018) and **auto-enables** the feature for sslip.io installs. Together these let QuickStart (`deploy.sh --mode=quick`) get real Let's Encrypt HTTPS for the console *and* apps with no eager per-host provisioning step — which is what lets the `testing` setup mode be folded into `quick` (companion change in `temps-landing-new`'s `deploy.sh`).

## Why

`testing` mode existed only to HTTP-01 the console host on sslip.io. Two gaps blocked `quick` from absorbing it:

1. **The console host was never cert-eligible.** It's served as a fall-through, not a `CachedPeerTable` route, so the gate's check-2 (cert-eligible route) rejected it.
2. **`temps setup` never enabled on-demand TLS.** ADR-018 §6 specified sslip.io auto-enablement but it was never implemented, so `on_demand_tls.enabled` stayed `false` after a quick install.

## Changes

- **`temps-proxy`** — add `console_host: Option<String>` to `OnDemandCertConfig`. The gate exempts *exactly* the console host from the cert-eligible-route check, while check-1 (in-zone) still runs first. Every other host still requires a cert-eligible route, so the random-SNI flood defense is preserved.
- **`temps-cli` (serve wiring)** — derive `console.<zone>` and pass it into the config.
- **`temps-cli` (setup)** — implement ADR-018 §6: when the preview domain is `*.sslip.io` and **not** loopback, set `on_demand_tls.enabled=true` + the sslip.io zone. Local mode (`127.0.0.1.sslip.io`) is skipped (LE can't reach loopback), and the proxy's own startup loopback guard remains as a second layer.

## Tests

`cargo check --lib` clean; **33 tests pass** across the three changed modules. New tests:

- Gate: console host accepted **without** a route; exemption is **exact-host-only** (a different routeless in-zone host is still rejected); exemption **still requires in-zone** (a console host outside the zone fails check-1).
- Wiring: `console.<zone>` derivation (incl. trim/lowercase, empty-zone → `None`).
- Setup: sslip.io / loopback / custom-domain auto-enable decision.

## Notes / follow-ups

- **Companion PR:** `deploy.sh` in `kfsoftware/temps-landing-new` collapses `testing` → `quick` (kept as a deprecated alias), binds `:443` in quick mode, and fixes the completion copy (console + apps now HTTPS). Not in this repo.
- **Security review:** ADR-018 flags the issuance gate for `security-auditor` sign-off. The change only *widens* eligibility to one known stable host (the console), still gated by the in-zone check — worth a look before merge.
- **Docs:** `docs/adr/018-*.md` still describes the old eager `testing` model; a doc pass is a sensible follow-up.


crates/temps-cli/src/commands/setup.rs
```diff
@@ -1232,6 +1246,26 @@ fn extract_preview_domain(wildcard_domain: &str) -> String {
     wildcard_domain.trim_start_matches("*.").to_string()
 }
 
+/// True if `zone` is a sslip.io base domain (e.g. `1.2.3.4.sslip.io`). Used to
+/// gate ADR-018 §6 on-demand-TLS auto-enablement to sslip.io installs.
+fn is_sslip_zone(zone: &str) -> bool {
+    zone.trim()
+        .trim_end_matches('.')
+        .to_ascii_lowercase()
+        .ends_with("sslip.io")
```
bherila: This treats custom domains whose name merely ends with these characters (for example `*.mysslip.io` or `*.notsslip.io`) as sslip.io installs, so `update_app_settings` will auto-enable on-demand TLS and write that custom domain as the issuance zone. Since this helper is the gate for silently opting setup into lazy HTTP-01, require either the exact `sslip.io` domain or a `.sslip.io` label boundary.

CHANGELOG.md
```diff
@@ -17,6 +17,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -
 
 
+## [0.1.0-beta.36] - 2026-06-20
+
+### Fixed
+- **Traces page no longer crashes on traces with an empty service name.** A
```
bherila, line 128: This is still outstanding after the later PR updates. `CHANGELOG.md` still leaves `Unreleased` empty, and the only added entry is the `0.1.0-beta.36` traces fix. The PR now includes multiple operator-visible changes that need greppable changelog coverage under `Unreleased`: on-demand TLS setup/proxy behavior, `--letsencrypt-email` / `settings.letsencrypt.email`, TLS cert-serving behavior, per-environment deploy-on-push behavior, telemetry heartbeat/first-touch changes, full log replay, bounded teardown, and the storage/domain UI changes. `AGENTS.md` requires every user-visible change in the same commit, so this should be filled in before merge.

bherila (changes requested):
I found two issues that should be addressed before merge.

crates/temps-cli/src/commands/setup.rs
dviejokfs: you are right, thanks for the notice. It's changed

bherila: Rechecked at PR head `e958f626`: the suffix-boundary issue is addressed. `is_sslip_zone` was removed from setup, and auto-enablement now keys off non-empty/non-loopback preview domains instead of `ends_with("sslip.io")`. I tried to resolve this thread, but GitHub returned `FORBIDDEN` for `ResolveReviewThread` from this account, so this can be marked resolved by someone with maintainer permissions.

crates/temps-proxy/src/tls_cert_loader.rs
```diff
@@ -54,11 +54,19 @@ impl CertificateLoader {
...
+        // This prevents two failure modes:
+        //   1. A valid cert becoming unserveable while re-issuance is in progress
+        //      (request_challenge sets status → "challenge_requested", which is not
+        //      in CERT_SERVING_STATUSES, so the old cert would stop being served
+        //      during the ~5-second challenge window — triggering concurrent retries
+        //      that exhaust the LE "5 duplicate certs per 7 days" limit).
+        //   2. An "on_demand_failed" domain row still holding a valid unexpired cert
+        //      (e.g. when rate-limited) becoming completely unserveable instead of
+        //      continuing to serve the last-known-good cert.
         let domain_entity = domains::Entity::find()
```
bherila, line 68: Dropping the status filter here also makes expired or otherwise non-serving rows eligible. `load_certificate` does not check `expiration_time`, so a row with an expired cert/key and status like `expired`, `failed`, or `challenge_failed` will still be returned to rustls. That blocks the on-demand path too: `certificate_callback` only enqueues on-demand issuance when `load_certificate` returns `None`, so an expired stored cert can keep being served instead of triggering a fresh HTTP-01 cert. The previous filter avoided that. Please keep the intended `challenge_requested` / `on_demand_failed with still-valid cert` behavior, but also require the stored cert to be unexpired (or restrict to serving statuses plus the explicit in-progress states that preserve a still-valid cert).

crates/temps-cli/src/commands/setup.rs
```diff
@@ -183,6 +183,14 @@ pub struct SetupCommand {
     #[arg(long, default_value = "false", env = "LETSENCRYPT_STAGING")]
     pub letsencrypt_staging: bool,
 
+    /// Contact email for Let's Encrypt / ACME certificate issuance. Stored as
+    /// `letsencrypt.email` and used as the ACME account contact for all cert
```
bherila, line 187: This says `--letsencrypt-email` is used as the ACME contact for all certificate provisioning, but the setup DNS-01 flow still creates its immediate order with `admin_email`: the initial `request_challenge(&wildcard_domain, &admin_email)` call and the retry path both pass the admin email. So `temps setup --letsencrypt-email ops@example.com --admin-email admin@example.com` persists `ops@example.com` for later renewals/on-demand issuance but registers the setup-time certificate order with `admin@example.com`. That contradicts the new flag semantics and leaves the first cert on the old contact path.

web/src/components/project/settings/environments/EnvironmentConfigurationCard.tsx
```diff
@@ -239,6 +241,7 @@ export function EnvironmentConfigurationCard({
           ? parseInt(formData.exposed_port)
           : null,
         protected: formData.protected,
+        automatic_deploy: formData.automatic_deploy,
```
bherila, line 244: This posts a concrete boolean on every environment settings save, while the backend model changed `automatic_deploy` to a tri-state where `null`/missing means inherit from the project. The form initializes missing `automaticDeploy` as `true`, so opening an environment that currently inherits and saving an unrelated field writes an explicit `true` override. That can unexpectedly enable deploy-on-push even when the project default is disabled, and there is no way to clear the override back to inherit. This needs the same kind of tri-state control/payload handling used for `attack_mode`, or it should omit `automatic_deploy` unless the user actually sets an override.

bherila (changes requested):
I re-reviewed PR #147 at `e958f626`. The original sslip suffix concern is fixed, but the changelog thread is still outstanding and I found three additional regressions that should be addressed before merge.

bherila:
Thanks for fixing the other issues, btw! (I didn't have permission to resolve my own review thread, seems that needs maintainer status)

PR #149, opened 2026-06-22
fix(cli): hide --database-url password from process listings

dviejokfs:
## Summary

- **Root cause**: passing `--database-url` as a CLI flag makes the full PostgreSQL connection string (including the password) visible to any user on the server via `pgrep -af`, `ps aux`, and `/proc/self/cmdline`.
- **argv scrubbing**: `run()` in `lib.rs` now calls `scrub_sensitive_argv()` immediately after clap finishes parsing, overwriting the `--database-url` value in the kernel argv array in-place with `x` characters. Works on both Linux (via `environ` stack layout + `/proc/self/cmdline` for arg count) and macOS (via `_NSGetArgv`). Fails silently if sanity checks don't pass — no worse than before.
- **`hide_env_values = true`**: added to all 11 `--database-url` arg definitions so the value is also masked in `--help` output.
- **Backwards compatible**: the `--database-url` flag continues to work. `TEMPS_DATABASE_URL` env var remains the recommended approach (env vars are never visible in process listings).
- **CHANGELOG** and **docs** updated (env-variables reference + production checklist).

## Files changed

- `crates/temps-cli/src/lib.rs` — `scrub_sensitive_argv()` + `scrub_argv_raw()` helpers
- `crates/temps-cli/Cargo.toml` — added `libc` dependency
- `crates/temps-cli/src/commands/{serve/mod,proxy,migrate,setup,api_key,backup,backfill,doctor,domain,reset_password}.rs` — `hide_env_values = true`
- `CHANGELOG.md` — [Unreleased] Fixed entry

## Test plan

- [ ] `pgrep -af 'temps serve'` after `temps serve --database-url=postgres://user:secret@host/db` shows `xxx` in place of the password
- [ ] `temps serve --help` shows `[TEMPS_DATABASE_URL]` instead of the actual env value
- [ ] `TEMPS_DATABASE_URL=postgres://... temps serve` still works (env var path unaffected)
- [ ] `temps serve --database-url=postgres://...` still starts correctly (backwards compat)

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -40,6 +40,12 @@ jobs:
       ADMIN_EMAIL: admin@localho.st
       ADMIN_PASSWORD: E2eTestPass123!
       API_BASE: http://localhost:8081/api
+      # Disable anonymous product telemetry for CI. Without this, every e2e run
+      # boots a fresh instance and emits instance_started + project_created +
+      # deploy_attempted for the example apps, polluting the real telemetry DB
+      # with phantom "tried N deploys, never shipped" instances. (Opt-out values:
+      # 0/false/off/no/disabled — see temps-telemetry/src/service.rs.)
+      TEMPS_TELEMETRY: "0"
 
     steps:
       - name: Free up disk space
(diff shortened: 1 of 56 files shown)


dviejokfs:
Superseded: the `--database-url` process-listing fix has been moved onto PR #147 (`feat/on-demand-console-tls`) where it now lives alongside the related TLS work. Closing this standalone PR and deleting the branch.

PR #151, opened 2026-06-24
fix(providers): harden postgres major upgrades

bherila:
## Summary

Extracts the Postgres major-upgrade hardening that was originally discovered while validating MariaDB PITR E2E coverage in #138.

This PR keeps the scope to Postgres upgrade correctness only:

- seed real `s3_sources` and `backups` rows in the Postgres upgrade test fixture so `postgres_major_upgrades.pre_upgrade_backup_id` satisfies the real FK constraint;
- wait for a usable Postgres database with `psql SELECT 1`, not just Docker health / `pg_isready`, and create the configured DB if entrypoint initialization has not finished that race yet;
- require snapshot and rollback handoffs to remove stale live volumes before recreating them, so pg18 containers cannot reopen pg17-era `PGDATA` after a supposedly clean copy;
- make volume copy sidecars explicit, wait for their removal, and fail with context when Docker cannot release them.

## Why this is required

The pg17 to pg18 upgrade path depends on a clean data-volume handoff. During CI validation, the old behavior could leave the live volume or copy sidecar around and then start the target pg18 container against stale pg17 data. That produces misleading readiness states and can make an upgrade appear to advance while the target database is not actually usable.

The fixture update is also required because the tests now run against the real schema: the fake backup id used by the upgrade tests violated the `backups` FK once the orchestrator persisted it to `postgres_major_upgrades.pre_upgrade_backup_id`.

Together these changes make the Postgres upgrade tests exercise the real control-plane constraints and make the runtime upgrade path fail early instead of silently reusing incompatible data.

## Test Plan

This branch was rebased onto the latest `upstream/main` (clean rebase, no conflicts). Only `crates/temps-providers` is touched.

Local:

- `cargo check --lib -p temps-providers` — passes, no warnings
- `cargo test --lib -p temps-providers postgres` — `87 passed; 0 failed` (Docker-dependent bodies skip gracefully at runtime when Docker is unavailable, per repo policy)
- `cargo test -p temps-providers --features docker-tests --lib orchestrator_phase_new_container_completes_under_timeout -- --nocapture --test-threads=1` (passes locally; Docker-dependent body skipped because local Docker is unavailable)

CI (full suite, including Docker/E2E, runs against this branch's head `94be5343`):

- https://github.com/gotempsh/temps/actions/runs/28252083090 (Rust Tests)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -180,6 +180,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **multi-node:** Real container deploy over mTLS (gated live test)
 - **dev-cluster:** From-scratch multi-node e2e harness (mTLS join + deploy)
 - **providers:** Isolate lifecycle tests' ports and container names ([#171](https://github.com/gotempsh/temps/issues/171))
+- **Postgres lifecycle waits for usable databases**: managed Postgres
+  container startup now waits for real `psql SELECT 1` connectivity, creates
+  the configured database if the cluster is alive but it is absent, and
+  includes container logs on readiness failures. This prevents major-upgrade
+  restores from treating `pg_isready` as sufficient before the target database
+  exists.
+- **Postgres major upgrades require a clean live volume**: the upgrade
+  snapshot and rollback phases now remove the temporary copy sidecar before
+  deleting the old live data volume, fail if Docker still cannot remove that
+  volume, and wait for the final healthy Postgres container before creating
+  the target database, preventing Postgres 18+ containers from reopening stale
+  17-era PGDATA after a supposedly clean handoff.
+
+### Tests
+- **Postgres upgrade Docker tests use real backup FK rows**:
+  `postgres_upgrade` integration fixtures now insert a matching `backups` row
+  for the fake pre-upgrade backup id, keeping the existing rollback/upgrade
+  assertions valid after FK enforcement.
 
 ## [0.1.0-beta.39] - 2026-06-25
 
--- a/crates/temps-providers/src/externalsvc/exec_util.rs
+++ b/crates/temps-providers/src/externalsvc/exec_util.rs
@@ -257,7 +257,7 @@ mod tests {
     /// is the exact trigger — and the shape every backup's
     /// `pg_dumpall … > file` uses, which is how the hang shipped.
     #[cfg(feature = "docker-tests")]
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn run_exec_returns_for_a_command_with_no_stdout() {
         use bollard::models::ContainerCreateBody;
         use bollard::query_parameters::{
@@ -311,11 +311,19 @@ mod tests {
 
         // Exits 0 and writes only to files -> the exec's stdout/stderr are
         // empty, so the output stream EOFs immediately.
-        let outcome = tokio::time::timeout(
-            Duration::from_secs(60),
+        //
+        // Run `run_exec` on a SEPARATE task (multi_thread runtime) and time out
+        // the JoinHandle. A regression re-introduces a synchronous busy loop
+        // that never yields `Pending`, so a `timeout` wrapping the future
+        // directly could never fire — but as a spawned task the spin occupies
+        // one worker while the timeout fires on another, so the test FAILS
+        // instead of hanging CI.
+        let docker_task = docker.clone();
+        let name_task = name.clone();
+        let handle = tokio::spawn(async move {
             run_exec(
-                &docker,
-                &name,
+                &docker_task,
+                &name_task,
                 vec![
                     "sh".to_string(),
                     "-c".to_string(),
@@ -323,9 +331,10 @@ mod tests {
                 ],
                 None,
                 Duration::from_secs(30),
-            ),
-        )
-        .await;
+            )
+            .await
+        });
+        let outcome = tokio::time::timeout(Duration::from_secs(60), handle).await;
 
         let _ = docker
             .remove_container(
@@ -337,10 +346,11 @@ mod tests {
             )
             .await;
 
-        let result = outcome.expect(
+        let joined = outcome.expect(
             "run_exec must return for a no-stdout command instead of spinning \
              forever (regression: biased-select busy loop on a drained stream)",
         );
+        let result = joined.expect("run_exec task panicked");
         let exec = result.expect("run_exec should succeed");
         assert_eq!(exec.exit_code, 0, "expected exit 0, got {}", exec.exit_code);
     }
--- a/crates/temps-providers/src/externalsvc/exec_util.rs
+++ b/crates/temps-providers/src/externalsvc/exec_util.rs
@@ -102,7 +102,15 @@ pub async fn run_exec(
             tokio::select! {
                 biased;
 
-                chunk = output.next() => {
+                // `if !output_done` is load-bearing: once the stream hits
+                // EOF, `output.next()` returns `Ready(None)` immediately on
+                // every poll. Without this guard, the `biased` select would
+                // always take this branch and starve the `inspect_exec`
+                // branch below — spinning at 100% CPU and never detecting that
+                // the exec finished (the exact hang that wedged pre-upgrade
+                // backups until the 6h timeout). Disabling the branch once
+                // drained lets the poll branch observe `running == false`.
+                chunk = output.next(), if !output_done => {
                     match chunk {
                         Some(Ok(msg)) => {
                             captured.push_str(&msg.to_string());
@@ -238,5 +246,102 @@ mod tests {
         let result = tail(&big, 100);
         assert!(result.contains("earlier bytes elided"));
         assert!(result.ends_with(&"x".repeat(100)));
+    }
+
+    /// Regression guard for the `biased`-select busy loop. Once the exec's
+    /// output stream hits EOF, `output.next()` returns `Ready(None)` on every
+    /// poll; without the `if !output_done` guard the biased select always
+    /// takes that branch and starves the `inspect_exec` branch — spinning at
+    /// 100% CPU forever, never returning and never even hitting `run_exec`'s
+    /// own timeout. A command that writes only to a file (empty exec stdout)
+    /// is the exact trigger — and the shape every backup's
+    /// `pg_dumpall … > file` uses, which is how the hang shipped.
+    #[cfg(feature = "docker-tests")]
+    #[tokio::test]
+    async fn run_exec_returns_for_a_command_with_no_stdout() {
+        use bollard::models::ContainerCreateBody;
+        use bollard::query_parameters::{
+            CreateContainerOptionsBuilder, CreateImageOptions, RemoveContainerOptions,
+            StartContainerOptions,
+        };
+        use futures::TryStreamExt;
+
+        let docker = match Docker::connect_with_local_defaults() {
+            Ok(d) => d,
+            Err(e) => {
+                println!("Docker unavailable, skipping: {e}");
+                return;
+            }
+        };
+        if docker.ping().await.is_err() {
+            println!("Docker daemon not responding, skipping");
+            return;
+        }
+
+        let _ = docker
+            .create_image(
+                Some(CreateImageOptions {
+                    from_image: Some("busybox".to_string()),
+                    tag: Some("latest".to_string()),
+                    ..Default::default()
+                }),
+                None,
+                None,
+            )
+            .try_collect::<Vec<_>>()
+            .await;
+
+        let name = format!("temps_run_exec_nostdout_{}", rand::random::<u32>());
+        docker
+            .create_container(
+                Some(CreateContainerOptionsBuilder::new().name(&name).build()),
+                ContainerCreateBody {
+                    image: Some("busybox:latest".to_string()),
+                    entrypoint: Some(vec!["sleep".to_string()]),
+                    cmd: Some(vec!["120".to_string()]),
+                    ..Default::default()
+                },
+            )
+            .await
+            .expect("create sidecar");
+        docker
+            .start_container(&name, None::<StartContainerOptions>)
+            .await
+            .expect("start sidecar");
+
+        // Exits 0 and writes only to files -> the exec's stdout/stderr are
+        // empty, so the output stream EOFs immediately.
+        let outcome = tokio::time::timeout(
+            Duration::from_secs(60),
+            run_exec(
+                &docker,
+                &name,
+                vec![
+                    "sh".to_string(),
+                    "-c".to_string(),
+                    "echo hi > /tmp/out 2>/tmp/err".to_string(),
+                ],
+                None,
+                Duration::from_secs(30),
+            ),
+        )
+        .await;
+
+        let _ = docker
+            .remove_container(
+                &name,
+                Some(RemoveContainerOptions {
+                    force: true,
+                    ..Default::default()
+                }),
+            )
+            .await;
+
+        let result = outcome.expect(
+            "run_exec must return for a no-stdout command instead of spinning \
+             forever (regression: biased-select busy loop on a drained stream)",
+        );
+        let exec = result.expect("run_exec should succeed");
+        assert_eq!(exec.exit_code, 0, "expected exit 0, got {}", exec.exit_code);
     }
 }
--- a/crates/temps-providers/src/externalsvc/postgres.rs
+++ b/crates/temps-providers/src/externalsvc/postgres.rs
@@ -4924,6 +4924,31 @@ mod tests {
     // ── Database Name SQL Injection Prevention Tests ─────────────────
 
     #[test]
+    fn postgres_upgrade_rejected_roundtrips_through_anyhow() {
+        // `ExternalServiceManager::upgrade_service` downcasts the anyhow error
+        // back to this typed variant to map it to HTTP 400. Guard that the type
+        // stays downcastable (a 'static std::error::Error) through the
+        // `?`/`.into()` path — if it stopped being downcastable the response
+        // would silently degrade to 500.
+        let mv: anyhow::Error =
+            PostgresUpgradeRejected::MajorVersionChange { from: 17, to: 18 }.into();
+        assert!(matches!(
+            mv.downcast_ref::<PostgresUpgradeRejected>(),
+            Some(PostgresUpgradeRejected::MajorVersionChange { from: 17, to: 18 })
+        ));
+        assert!(mv
+            .to_string()
+            .contains("Cannot change PostgreSQL major version"));
+
+        let dg: anyhow::Error = PostgresUpgradeRejected::Downgrade { from: 18, to: 17 }.into();
+        assert!(matches!(
+            dg.downcast_ref::<PostgresUpgradeRejected>(),
+            Some(PostgresUpgradeRejected::Downgrade { .. })
+        ));
+        assert!(dg.to_string().contains("Cannot downgrade PostgreSQL"));
+    }
+
+    #[test]
     fn test_validate_pg_username_accepts_realistic_names() {
         assert!(validate_pg_username("postgres").is_ok());
         assert!(validate_pg_username("appuser").is_ok());
--- a/crates/temps-providers/src/externalsvc/postgres.rs
+++ b/crates/temps-providers/src/externalsvc/postgres.rs
@@ -32,15 +32,44 @@ use super::{
 
 /// POSIX-safe shell escaping: wraps value in single quotes, escaping any
 /// embedded single quotes. Safe for use in `sh -c` command strings.
-fn shell_escape(s: &str) -> String {
+///
+/// Shared across the Postgres provider files (`postgres_upgrade`,
+/// `postgres_lifecycle`) instead of each keeping its own copy -- the drift
+/// between duplicated copies of this exact kind of string-builder is what
+/// caused the healthcheck bug this crate's `postgres_healthcheck_cmd` fixes.
+pub(crate) fn shell_escape(s: &str) -> String {
     format!("'{}'", s.replace('\'', "'\\''"))
+}
+
+/// Rejections from same-version-only [`PostgresService::upgrade`]. The
+/// `ExternalService::upgrade` trait method returns `anyhow::Result<()>`
+/// across every provider, so this can't be a variant of a typed
+/// `Result<(), E>` without changing that trait's signature everywhere.
+/// Instead it's wrapped into the `anyhow::Error` via `?`/`.into()` so
+/// callers can `downcast_ref::<PostgresUpgradeRejected>()` for an exact,
+/// wording-independent match instead of matching on `e.to_string()`.
+#[derive(Debug, thiserror::Error)]
+pub enum PostgresUpgradeRejected {
+    #[error("Cannot downgrade PostgreSQL (from {from} to {to})")]
+    Downgrade { from: u32, to: u32 },
+
+    #[error(
+        "Cannot change PostgreSQL major version via this endpoint (from {from} to {to}). \
+         Use the major-version upgrade API (POST /external-services/{{id}}/upgrades) instead, \
+         which performs a real pg_dumpall/restore with a retained rollback volume and a \
+         mandatory pre-upgrade backup."
+    )]
+    MajorVersionChange { from: u32, to: u32 },
 }
 
 /// Builds the `pg_isready` healthcheck command pinned to the configured
 /// username/database. Without `-d`, `pg_isready` (via libpq) defaults the
 /// target database to the username, so any service where `database !=
 /// username` would log a FATAL on every healthcheck tick.
-fn postgres_healthcheck_cmd(username: &str, database: &str) -> String {
+///
+/// Shared with `postgres_lifecycle` so both container-creation sites build
+/// the exact same healthcheck command.
+pub(crate) fn postgres_healthcheck_cmd(username: &str, database: &str) -> String {
     format!(
         "pg_isready -U {} -d {}",
         shell_escape(username),
@@ -3088,11 +3117,11 @@ impl ExternalService for PostgresService {
         );
 
         if old_version > new_version {
-            return Err(anyhow::anyhow!(
-                "Cannot downgrade PostgreSQL (from {} to {})",
-                old_version,
-                new_version
-            ));
+            return Err(PostgresUpgradeRejected::Downgrade {
+                from: old_version,
+                to: new_version,
+            }
+            .into());
         }
 
         // Major version upgrades (e.g. 17 -> 18) are handled exclusively by
@@ -3110,14 +3139,11 @@ impl ExternalService for PostgresService {
         // nothing has been touched yet (no stop, no volume changes), so this
         // is a safe no-op rejection, not a partial failure.
         if old_version != new_version {
-            return Err(anyhow::anyhow!(
-                "Cannot change PostgreSQL major version via this endpoint (from {} to {}). \
-                 Use the major-version upgrade API (POST /external-services/{{id}}/upgrades) \
-                 instead, which performs a real pg_dumpall/restore with a retained rollback \
-                 volume and a mandatory pre-upgrade backup.",
-                old_version,
-                new_version
-            ));
+            return Err(PostgresUpgradeRejected::MajorVersionChange {
+                from: old_version,
+                to: new_version,
+            }
+            .into());
         }
 
         // Verify the new image can be pulled BEFORE stopping the old container
--- a/crates/temps-providers/src/externalsvc/postgres.rs
+++ b/crates/temps-providers/src/externalsvc/postgres.rs
@@ -36,6 +36,18 @@ fn shell_escape(s: &str) -> String {
     format!("'{}'", s.replace('\'', "'\\''"))
 }
 
+/// Builds the `pg_isready` healthcheck command pinned to the configured
+/// username/database. Without `-d`, `pg_isready` (via libpq) defaults the
+/// target database to the username, so any service where `database !=
+/// username` would log a FATAL on every healthcheck tick.
+fn postgres_healthcheck_cmd(username: &str, database: &str) -> String {
+    format!(
+        "pg_isready -U {} -d {}",
+        shell_escape(username),
+        shell_escape(database)
+    )
+}
+
 /// Input configuration for creating a PostgreSQL service
 /// This is what users provide when creating the service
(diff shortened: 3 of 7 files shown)


dviejokfs:
could you handle the conflicts @bherila ? I merged the other PR #152 and now CHANGELOG.md has conflicts

dviejokfs:
Thanks for this @bherila — the volume-handoff hardening is the right direction, and these are genuinely dangerous paths (a bad handoff = pg18 opening stale pg17 PGDATA), so the "fail early instead of silently reusing incompatible data" framing is exactly right. The `auto_remove: true → false` + explicit remove/poll/exit-code check on the copy sidecar, and `remove_volume_best_effort → remove_volume_or_fail`, both look correct. I confirmed `wait_for_container_healthy` is safe here because the container defines a `HEALTHCHECK` programmatically (`postgres_lifecycle.rs:242`, `pg_isready` with a 30s `start_period`), so `state.health` will be populated and it won't false-timeout.

A few items to address before merge (CHANGELOG conflict aside):

**1. Route the new exec through `exec_util::run_exec` instead of the new `exec_in_container` helper.**
`crates/temps-providers/src/externalsvc/exec_util.rs` already exists specifically to make `docker exec` resilient, and its module docs call out the exact failure modes the new helper is exposed to:
- drain-or-stall on stdout backpressure (the new code does drain — good), and
- `inspect_exec.running`/exit-code transiently returning `None`, which the naive `inspect_exec` read in `exec_in_container` doesn't tolerate.

The code being replaced was already a bespoke inline loop, so this isn't introducing the first duplication — but consolidating onto `run_exec` (it takes `cmd: Vec<String>` + `env` + a timeout and returns `ExecResult { exit_code, output }`) means a single fix benefits every engine, and avoids re-treading the bugs `exec_util` was written for.

**2. Copy-sidecar leak on the `wait_container` error path.**
Now that the sidecar is `auto_remove: false`, the explicit `remove_container` only runs *after* `wait_container().await?`. If `wait_container` itself errors, we return early and leak a stopped container. Impact is cosmetic (stopped container, not data; the next run force-removes same-named containers), but it'd be cleaner to remove the sidecar on that error path too.

**3. Real pg17→pg18 validation before merge (the important one).**
The behavioral changes — `remove_volume_or_fail` hard-fail, the copy exit-code check, the healthcheck+`psql SELECT 1`+ensure-database readiness — are all Docker-dependent integration paths. The `postgres-upgrades` check is green, but on a fork PR these heavy bodies may skip rather than actually exercise the new handoff logic, so the riskiest paths aren't proven by the CI run. Given this is data-loss-class code, it'd be good to see one real pg17→pg18 upgrade against a DB with actual data observed green before merging. I know that's the same constraint that came up on #138 (no large instance to test against) — if you can't, I (or another maintainer) can run it on a seeded instance; just flag it.

Test changes themselves look good — swapping the `Ok(42)` stub for real `s3_sources`/`backups` rows and asserting against the persisted id is a strengthening, not a loosening. 👍

bherila:
If you have time to run this on a site that actually uses pg, that's super helpful -- personally I only use mysql for my sites

bherila:
I'm going to rebase this and bring back my other PR either tonight or tomorrow

dviejokfs:
perfect, I'm with #158 currently

When I merge it I will be able to test this

bherila:

dviejokfs:
I'm looking at merging this before mariadb, don't upgrade to beta.40 @bherila there's a bug with the DNS I'm trying to solve

dviejokfs:
I'm going to add a guard to prevent start/stop from working if there's an upgrade in progress.

dviejokfs:
@bherila Pushed a follow-up commit (d2c86d74) directly to this branch — while reviewing the volume-handoff fixes here, found a related gap: nothing in `postgres.rs::start()` (or the `ExternalServiceManager::start_service` path that calls it) is aware of an in-progress `postgres_major_upgrades` row.

The orchestrator stops the container and deletes/recreates its data volume across phases (`snapshot` → `dump` → `new_container` → `restore` → `swap`). If a health-check-triggered restart, operator restart, or startup reconcile calls `start_service()` for that same service mid-upgrade, it recreates the container against an implicitly-created empty volume — Docker auto-creates named volumes on container start if missing. That container would look healthy and could accept writes that never make it into the volume the orchestrator is actually dumping/restoring from. Not data loss in the dump/restore logic itself (that part's provably safe — the container is stopped and the volume deleted before `pg_dumpall` runs), but a real gap if anything races the orchestrator during the (potentially long, with real data volumes) downtime window.

Added `ExternalServiceManager::ensure_no_active_upgrade()`, checked at the top of `start_service()` — short-circuits with a new `UpgradeInProgress` error if a `PENDING`/`RUNNING` row exists for the service. Terminal statuses (`failed`/`completed`/`cancelled`/`rolled_back`) don't block a start. `cargo check -p temps-providers --lib` and `cargo clippy -p temps-providers --lib --all-targets` both pass; pre-commit hooks (fmt/clippy/conventional-commit) passed on the commit.

Scoped narrowly to just the guard — didn't touch the volume/readiness logic already in this PR. Happy to split this into its own PR if you'd rather keep #151 scoped to the original two fixes; pushed here since maintainer edits were enabled and it directly follows from the same review pass.

dviejokfs:
## Verified the upgrade end-to-end with real data

Ran a full v17 → v18 upgrade locally against a seeded Postgres service (3,000,000 rows / ~400MB table) using this branch's binary. Result: **fully successful**, and the volume-handoff hardening in this PR works as intended.

- Phase log: `snapshot(3s) → dump(<1s) → new_container(6s) → restore(5s) → swap(instant) → analyze(instant)` — clean single pass, no retries within these phases.
- Row count and content checksum (`md5(string_agg(payload, '' order by id))`) identical before and after: `61d04682e00df2ec7098eaeaa60f3dd7`, 3,000,000 rows.
- New container correctly on `postgres:18-bookworm` (verified via `SELECT version()` → PostgreSQL 18.4), writable post-upgrade (inserted and read back a new row).
- Rollback volume retained with correct 7-day expiry, no orphaned copy/dump sidecar containers left behind.

## Second follow-up commit: `511299c0`

While doing this deeper verification (inspecting the fresh pg18 container's logs, not just querying data), found the new container was logging `FATAL: database "appuser" does not exist` roughly once per second, indefinitely.

Root cause: the Docker healthcheck (`pg_isready -U {username}`) has no `-d` flag, so libpq defaults the target database to the username. For any service where `database != username` (the common case), every healthcheck tick opens a connection to a nonexistent database and postmaster logs a FATAL for it. `pg_isready` still reports the container `healthy` regardless (the server did respond), so this is completely invisible in the UI — it just permanently spams the container's log volume for the service's entire lifetime.

Found the same bug in **two** places:
- `postgres_lifecycle.rs`'s `create_and_start()` (used by this PR's upgrade orchestrator lifecycle) — parameterized on username but missing `-d`.
- `postgres.rs`'s `create_container()` (the normal external-service creation path) — actually worse, hardcoded to `-U postgres` regardless of the configured username.

Fixed both to pass `-U` and `-d` explicitly (shell-escaped via each file's existing `shell_escape` helper). Verified locally: before the fix, a fresh `username=appuser/database=appdb` service logged a FATAL ~once/sec; after, zero FATALs over the same window, Docker still reports `healthy`.

Not related to the exec-hang bug flagged in the earlier comment (that one's still open, in `exec_util.rs`'s unbounded output-drain loop, hit during a generic pg_dump-sidecar backup, unrelated to this PR's diff) — routed around it during testing rather than fix inline here, since it's a different subsystem.

bherila:
## Review + rebase update

Rebased this branch onto latest `upstream/main` (clean rebase, no conflicts) — head is now `ea836e71`. Also rebased #161 onto the same base (one small migration-ordering conflict in `temps-migrations/src/migration/mod.rs`, resolved by inserting `m20260623_000001_add_external_services_default_backup_provisioned` chronologically; both branches now `cargo check --lib` clean with no warnings).

Found a few issues while reviewing the volume-handoff hardening, ranked by severity:

**1. `copy_volume`'s sidecar still has `auto_remove: Some(true)`, which races the new explicit `remove_container` call.**
`postgres_upgrade.rs`'s `copy_volume()` container config (used for both the pre-upgrade snapshot copy and the rollback restore copy) was never flipped to `auto_remove: false` — that flip actually landed on `dump_marker_present`'s probe container instead (see #2). Since `copy_volume`'s sidecar still auto-removes itself the moment `wait_container` returns, the new explicit `remove_container(...)` call added right after can race Docker's own auto-remove and get back "No such container" / 404. The tolerance check only forgives `"removal of container" + "is already in progress"` text — not "No such container"/404 (that broader match only exists later, in the removal-wait loop). So this 404 is treated as a hard failure, intermittently aborting exactly the snapshot and rollback paths this PR set out to harden.

**2. `dump_marker_present`'s probe container now leaks on every call.**
The `auto_remove: true → false` flip that was meant for `copy_volume`'s sidecar landed on this function's throwaway `busybox` probe container instead. No `remove_container` call was added for it. `dump_marker_present` runs at the top of the dump phase on every attempt/retry, so a stalled or repeatedly-retried upgrade now accumulates unbounded `temps_pg_upgrade_probe_*` containers on the host.

**3. `ensure_no_active_upgrade()` guard's error doesn't get proper HTTP treatment.**
The new `ExternalServiceError::UpgradeInProgress` variant returned by `start_service()` isn't handled by the `start_service` HTTP handler (`handlers.rs`) — that handler (and the `kv`/`blob` start-service handlers) do `match e.to_string().as_str() { "Service not found" => 404, _ => 500 }`, so `UpgradeInProgress` falls into the catch-all and comes back as a generic `500` instead of a `409 Conflict`. Also worth a look: the guard is only wired into `start_service()`, not `stop_service()` or other container-mutating entry points — the invariant it's protecting ("nothing may recreate this container/volume while an upgrade owns it") arguably belongs at a lower choke point than one call site.

**4. Missing test coverage for the `pg_isready -d` fix (511299c0).**
Nice catch on the FATAL-log-spam bug, but there's no test asserting the healthcheck command actually includes `-d <database>` for a service where `username != database` — worth adding a small unit test on the `HealthConfig.test` command built in both `postgres.rs::create_container` and `postgres_lifecycle.rs::create_and_start` so this can't silently regress.

**On grafting into #161:** not needed. #161 (MariaDB) doesn't touch `auto_remove`/volume-copy sidecar logic at all, and its healthcheck (`mariadb-admin ping`) doesn't have the pg_isready-style "-d defaults to -U" footgun since `ping` isn't database-scoped — so neither of these bug classes carries over. #161 has also been rebased onto the same latest `upstream/main` and is green.

Happy to push fixes for #1–#3 if useful, or if you'd rather keep this PR scoped, I can open them as a fast-follow once this merges.

bherila:
fixes for #1-3 coming shortly

bherila:
Pushed fixes for #1–#3 above (head is now `cdeb5812`):

- `40ae7453` — fixed the `auto_remove` mix-up: `copy_volume`'s sidecar now gets `auto_remove: false` (so it doesn't race the explicit removal/wait-for-removal check already in this PR), and `dump_marker_present`'s probe container is back to `auto_remove: true` (it has no removal logic of its own, so the earlier flip to `false` was leaking a container on every dump-phase check).
- `cdeb5812` — `start_service`'s error handling now matches on the `ExternalServiceError::UpgradeInProgress` variant directly and returns `409 Conflict` (with the OpenAPI doc updated), instead of falling through the `e.to_string()` catch-all to a generic `500`.

Left #4 (test coverage for the `pg_isready -d` fix) out of this push since it's an addition rather than a bug fix — happy to add it if you'd like it in this PR, otherwise it's a reasonable fast-follow.

Both `cargo check --lib -p temps-providers` and the existing `postgres_upgrade`/`postgres_upgrade_service` unit tests (17 passing, Docker-gated ones skip per repo policy) are clean on top of these changes.

bherila:
## Test coverage added for the `pg_isready -d` fix, CI running

Pushed `f7790765`, extracting the `pg_isready` healthcheck command construction into a small pure `postgres_healthcheck_cmd()` helper in both `postgres.rs` and `postgres_lifecycle.rs` (previously inline `format!` calls), with unit tests asserting `-d` is pinned to the configured database and that username/database are shell-escaped. This closes the gap flagged earlier — a future edit could otherwise silently drop `-d` again and reintroduce the FATAL-log-spam bug fixed in `511299c0` with nothing to catch it.

Locally: `cargo check --lib -p temps-providers` clean, `cargo fmt --check` clean, and the full `postgres` test suite (91 tests, including the 4 new ones) passes with no regressions.

Triggered CI on this push: **https://github.com/gotempsh/temps/actions/runs/28610805405**

Status as of this comment:
- Validate commits & preview changelog — ✅ pass
- Formatting — ✅ pass
- Cargo Check — ✅ pass
- Clippy (advisory) — ✅ pass
- Build Test Binaries — 🟡 in progress
- E2E Deployment Tests — 🟡 in progress ([run](https://github.com/gotempsh/temps/actions/runs/28610805578))

Will follow up if either of the in-progress jobs turns up something.

bherila:
## CI finished — all green ✅

Full run: https://github.com/gotempsh/temps/actions/runs/28610805405

| Check | Result | Duration |
|---|---|---|
| Validate commits & preview changelog | ✅ pass | 16s |
| Formatting | ✅ pass | 32s |
| Cargo Check | ✅ pass | 3m54s |
| Clippy (advisory) | ✅ pass | 4m46s |
| Build Test Binaries | ✅ pass | 8m53s |
| Unit Tests (unit-a) | ✅ pass | 9m19s |
| Unit Tests (unit-b) | ✅ pass | 19m36s |
| Integration Tests (postgres-upgrades) | ✅ pass | 10m8s |
| Integration Tests (docker-providers) | ✅ pass | 12m2s |
| Integration Tests (docker-deployments) | ✅ pass | 20m6s |
| Integration Tests (docker-backups) | ✅ pass | 23m25s |
| E2E Deployment Tests | ✅ pass | 23m42s |

`Integration Tests (postgres-upgrades)` is the group that actually exercises this PR's Docker-dependent hardening (volume handoff, readiness waits, the `auto_remove` fix) — green there is the most relevant signal for this diff.

dviejokfs:
I'm checking and testing it in my testing environment. I expect to be merging it today/tomorrow

dviejokfs:
### Why `b43c4a85` is needed

The 6 commits already on this PR harden the **correct** major-upgrade path (the `PostgresUpgradeOrchestrator`). But the branch HEAD still contained the **broken legacy path** this PR is meant to supersede — so merging as-is would still ship a data-loss bug.

`ExternalService::upgrade()` (the singular `PUT /external-services/{id}` upgrade) handled a major-version change (e.g. 17 → 18) by calling `run_pg_upgrade()`, which **never actually migrated data**. It:

- deleted the live PGDATA volume,
- kept an unused raw-file copy in a `_backup_N` volume, and
- booted a brand-new **empty** cluster hardcoded to `postgres:{version}-alpine` + `POSTGRES_USER=postgres`

…silently discarding the real configured username/database and **all data** while returning `Ok`. The original `test_postgres_v17_to_v18_actual_upgrade` only asserted `SELECT version()` on the post-upgrade container — which is exactly why the loss went unnoticed: it never checked that the pre-upgrade data survived.

**This commit:**

- deletes the 286-line broken `run_pg_upgrade()`;
- makes `upgrade()` **refuse** any major-version change up front — before any container stop or volume mutation, so it's a safe no-op rejection, not a partial failure — and directs callers to the orchestrator (`POST /external-services/{id}/upgrades`). Same-major-version image swaps (e.g. `postgres:18` → `gotempsh/postgres-walg:18`) still work as before;
- maps the rejection (and downgrade attempts) to **HTTP 400** instead of a generic 500;
- rewrites the test to assert the rejection **and** that the v17 container and a seeded marker row are left completely untouched.

`cargo clippy -p temps-providers --all-targets --features docker-tests` is clean; pre-commit fmt/clippy/conventional-commit hooks pass.

---

### ⚠️ Follow-up (not in this PR)

Majors now route **exclusively** through the orchestrator, which stages a rollback-volume copy + a logical `pg_dumpall` dump file + the rebuilt new cluster. Peak disk is therefore **~2–2.5× the database size** — so this is **not yet viable for very large (100s of GB) databases**, and there's currently no pre-flight disk check (a too-large upgrade fills the disk mid-restore).

Tracked separately:
1. Pre-flight free-disk guard (fail fast with a clear error instead of filling the disk).
2. Optional local rollback (rely on the mandatory S3 pre-upgrade backup) and/or streamed `pg_dumpall | psql` restore to cut peak toward ~1× extra.
3. Longer term: native `pg_upgrade --link` for large DBs (near-zero extra disk).

bherila:
This still isn't quite ready, I'll push some more fixes and another comment soon.

bherila:
## Review (multi-agent, verified against PR head `4701fdbe`)

This PR is a solid, well-tested hardening pass: it stops the legacy `upgrade()` path from silently discarding data on a major-version bump, tightens the healthcheck to the configured database, requires clean volume handoffs before pg18 reopens pg17-era `PGDATA`, and the latest commit extends the dump/restore test to assert full-cluster survival across multiple databases (a real gap in the old single-table test). I fanned out 8 independent reviewers across correctness, reuse, simplification, efficiency, architectural depth, and CLAUDE.md conventions, then verified each candidate by reading the code directly at the PR head. Two of the surviving findings are real correctness/data-integrity gaps worth fixing before merge; the rest are lower-severity but still worth a look.

### Correctness

**1. `copy_volume` leaks its sidecar container on a transient `wait_container` error, permanently wedging the upgrade** — `postgres_upgrade.rs:1544` disables `auto_remove` ("we remove manually below"), but the manual `remove_container` call only runs *after* `wait_container` succeeds (`postgres_upgrade.rs:1579-1591`). If `wait_container` itself errors (daemon hiccup, container removed out-of-band), the function returns via `?` before ever calling `remove_container` — the copy sidecar is left running, still holding references to both the source and dest volumes. Any retry then hits the new `remove_volume_or_fail` hard-fail ("volume is in use") forever, until an operator manually `docker rm`s the leaked container. Worth wrapping the wait/remove pair so removal always runs (e.g. `defer`-style cleanup or matching on the wait error and still attempting removal before returning).

**2. The concurrent-upgrade guard (`ensure_no_active_upgrade`) doesn't cover all the paths it needs to** — `services.rs:6732` is only called from `start_service` (`services.rs:6763`); `initialize_service` (`:3937`), `update_service` (`:1571`), and the legacy same-version `upgrade()` path never call it. More importantly, during `rollback()` (`postgres_upgrade.rs:1819-1933`) the row's `status` stays `"completed"` — it only flips to `rolled_back` in the very last step (`:1928`) — even while step 2 is actively deleting the live volume and step 2-3 are recopying it. A concurrent `start_service` mid-rollback sails right past the guard's `PENDING`/`RUNNING`-only check, reproducing exactly the "recreate against an implicitly re-created empty volume" hazard the guard's own doc comment (`services.rs:6725-6728`) says it exists to prevent.

**3. `wait_for_container_healthy` hard-fails on the first transient error, inconsistent with its sibling loop** — `postgres_lifecycle.rs:237` propagates any `inspect_container` error immediately via `?` inside the poll loop, aborting the whole 120s readiness wait on one hiccup. `wait_for_psql_database` a few lines above (`:168-203`) handles the same class of transient exec error by simply ignoring it and retrying (`if let Ok(...)` at `:197`). Two loops solving the same "wait for Postgres to be usable" problem in the same file, with two different failure-tolerance policies.

**4. New major-version rejection is a bare `anyhow` string classified by substring match** — `postgres.rs:3090-3121` returns `anyhow::anyhow!("Cannot change PostgreSQL major version...")` / `"Cannot downgrade PostgreSQL..."`, and `handlers.rs:713-719` routes it to 400 via `msg.contains(...)`. This is the exact pattern CLAUDE.md's error-handling section forbids ("NEVER use `anyhow::Result` in service layer"; "errors flow upward through typed conversions, never through string coercion"), and it's inconsistent with this same PR's own `UpgradeInProgress` variant a few lines below, which is a properly typed, exhaustively-matched enum variant. A future rewording of either message silently turns the intended 400 into a 500 with nothing to catch it.

**5. (Pre-existing, but touched by this PR) `handlers.rs:1215` — dead string-equality guard returns 500 instead of 404** — `_ if e.to_string() == "Service not found"` can never match: `ExternalServiceError::ServiceNotFound`'s `#[error(...)]` renders as `"Service {id} not found"` (`services.rs:35-36`), never the bare literal. Starting a since-deleted service falls through to the catch-all 500 instead of the 404 the OpenAPI annotation two lines above promises. This exact dead comparison already existed before this PR, but the PR rewrites this precise match block to add the `UpgradeInProgress` → 409 arm and left the neighboring dead arm as-is — `ExternalServiceError::ServiceNotFound { .. } => ...` was a one-line fix away.

**6. (Narrow/timing-dependent) `ensure_database_exists` has no retry of its own** — `postgres_lifecycle.rs:279-323` runs a single exec right after `wait_for_psql_database` confirms one successful `SELECT 1` against the `postgres` database. The surrounding comment (`:539-541`) already acknowledges "a transient init server can accept SQL briefly before shutdown," but a single successful check doesn't guarantee stability — if that one success lands against the entrypoint's temporary init-phase server right before it's swapped for the final one, the subsequent un-retried `ensure_database_exists` exec can land in the gap and fail the whole phase outright.

### Cleanup

**7. Duplication across `postgres.rs` / `postgres_upgrade.rs` / `postgres_lifecycle.rs`** — `shell_escape` now exists in 3 places (`postgres.rs:35`, `postgres_upgrade.rs:227`, `postgres_lifecycle.rs:90`); `postgres_healthcheck_cmd` is copy-pasted verbatim (comment included) into `postgres_lifecycle.rs:98`; and `postgres_lifecycle.rs`'s `wait_for_container_healthy`/`wait_for_psql_database`/`exec_in_container` re-implement the same exec-drain-then-inspect / poll-until-deadline plumbing that `postgres_upgrade.rs` already has (`exec_and_wait`, `wait_for_pg_ready`, `wait_container_exit`). Notably, this PR's own motivating bug (per the CHANGELOG) was drift between two duplicated healthcheck command builders — this duplication invites the exact same class of bug again on the next readiness/healthcheck tweak.

---

**Recommendation:** not quite ready to merge as-is. #1 and #2 are real correctness/data-integrity gaps in the upgrade path this PR is specifically hardening — worth fixing before merge. #3 and #4 are lower risk but cheap to fix and directly relevant to the PR's own stated goals. #5–#7 are good cleanup but not blocking.

*Reviewed with 8 parallel Fable-model agents (correctness × 3, reuse/simplification/efficiency × 3, altitude, CLAUDE.md conventions), findings verified by direct code reading at the PR head.*

bherila:
## Follow-up: review findings addressed

All 7 findings from the review above are fixed, in `e2d13e61` (fixes) and `e1aedb4e` (rustfmt/clippy cleanup on top):

1. **`copy_volume` sidecar leak** — removal now always runs even if `wait_container` itself errors, instead of leaking the copy container and permanently wedging retries.
2. **`ensure_no_active_upgrade` guard gaps** — now also guards `initialize_service` (covers `update_service`, service creation, and the resource-limit recreate path) and the legacy same-version `upgrade_service`. Added a `ROLLING_BACK` status so the guard also blocks during `rollback()`'s multi-step restore, not just while the row was PENDING/RUNNING.
3. **`wait_for_container_healthy`** now tolerates transient `inspect_container` errors and retries until the deadline, consistent with its sibling `wait_for_psql_database`.
4. **Major-version rejection** is now a typed `PostgresUpgradeRejected` error instead of a bare `anyhow!` string, downcast at the `services.rs` boundary into a typed `ExternalServiceError::UpgradeRejected` variant — no more classifying by substring-matching the error message.
5. **`start_service`'s dead `"Service not found"` string match** replaced with a typed match on `ExternalServiceError::ServiceNotFound` (was silently falling through to 500 instead of the documented 404).
6. **`ensure_database_exists`** now retries for up to 30s instead of a single exec, closing the race against the postgres entrypoint's transient init-phase server.
7. **Deduped `shell_escape`/`postgres_healthcheck_cmd`** across `postgres.rs`, `postgres_upgrade.rs`, and `postgres_lifecycle.rs` (plus their duplicated unit tests). Scoped narrowly to these two pure string-builders — left the deeper exec/wait-loop overlap between `postgres_lifecycle` and `postgres_upgrade` as a follow-up rather than risk an untested refactor.

CI is fully green on the updated head (all 12 checks): Cargo Check, Clippy, Formatting, Build Test Binaries, both Unit Tests shards, all four Integration Tests shards (including `postgres-upgrades`, which directly exercises the fixed paths), E2E Deployment Tests, and changelog validation.

dviejokfs:
### `3ed8d13b` — fixes a 6-hour, 100%-CPU hang that made the upgrade feature non-functional, + regression tests

While validating this PR end-to-end (create v17 → load data → upgrade via the endpoint) the upgrade **hung indefinitely in `pre_backup`** every time. Root-caused it to a real bug in `exec_util::run_exec` (the shared exec helper used by **all** external-service backups — postgres/mongodb/redis, not just upgrades):

- Its `tokio::select!` is `biased` and polls the exec **output stream** first. Once the exec's output hits EOF, `output.next()` returns `Ready(None)` on *every* poll, so the biased select **always** takes that branch and **never runs the `inspect_exec` branch** that detects completion. Result: a **100% CPU busy loop that never returns and never even hits `run_exec`'s own 6-hour timeout.**
- It triggers whenever the exec'd command writes to a file / redirects (empty exec stdout) — i.e. the normal `pg_dumpall … > file` backup shape. Observed live: temps process pinned at **107% CPU**, upgrade stuck in `pre_backup`, dump file already complete on disk, `run_exec` never returning.

**Fix:** gate the output branch with `if !output_done` so once the stream is drained the select falls through to the completion/deadline branch.

**Why it shipped (and the tests that close the gap):**
- The existing `orchestrator_happy_path_v17_to_v18` runs the full orchestrator to `completed`, but **stubs the backup**, so it never touches `run_exec`.
- Added `run_exec_returns_for_a_command_with_no_stdout` (direct unit guard, timeout-bounded) and `orchestrator_full_run_exercises_backup_exec_v17_to_v18` (+ `RunExecBackupProvider`) which drives the full orchestrator with `pre_backup` routed through the **real** `run_exec`. Both pass; both fail (hang → bounded timeout) without the fix.

### End-to-end validation (real, via the `POST …/upgrades` endpoint)

| Scenario | Result |
|---|---|
| 1M rows / 102 MB, 17→18 | ✅ completed ~25s; fingerprint (rows, `sum(id)`, `sum(n)`) identical; v18.4 |
| **20M rows / 22 GB**, 17→18 | ✅ completed **~4m40s**; fingerprint identical; databases preserved |
| Cancel mid-`pre_backup` | ✅ data intact, still v17 (safe no-op confirmed) |

Measured at 22 GB scale (real numbers, worth knowing for prod):
- **Downtime ≈ 2m36s** (snapshot→completed). `pre_backup` (~2m) is zero-downtime — old container serves while it's dumped. Downtime = snapshot-copy + dump + restore, so it **scales with DB size** (this is a logical dump/restore, *not* low-downtime).
- **Peak disk ≈ +42 GB** on top of the DB (~2× the DB size) — confirms the disk-footprint concern.

### Merge-readiness (my take)

Now solid: data-loss reject, sed injection fix (security-auditor signed off), typed errors, concurrency guards + `ROLLING_BACK`, no new migrations, and now the `run_exec` hang + regression coverage.

Still recommend before calling it production-ready (tracked as follow-ups, not necessarily this PR):
1. **Pre-flight disk guard** — a large-DB upgrade fills disk mid-`restore`; fail fast instead.
2. **Verify the pre-upgrade backup targets the right DB** — on a shared box the main backup connected via `host=localhost` and dumped a *different* pg18 container; confirm on a clean single-tenant node.
3. **Honest downtime/disk confirmation** in the flow (offline upgrade, est. downtime scales with size, needs ~2.5× disk).

Zero-downtime (logical-replication blue/green + a repointable Postgres endpoint / pgbouncer) is a **separate feature/ADR**, not a blocker — the offline path here is the right v1 for small/standalone DBs.

dviejokfs:
### `692f0032` — review punch-list done (all 3 items)

Follow-up to the review above; all five tests pass and `cargo clippy --all-targets --features docker-tests` is clean.

1. **Regression tests now fail cleanly instead of hanging.** The review noted the no-stdout `run_exec` guard + full-orchestrator test wrapped the call in `tokio::time::timeout`, but a regression is a *synchronous* busy loop that never yields `Pending`, so the timeout could never fire — a revert would hang CI. Fixed: run the call on a separate task (`#[tokio::test(flavor = "multi_thread")]` + `tokio::spawn`) and time out the `JoinHandle`, so the spin occupies one worker while the timeout fires on another.

2. **`UpgradeInProgress` → 409 everywhere.** Extracted a shared `upgrade_error_problem` helper (`UpgradeRejected`→400, `ServiceNotFound`→404, `UpgradeInProgress`→409) used by `upgrade_service`, `update_service`, and `start_service` (previously only `start_service` mapped 409; the other two returned 500). Added 409 to their OpenAPI, and dropped a dead `"Service not found"` string arm in `update_service`.

3. **Tests for the safety-critical bits the review flagged:**
   - `ensure_no_active_upgrade_blocks_only_active_statuses` — the core guard blocks PENDING/RUNNING/ROLLING_BACK, allows terminal statuses.
   - `upgrade_error_problem_maps_shared_variants_to_canonical_status` (Docker-free) — asserts each variant's status via `into_response().status()`.
   - `postgres_upgrade_rejected_roundtrips_through_anyhow` (Docker-free) — guards the anyhow→typed downcast that maps a rejected upgrade to 400.

Remaining follow-ups (tracked separately, not blockers): pre-flight disk guard, verify the pre-upgrade backup targets the right DB on a single-tenant node, downtime/disk confirmation UX, and the pre-existing password-in-restore-entrypoint LOW. This branch is now clean to merge from my side.

dviejokfs:
LGTM ✅ — good to merge from my side.

Verified locally: **253 tests pass, 0 fail** (242 non-Docker unit + 11 Docker-gated upgrade integration — orchestrator happy-path/rollback/phases, dump-restore, `run_exec`, rejection, and the `ensure_no_active_upgrade` guard), plus `cargo clippy --all-targets --features docker-tests` clean. Also validated the real endpoint end-to-end at scale (17→18 on a 20M-row / 22 GB DB, data intact).

The review's punch-list is done (`692f0032`): unified `UpgradeInProgress`→409 mapping, the busy-loop regression tests now fail cleanly instead of hanging, and the previously-untested guard/mapping/downcast paths are covered.

Remaining items are tracked follow-ups, not blockers: pre-flight disk guard, verify the pre-upgrade backup targets the right DB on a single-tenant node, downtime/disk confirmation UX, and the pre-existing password-in-restore-entrypoint LOW.

PR #153, opened 2026-06-24
fix(temps-cli): per-instance default project, static deploy in `up`, and full build logs

dviejokfs:
## Summary

A batch of `@temps-sdk/cli` fixes and improvements, all scoped to `apps/temps-cli/`. Bumps the CLI to **0.1.28**.

### 1. Default project is now scoped per instance
Previously `defaultProject` lived in a single global config key with no instance association, while `apiUrl` was already context-aware. So a project created on one Temps server leaked as the default on another and 404'd:
```
ℹ Using project telemetry-dashboard (from global-config)
✖ Project "telemetry-dashboard" not found
```
- `defaultProject` now lives on the **active CLI context** (`~/.temps/.contexts.json`). A new instance starts with no default.
- Resolution reads the active context's default (`context-default`) before the legacy global key (kept for back-compat, drains over time).
- `create` / `delete` / `configure get|set defaultProject` all write through the new `setDefaultProject` helper.

### 2. `temps up` falls back to the create wizard on a stale default
When a resolved-but-stale slug 404s, `up` now runs the setup wizard (offers to create) instead of dead-ending on a 404 — unless the slug came from an explicit `--project` flag.

### 3. Static deploy support in the `up` wizard
Static sites were being routed through the Docker-image **manual** path (Dockerfile generation), which made no sense for a pre-built folder.
- New **"Static (upload a pre-built folder)"** option → `source_type: static_files`, `project_type: static` (preset stays the valid `dockerfile` placeholder; `static` is only a local detection label, not an API preset — this was the `Invalid preset: static` error).
- `detectStaticDir()` auto-detects the build-output folder (`dist`/`build`/`out`/`public`/`_site`/`output`, or a root `index.html`), and the wizard **always lets you pick a custom folder**.
- New `--static` / `--static-dir` flags. Uploads via `deployStatic`.

### 4. Full Docker build logs instead of one spinner line
The local-image build funneled every output line into a single rewriting spinner, so the per-step BuildKit log was lost — you'd only see e.g. `#25 exporting layers`.
- Force `--progress=plain` (+ `BUILDKIT_PROGRESS=plain`) and stream each build line to the terminal (dimmed/indented). Quiet/`--json` mode stays silent.
- Affects both `deploy:local-image` and the `up` manual path.

## Testing
- `bun run typecheck`: zero errors in touched files (14 pre-existing errors remain in untouched `env-sync`/`environments`/`providers`/`notifications`).
- New `detect-project.test.ts` (6 tests for `detectStaticDir`) + existing `contexts.test.ts` → 15 pass, 0 fail.
- Static path verified end-to-end against a local server: project created, bundle uploaded, deployment completed, live URL served the content.
- `--progress=plain` verified to stream the full step log on a real `docker build`.

## Note
This is the `@temps-sdk/cli` npm package — changes take effect for `bunx @temps-sdk/cli` only after publishing 0.1.28 to npm.

### DIFF
--- a/apps/temps-cli/package.json
+++ b/apps/temps-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@temps-sdk/cli",
-  "version": "0.1.27",
+  "version": "0.1.28",
   "description": "CLI for Temps deployment platform",
   "type": "module",
   "bin": {
--- a/apps/temps-cli/package.json
+++ b/apps/temps-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@temps-sdk/cli",
-  "version": "0.1.26",
+  "version": "0.1.27",
   "description": "CLI for Temps deployment platform",
   "type": "module",
   "bin": {
--- a/apps/temps-cli/src/commands/configure.ts
+++ b/apps/temps-cli/src/commands/configure.ts
@@ -1,5 +1,7 @@
 import type { Command } from 'commander'
 import { config, credentials, type TempsConfig } from '../config/store.js'
+import { setDefaultProject } from '../config/resolve-project.js'
+import { getActiveDefaultProjectSync } from '../config/contexts.js'
 import { promptUrl, promptSelect, promptConfirm, promptPassword } from '../ui/prompts.js'
 import { success, info, colors, newline, header, keyValue, icons, box } from '../ui/output.js'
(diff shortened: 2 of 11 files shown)


dviejokfs:
Superseded by #154 — re-homed onto a correctly-named branch (`feat/temps-cli-up-improvements`). This branch was misnamed; it never contained rollback changes.

PR #161, opened 2026-06-26
Add opt-in MariaDB-compatible services

bherila:
## Summary

Adds an opt-in MariaDB-compatible external service for hosted projects while leaving Temps' own internal PostgreSQL database unchanged.

> Recreated from #138, which could not be reopened after the branch was rebased. This PR supersedes #138 and is now **scoped to MariaDB/PITR only** — the Postgres upgrade hardening that previously rode along has been split out into #151.

This includes:
- a new MariaDB service implementation with create/import/update/delete, credentials, environment bindings, and backup/restore support via `mariadb-dump` with `mysqldump` fallback
- MariaDB query explorer support, including schema/table discovery and guarded SQL execution helpers
- frontend service presets/forms/icons so users can choose MariaDB from the existing service workflows
- conservative small-host defaults for MariaDB containers, plus standard/dedicated size profiles for larger hosts
- MDX docs covering MariaDB managed services, project linking, Laravel usage, backup coverage, and opt-in service startup behavior

### Point-in-time recovery (PITR)

- MariaDB binary logging enabled for PITR, with a configurable interval setting
- binary logs shipped to S3 for PITR
- restore + point-in-time recovery using `mariadb-binlog` for replay (with parity tests)
- a binlog-health probe surfacing operational warnings
- non-InnoDB table warnings as a PITR safety guard
- auto-provisioning of a default daily base schedule for MariaDB PITR
- repaired MariaDB PITR restore chain

## Notes

The service uses MariaDB images by default while exposing MySQL-compatible connection URLs and environment variables, since most frameworks and drivers connect to MariaDB through MySQL-compatible drivers.

MariaDB remains opt-in. Installing or starting Temps does not pull or run a MariaDB container; a container is created only when a user creates or imports a MariaDB external service. Linked projects receive separate per-project/per-environment databases inside that shared service.

## Rebase & scope

- Rebased cleanly onto upstream `main` (`f815de0b`); 14 MariaDB-only commits, 0 behind.
- Postgres upgrade hardening removed from this branch and tracked separately in #151. The MariaDB PITR e2e suite depends on those Postgres readiness fixes, so this PR is best reviewed/merged after #151 lands.

## Validation

Ran from `~/proj/temps`:

- `cargo check --lib -p temps-providers -p temps-backup -p temps-migrations` — clean (exit 0) on the rebased branch
- `cargo test -p temps-providers mariadb --lib` — MariaDB-focused provider tests pass (13 tests)
- `cd web && bun install && TEMPS_VERSION=pr-validation bun run build`

Draft until #151 (Postgres upgrade hardening) is merged, which unblocks the MariaDB PITR e2e suite.


dviejokfs:
@bherila I'm going to try and merge this, any work missing?

bherila:
This is working in my release branch, so I think it's ok. Small rebase may be needed, you can do it or let me know if you want me to do it tonight.

bherila:
@dviejokfs  I will get this one ready to merge first, then after it's merged I will address my other open PRs.

bherila:
## CI status after rebase onto `main` + #179

Rebased this branch onto current `main` and merged in #179 (fork-CI comment-permission fix). Result ([full run](https://github.com/gotempsh/temps/actions/runs/28566925411)):

- ✅ **Changelog** — now passes (was 403'ing on the fork-token issue #179 fixes)
- ✅ **E2E Tests** — passes
- ❌ **Rust Tests** — fails, but for reasons unrelated to this PR's MariaDB code:
  - [`Integration Tests (postgres-upgrades)`](https://github.com/gotempsh/temps/actions/runs/28566925411/job/84699608115): fails on `orchestrator_happy_path_v17_to_v18`, `orchestrator_phase_dump_is_idempotent`, `orchestrator_rollback_restores_pre_upgrade_data`. This is the documented dependency on #151 — this suite needs those Postgres readiness fixes to pass.
  - [`Integration Tests (docker-providers)`](https://github.com/gotempsh/temps/actions/runs/28566925411/job/84699608127): `test_update_service_parameters` — Docker `port is already allocated` on container start. Looks like CI-runner port-reuse flakiness, not MariaDB-related.
  - [`Integration Tests (docker-backups)`](https://github.com/gotempsh/temps/actions/runs/28566925411/job/84699608129): MongoDB/Postgres/Redis S3 backup-and-restore tests time out. None touch MariaDB; looks like runner resource contention.
  - [`Integration Tests (docker-deployments)`](https://github.com/gotempsh/temps/actions/runs/28566925411/job/84699608138): ACME/Pebble HTTP-01 challenge timeouts + one `temps-vulnerability-scanner` integration test. Also unrelated to this PR.

I verified locally that the rebase itself is clean: `cargo check --lib`, `cargo fmt --all -- --check`, and the MariaDB-focused suite (92 tests across `temps-providers`/`temps-backup`) all pass.

@dviejokfs — could you take a look and weigh in on whether these should block merge? My read: `postgres-upgrades` is an expected/intentional block until #151 lands (per this PR's description), and the other three look like pre-existing flakiness rather than regressions, but I'd like a second opinion before treating them as safe to ignore.

bherila:
I will rebase and fix this after #151

bherila:
While chasing down CI flakiness on this branch, found and fixed a real bug in `update_service()` (`crates/temps-providers/src/services.rs`) — not specific to MariaDB, affects any service type.

**Bug:** `update_service()` persists a rename to the database *before* calling `initialize_service()`. `initialize_service()` rebuilds its Docker service instance — and therefore which container it targets for `stop()` — from whatever name is currently in the DB. With the rename already persisted, `stop()` looks for a container under the *new* name (which never existed), silently no-ops ("may not exist"), and the still-running *old* container keeps holding its host port. The replacement container then fails to start with `"port is already allocated"`.

This reproduced deterministically — `services::tests::test_update_service_parameters` failed identically on 3 consecutive runs (not flaky), and would affect any production rename of a Postgres/MariaDB/MongoDB/Redis/etc. service via the update API, orphaning the old container and failing the update.

**Fix** (`e00dd84f`): before persisting the rename, if the name is actually changing, stop the old container using its pre-rename identity (name + service type), so the container that's actually running gets stopped and its port released before the recreate step tries to bind it under the new name.

Verified locally: `cargo test --lib -p temps-providers --features docker-tests services::tests::test_update_service_parameters` — was failing consistently, now passes.

Also pushed a separate fix (`cc7473b9`) for a hardcoded port (`15432`) in the Postgres backup/restore test that was causing intermittent `docker-backups` CI failures — matched the existing `find_available_port` pattern already used by the Redis/Mongo equivalents.

bherila:
Update on the remaining CI failures, plus independent verification.

## Both fixes confirmed clean
Ran the full suite twice more — once via the normal PR-triggered CI, and once as an independent trigger on my own fork's Actions (bherila/temps run [28699750073](https://github.com/bherila/temps/actions/runs/28699750073), same branch, separate runner/infra) since I can't retrigger checks on this repo directly. Both runs agree:
- `docker-providers` (validates the `update_service` rename fix, `e00dd84f`/`ba046a11`) — **pass**, consistently.
- `postgres-upgrades` — **pass**, consistently.
- No more `"port is already allocated"` errors anywhere (the original `postgres.rs:4496` hardcoded-port bug, fixed in `cc7473b9`, is gone).

Also got an independent second-opinion review (different reviewer, fresh read of the code, explicitly asked to look for reasons my analysis could be wrong) that reached the same conclusion on both fixes.

## `docker-backups` still fails, but it's a separate, pre-existing bug — not caused by these fixes
Symptom changed from "port already allocated" to a straight hang: all three backup tests (mongodb, postgres, redis) now time out at their own per-test deadline (300s/300s/180s) mid-operation, with **the exact same durations and ~785s total wall time** across both independent CI runs — this is deterministic, not flaky infra.

Mechanically it can't be either of my fixes: `docker-backups` runs `--test-threads=1` in alphabetical order, so `mongodb`'s test runs (and hangs) *before* the postgres/rename code either fix touches ever executes.

**Root cause, as far as I can tell:** `MinioTestContainer::start` (`externalsvc/test_utils.rs:70-88`) creates the MinIO container with no `NetworkingConfig`/`network_mode` — it lands on Docker's default `bridge` network. `resolve_endpoint_for_container` (`externalsvc/mod.rs:131-257`) then finds MinIO "on the same network" as the mongo/postgres/redis container and resolves the S3 endpoint to `http://<minio-container-name>:9000` — but Docker's default bridge network doesn't provide container-name DNS resolution (only user-defined networks do). So wal-g inside the service container can't actually reach that hostname, and appears to hang/retry until the outer test timeout fires rather than failing fast. (There's also a separate, smaller gap: no `aws_sdk_s3::Client` construction anywhere in the workspace sets an explicit `TimeoutConfig`, so only the SDK's default ~3.1s *connect* timeout applies — no operation timeout. Didn't confirm this as the trigger here, but worth its own fix regardless.)

Likely fix: have the test setup attach the MinIO + service containers to an explicit user-defined Docker network (bollard `NetworkingConfig`) instead of relying on the default bridge, so the resolved container-name endpoint is actually DNS-resolvable. Happy to take a pass at this if useful, but flagging it now since it's out of scope for what I was originally chasing down and deserves its own review.

bherila:
Rebased this branch onto \`upstream/main\` (now at \`5265ba0f\`) and pushed (\`cd6e1eb2\`).

## The docker-backups hang fix is now bundled — via upstream, not a new commit
While investigating, I found that \`upstream/main\` already has a comprehensive fix for the exact hang I diagnosed above (\`8a5a8a70\`, "fix(ci): make rust-tests reliably fail on test failures", by David Viejo) — it covers the MinIO-never-joins-the-shared-network root cause precisely, plus the Pebble ACME hardcoded-gateway issue, Trivy DB caching/locking, and the cross-process Postgres deadlocks I'd separately flagged as unrelated `docker-deployments` flakiness. Writing a new fix myself would have been redundant and conflict-prone against that work.

Since it's rebased in as a normal upstream commit, it's already standalone and trivially cherry-pickable on its own if you ever need to pull just that piece elsewhere — no extra work needed there.

## CHANGELOG.md edits removed from this branch
This branch's manual CHANGELOG.md entries (added across several commits) had gotten tangled with upstream's own changelog evolution and were a source of conflict noise. Reset to match `upstream/main` exactly — zero diff on that file now.

## Conflict resolution
All 18 commits replayed automatically except one (`fb184434`, now `e7f529f1` after rebase), which conflicted in `.github/workflows/rust-tests.yml`: upstream replaced this branch's old retry-based test script (in `8a5a8a70`, for the same reasons above — it silently swallowed real failures) with a direct `cargo test` invocation. Resolved by taking upstream's cleaner version while preserving this branch's own addition, the "MariaDB PITR E2E" job, verbatim.

Verified locally: `cargo check --lib --workspace` — clean. Fresh CI run in progress now; will report back once it completes, including whether `docker-backups` finally passes given the upstream fix.

bherila:
✅ **All CI checks now pass.**

Final status after the rebase onto \`upstream/main\`:

| Check | Status |
|---|---|
| Formatting | ✅ |
| Cargo Check | ✅ |
| Clippy (advisory) | ✅ |
| Build Test Binaries | ✅ |
| Validate commits & preview changelog | ✅ |
| Unit Tests (unit-a, unit-b) | ✅ |
| E2E Deployment Tests | ✅ |
| MariaDB PITR E2E | ✅ |
| Integration Tests (docker-providers) | ✅ |
| Integration Tests (postgres-upgrades) | ✅ |
| **Integration Tests (docker-backups)** | ✅ (previously hung — fixed by upstream's `8a5a8a70` networking fix, bundled via the rebase) |
| **Integration Tests (docker-deployments)** | ✅ (previously flaky on unrelated Postgres deadlocks/Pebble ACME/Trivy — fixed by upstream's test-serialization commit, also bundled via the rebase) |

Both remaining failures from earlier in this thread are now resolved, entirely by rebasing onto upstream's own already-merged fixes rather than adding anything new/redundant here. Branch is otherwise unchanged in intent — still the same MariaDB PITR + backup/restore feature set, now on a current base with a clean CHANGELOG.md and no lingering conflicts.

bherila:
@dviejokfs thoughts on merging this while i'm working on the other minIO test

dviejokfs:
merged, let's get all the PRs merged to main so it can be tested with all the fixes

PR #163, opened 2026-06-28
fix(notifications): disambiguate update_email_provider operationId

dviejokfs:
## Problem

Both `temps-email` and `temps-notifications` exposed a handler named `update_email_provider`:

- `temps-email` — transactional email providers, `PATCH` (crates/temps-email/src/handlers/providers.rs:314)
- `temps-notifications` — notification email provider, `PUT /notification-providers/email/{id}` (crates/temps-notifications/src/handlers.rs)

Both generated the **same `updateEmailProvider` operationId** in the OpenAPI spec. The generated TypeScript SDK silently shadowed one operation with the other, so callers could hit the wrong endpoint.

## Fix

Rename the **notifications** handler to `update_notification_email_provider` so its generated op is `updateNotificationEmailProvider`, distinct from the email crate's `updateEmailProvider`.

- Renamed the handler fn, its utoipa `paths(...)` registration, the route binding, and the unit test
- Regenerated the web SDK (`sdk.gen.ts`, `types.gen.ts`, `@tanstack/react-query.gen.ts`)
- Updated the two callers: `ProvidersManagement.tsx`, `EditNotificationProvider.tsx`

## Notes

- Route path is unchanged (`PUT /notification-providers/email/{id}`); only the operationId/SDK symbol changes.
- `cargo fmt` + `cargo clippy` pass (pre-commit).


bherila (commented):
## Review: fix(notifications): disambiguate `update_email_provider` operationId

**Summary:** This is a correct, minimal, well-scoped fix for a real bug. Two backend handlers (`temps-email`'s transactional provider update and `temps-notifications`' email-provider update) both compiled to the OpenAPI operationId `updateEmailProvider`. The generator resolved the collision by auto-suffixing one symbol to `updateEmailProvider2` — but the React callers for the *notification* email provider were wired to the un-suffixed `updateEmailProviderMutation`, i.e. the **transactional email** endpoint. So editing a notification email provider in the UI was hitting the wrong API. Renaming the notifications handler to `update_notification_email_provider` removes the collision and re-points the callers correctly.

I verified the change at head (`89bc6c3`):
- Only one backend `update_email_provider` remains (`crates/temps-email/src/handlers/providers.rs:314`); the notifications handler is now `update_notification_email_provider` (`crates/temps-notifications/src/handlers.rs:945`). Collision resolved.
- No `updateEmailProvider2` / `EmailProvider2` symbols remain anywhere under `web/`.
- The rename is consistent across the utoipa `paths(...)` registration, the route binding (`PUT /notification-providers/email/{id}` path unchanged), the request type (`UpdateNotificationEmailProviderRequest`), the unit test, the regenerated SDK (`sdk.gen.ts`, `types.gen.ts`, `@tanstack/react-query.gen.ts`), and both callers (`ProvidersManagement.tsx`, `EditNotificationProvider.tsx`).
- Handler body is an exact rename — `permission_guard!(auth, NotificationProvidersWrite)`, audit logging, and config decryption are all preserved.
- Despite the 188 KB diff, the SDK regeneration introduced **no drift**: net change is +20/-20 (the large hunks are single-line `import { ... }` lists). Good sign that the generator was run cleanly and nothing else moved.

### Blocking
None.

### Should-fix
None blocking. One robustness suggestion: the operationId is derived implicitly from the Rust fn name, which is exactly what allowed this silent collision. Consider adding an explicit `operation_id = \"updateNotificationEmailProvider\"` to the `#[utoipa::path(...)]` macro so the wire/SDK contract is decoupled from the fn name, and (separately) a CI check that fails on duplicate operationIds in the generated spec so this class of bug can't regress unnoticed.

### Nits
- PR description says the SDK \"silently shadowed one operation with the other.\" In practice the generator emitted a distinct `updateEmailProvider2`; the actual defect was that the notification callers referenced `updateEmailProvider` (the email crate's op), so the wrong endpoint was called. Minor wording only.
- Back-compat: the SDK symbol/operationId changed (`updateEmailProvider2` → `updateNotificationEmailProvider`). Any external consumer pinned to the old auto-suffixed name would break — but since `*2` was an unstable, generator-assigned name, this is the right call; worth a one-line note in release notes.
- CI: `gh pr checks` reports no checks on this branch. Worth confirming the web typecheck/`cargo` build ran somewhere before merge, since this touches generated TS.

### Praise
- Fixes a genuine, user-facing correctness bug (notification email edits hitting the transactional-email endpoint) rather than just a cosmetic naming clash.
- Minimal blast radius, clean regeneration with no diff drift, and the existing test was renamed/kept rather than dropped.

_Automated review via Claude Code._

PR #164, opened 2026-06-28
fix(deployments): cap hosted website memory by default

bherila:
## Summary

Provide a conservative default memory cap for hosted-website containers while preserving a clear operator escape hatch.

- New projects and their default `production` environment seed `memory_limit=512` MB.
- `None` continues to inherit the parent/default value, while `Some(0)` explicitly selects an uncapped container.
- Docker sets `memory_swap == memory` for capped containers so the advertised hard cap includes swap.
- Project and environment settings expose the distinction as **Use default** versus **Uncapped**.
- Tests cover seeded defaults, validation, and inheritance of the uncapped sentinel.

## Scope revision after review

The host-wide admission gate has been removed from this PR. It treated container limits as reservations, discouraged normal overcommit, and required an N+1 list/inspect pass during deployment. The prototype is preserved separately on the fork branch [`feature/host-memory-admission-prototype`](https://github.com/bherila/temps/tree/feature/host-memory-admission-prototype) for a future design based on explicit capacity policy and better accounting.

The hand-maintained `CHANGELOG.md` edit is also removed because release notes are generated from conventional commits.

## Landing order

This branch is intentionally stacked on the current head of #170. Merge #170 first with commit ancestry preserved, then #164 reduces to only the hosted-memory-cap commits. If #170 is squash-merged or rebase-merged, rebase #164 once more onto the resulting `main` before merging.

## Validation

CI runs on this rebuilt PR head. Local validation includes `cargo fmt --check`, the affected Rust workspace packages, and `bun x tsc --noEmit`.


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+<<<<<<< HEAD
```
dviejokfs: careful with this

bherila:
Fixed. Changelog will cause every PR to conflict btw. Can we do something structurally to make that better?

dviejokfs:
@bherila you are right, how do other projects handle it? updating the changelog once all PRs are merged?

crates/temps-deployer/src/docker.rs
```diff
@@ -732,6 +756,96 @@ impl DockerRuntime {
...
+                }
+                Err(e) => warn!(
+                    "Failed to inspect container {} for memory admission ({}); excluding from reserved total",
+                    id, e
+                ),
+            }
+        }
+        reserved_mb
+    }
+
```
dviejokfs: reserved_managed_memory_mb does N+1 inspect_container calls, it can be very costly and it's not deterministic, I'd rely on the user and not impose a limit on the max memory. Although an intermediate solution would be ideal

dviejokfs:
@bherila 
Seeded production env gets an explicit Some(512), which shadows project-level inheritance. Because the env row is seeded with Some(512) rather than None, later edits to the project-level memory (via 
GeneralSettings.tsx) won’t propagate to the production env (resolve_i32 returns the env’s own Some(512) first). Seeding the project and leaving the env None (inherit) would preserve the inheritance model the PR itself introduces. Minor UX surprise; the “Use default” button does let users opt back into inheritance.

dviejokfs:
@bherila 
- Uncapped managed containers contribute 0 to the reserved budget (documented). This means one explicitly uncapped “dedicated” workload makes admission blind to the largest consumer on the box. Acceptable by design; note it in the operator-facing error.

- Admission error leaks host total RAM to the deploying user (memory_admission_error prints total_memory_mb). I'd drop total_memory_mb from the user-facing string and keep it at debug! for operators.

bherila:
I can get this rebased/updated tonight (Pacific time)

bherila:

bherila:
Getting this rebased, changelog removed, and re-checked @dviejokfs

dviejokfs:
@bherila following up on my earlier note about `reserved_managed_memory_mb` — after another look this is a **potential blocker** unless the host-memory admission check is gated behind a feature flag (default off).

The core issue is conceptual, not just the N+1 cost: the check sums container memory **limits** and refuses to admit when the total exceeds ~80% of host RAM. But a memory limit is a cgroup *ceiling*, not a *reservation* — the container only consumes what it actually uses. Summing limits treats a ceiling as reserved memory, so e.g. ten 512 MB-limit containers idling at ~60 MB each (600 MB real usage) would block the next deploy as if the host were full. Overcommitting limits is normal and expected (same as the requests-vs-limits distinction in k8s); only **requests** should ever gate admission. On top of that the live `list_containers` + per-container `inspect_container` loop is costly and races against concurrent deploys, so it can be both expensive and still fail to prevent the overcommit it's guarding.

Two ways forward, either is fine by me:

1. **Drop the admission gate** (`reserved_managed_memory_mb` + `memory_admission_error` + the 80% budget) entirely — this also disposes of the N+1 and the race. Users can deploy past the sum of limits, which is correct behavior.
2. **Keep it but put it behind a feature flag, default off**, so operators who explicitly want hard host-capacity gating can opt in, and the default path doesn't impose it.

Either way, please **keep** the genuinely useful parts, which I'm happy with:
- seeding the 512 MB default `memory_limit` on new hosted websites (the real protection — Docker's cgroup caps any single runaway app),
- the `Some(0)` = uncapped sentinel plumbing,
- the swap cap (`memory_swap == memory`) so a "512 MB" container can't reach ~1 GiB via swap (per-container, so it doesn't conflict with the overcommit point),
- the UI (Use default / Uncapped).

Also please drop the hand-edited `## [Unreleased]` block in `CHANGELOG.md` — that file is generated by git-cliff and shouldn't be edited by hand.

dviejokfs:
Re-reviewed after the latest push — thanks @bherila, this addresses my blocker.

Confirmed resolved:
- The host-side memory **admission gate is removed** (`reserved_managed_memory_mb` / `memory_admission_error` / 80% budget all gone) — this clears both the N+1 cost and the limit-vs-reservation correctness concern. A memory *limit* is a cgroup ceiling, not a reservation, so blocking deploys on the sum of limits was the wrong model; deleting it is the right call.
- The hand-edited `## [Unreleased]` CHANGELOG block is gone.

What remains is clean and correctly scoped: seeded 512 MB default on new project + `production` env, the `Some(0)` = uncapped sentinel honored consistently across `validate`/`merge`/deployer, the swap cap (`memory_swap == memory`), and the UI. Tests cover the sentinel, env-uncapped-over-project-cap merge, and the seeded defaults. Once CI (currently re-running) goes green this is good to merge behind #170.

**Design note for any future host-memory admission work (not for this PR):** if we ever reintroduce host-capacity gating, it must be behind a **global feature flag, disabled by default, enableable globally** — never on by default. And rather than a hard "80% of physical RAM" cap, it should use an operator-configurable **overbooking multiplier** (allow sum-of-limits up to `hostRAM × N`, default generous), precisely because limits are ceilings and some overcommit is normal and desirable. Filing that as the constraint so it's captured.

PR #168, opened 2026-06-28
fix(deployer): validate inline Docker Compose overrides to prevent host escape

bherila:
## Summary
The deploy path accepted dashboard-provided `composeOverride` YAML and wrote it verbatim into `docker-compose.temps-override.yml`, letting low-privilege users specify host-affecting directives and escape the container boundary.

- Parse inline override YAML with `serde_yaml` and validate it before it is included in a `docker compose` invocation (new `ComposeError::InvalidOverride`).
- The inline override may only modify services that already exist in the base compose (cannot add services), may only contain a `services:` top-level key, and forbids host-affecting keys (privileged, namespaces, cap_add/cap_drop, devices, security_opt, sysctls, userns_mode, volumes, …).
- Unit tests cover allowed overrides and each rejected case.

This is the dashboard-supplied-override trust boundary; it is complementary to the base compose host-escape hardening (separate PR).

## CI
CI for these exact commits runs on the source PR: bherila/temps#15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   can get a local build working without trial and error.
 
 ### Fixed
--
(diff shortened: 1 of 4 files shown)


bherila:
Consolidated into #170. The inline-override structural allow-list (`validate_compose_override` + the `InvalidOverride` error, wiring, and tests) was folded into #170 as defense-in-depth alongside its value-level compose security policy. Closing in favor of #170.

PR #167, opened 2026-06-28
fix(deployer): stop deleting host containers on compose name conflicts

bherila:
## Summary
The deployer previously scanned user-controlled Compose YAML for `container_name:` lines and unconditionally stopped/removed matching Docker containers, allowing a deploy to delete unrelated host containers.

- Remove the pre-deploy `remove_conflicting_containers` call and its implementation, so untrusted `compose_content` no longer drives destructive host-wide container operations. Docker Compose surfaces genuine name collisions safely on its own.

## CI
CI for these exact commits runs on the source PR: bherila/temps#14

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #170, opened 2026-06-28
fix(deployer): harden Docker Compose deployments against host escape

bherila:
## Summary

Harden the Docker Compose executor so repository-controlled Compose YAML cannot escape the host sandbox or weaken the generated security policy.

- Parse base and override YAML with merge-key expansion and reject privileged containers, host/container namespace joins, capabilities, devices, Docker socket/API access, host-affecting service keys, unsafe build options, external/network volume drivers, and interpolation in guarded fields.
- Reject absolute and lexically escaping paths, `include:`, unsafe service names, unreviewed inline-override structure, and host-container inheritance through `volumes_from`.
- Canonicalize the selected repository directory, base Compose path, bind sources, `configs.*.file`, `secrets.*.file`, local-driver bind devices, build contexts, and Dockerfiles. Every resolved host path must already exist inside the project checkout, closing committed-symlink escapes such as `data -> /`.
- Refuse symlinked Compose, `.env`, `.env.temps`, and generated override write destinations so a malicious checkout cannot turn deployment into a host-file overwrite.
- Reject alternate OCI `runtime`, `oom_kill_disable`, `shm_size`, `tmpfs`, and `ulimits`. The resource fields are denied rather than capped per value because services, replicas, and mounts could multiply any per-value cap into unbounded aggregate host use.
- Run filesystem-aware validation before tearing down the prior stack, then repeat it immediately before `docker compose build/up`.
- Apply the generated `docker-compose.temps-security.yml` last with `privileged: false`, `cap_drop: ALL`, `no-new-privileges:true`, a PID limit, and init enabled.

## Consolidation note

This PR consolidates the overlapping Compose-hardening proposals from bherila/temps#9, bherila/temps#10, and bherila/temps#12, plus the container-name conflict fix from upstream #167.

## Landing order

#164 is stacked on this branch. Merge this PR with a merge commit so ancestry is preserved; otherwise #164 must be rebased once more onto the resulting `main`.

## Validation

- `cargo check -p temps-deployer -p temps-deployments --lib`
- `cargo clippy -p temps-deployer -p temps-deployments --lib -- -D warnings`
- 49/49 focused `temps-deployer` Compose tests
- 4/4 focused `temps-deployments` path-confinement tests

The manual `CHANGELOG.md` and unrelated cargo-audit ignore-list edits are intentionally absent; release notes are generated from conventional commits.


bherila:
Still working on this, PTAL at the other PRs and I will rebase this one later.

bherila:
Folded in the two sibling compose-hardening PRs so the host-escape protections are reviewable as one change:

- **#167** (cherry-picked): stop deleting host containers on Compose `container_name` conflicts.
- **#168** (folded in): structural allow-list for **inline overrides** — an override may only modify services already present in the repo compose file, may not add top-level keys other than `services`, and may not use host-affecting service keys (`privileged`, `network_mode`, `pid`/`ipc`/`uts`/`cgroup`/`userns_mode`, `cap_add`/`cap_drop`, `devices`, `device_cgroup_rules`, `security_opt`, `sysctls`, `volumes`). Runs as defense-in-depth alongside the value-level policy, with tests. Both #167 and #168 are now closed in favor of this PR.

Verified locally: `cargo check -p temps-deployer --tests` and `cargo fmt --check` clean.

bherila:

dviejokfs:
Hi @bherila — thanks for consolidating the three Compose-hardening proposals into this PR; the validator design is solid (merge-key expansion, interpolation guards, socket-alias normalization, and the last-in-\`-f\`-order security override are all done right).

A review pass surfaced a few gaps in the allowlist-by-omission approach. I've handled the **two blocking ones** for you and pushed them, along with a fresh merge of \`main\`, to a companion branch: **\`fix/compose-host-escape-blocking\`** (based on this PR's head + \`origin/main\`).

**Blocking fixes applied on that branch:**
1. **\`volumes_from\` was unvalidated** everywhere — not in the base service loop, \`INTERPOLATION_GUARDED_FIELDS\`, or the override \`FORBIDDEN_SERVICE_KEYS\`. \`volumes_from: container:<name>\` inherits every volume of an arbitrary host container (another tenant's, or Temps infra) without naming a path. Now rejected in all three places, with single-field tests.
2. **Absolute \`/tmp/\` bind mounts were allowed** — \`is_dangerous_host_path\` exempted anything under \`/tmp/\`, which is world-writable and can hold another project's data-dir artifacts (\`.env.temps\`, encryption keys). Now all absolute host bind sources are rejected; the stale \`is_dangerous_host_path("/tmp/ok")\` test assertion is flipped accordingly.

\`cargo check\`/\`clippy\`/\`fmt\` clean on \`temps-deployer\`; all 38 compose tests pass including the 3 new ones.

**Merge note:** the only conflict with current \`main\` was \`node_pki.rs\` (the rcgen 0.14 rename already landed via #250) — resolved by taking main's version. The big diff-vs-main was just staleness; this branch is now current.

**Still open (majors, not yet fixed — your call how to fold in):**
- \`sysctls\`, \`cgroup_parent\`, \`group_add\` guarded in the override path but not the base validator
- \`build.context\` / \`build.dockerfile\` / \`build.ssh\` unvalidated (host dir → build context; SSH-agent forward)
- \`deploy.resources.reservations.devices\` bypasses the \`gpus:\` block (Compose treats them as equivalent)
- named-volume \`driver: nfs/cifs\` not checked (only \`driver_opts.device\`)
- service names \`format!\`-injected into the generated override YAML without a charset check

Happy to either push those onto the same branch or hand them back to you — whichever you prefer. Once we agree on the approach we can fold \`fix/compose-host-escape-blocking\` back into this PR (or supersede it) and re-run CI on the rebased head.

dviejokfs:
Update: rather than a companion branch, I've pushed the merge + both blocking fixes **directly onto this PR branch** (\`fix/compose-host-escape-upstream-pr\` is now at \`693dbd27\`, and the PR shows MERGEABLE). The separate \`fix/compose-host-escape-blocking\` branch I mentioned above has been deleted — everything lives on this PR now.

CI will re-run on the updated head. The two blocking fixes (\`volumes_from\` + absolute-\`/tmp\`-bind) and the merge-conflict resolution are all here; the five **major** findings listed above are still open for a follow-up pass.

dviejokfs:
Pushed the **five major findings** onto this branch as well (\`7f95e57d\`), so the review is fully addressed in one PR:

- \`sysctls\` / \`group_add\` / \`cgroup_parent\` now rejected in the base validator (were override-only)
- \`deploy.resources.reservations.devices\` rejected — the long form of the already-blocked \`gpus:\`
- \`build.context\` / \`build.dockerfile\` path-confined; \`build.ssh\` rejected
- named volumes with a non-\`local\` driver or a network filesystem \`type\` (nfs/cifs/…) rejected
- service names validated against the Compose charset so they can't corrupt the generated security override

Each has a single-field rejection test. **43 compose tests pass**, clippy + fmt clean on \`temps-deployer\`. Together with the earlier blocking fixes and the \`main\` merge, this PR should now be complete — over to CI and a final security-auditor pass.

bherila:
Thanks @dviejokfs , pulling this now and will build and upgrade temps on my server which has some external docker compose (Fyi, I run it on the same host that's already running Sentry, which is a pretty complicated external deployment)

dviejokfs:
@bherila this is close, but a few things before it can merge:

1. **Split out the RUSTSEC ignore-list changes.** This PR adds six advisories to the `cargo-audit` ignore list in `.github/workflows/dependency-scan.yml` (daemonize, derivative, gcc, number_prefix, paste, proc-macro-error2). None of these relate to compose hardening and none are on `main` yet — please move them to their own PR so the security review here isn't approving audit suppressions as a side effect.

2. **Security sign-off required.** This is a security-sensitive change, so per our process it needs security-auditor sign-off before merge. That review is in progress; I'll relay the verdict here.

3. **Merge with a merge commit, not squash.** #164 is stacked on this branch and its ancestry needs to be preserved. If this gets squash-merged, #164 will need a rebase onto the resulting `main` before it can go in.

The hardening itself looks thorough and CI is green — just want the audit-suppression separated and the sign-off recorded first.

dviejokfs:
### Security-auditor sign-off: **BLOCK** (1 HIGH must fix; rest are follow-ups)

The compose hardening is genuinely thorough — interpolation guard, `-f` ordering, merge-key expansion, non-string service names, `include:` rejection, and validated-==-executed content all check out. But one host-escape bypass remains.

**[HIGH] Path confinement is lexical-only → symlink bypass** (`is_dangerous_host_path` / `validate_service_volumes` / `validate_top_level_files`)
The bind-source checks are pure string checks that never touch the filesystem. A deployed repo can commit a symlink (git preserves symlinks) like `data -> /`, then reference `./data` as a bind source. `is_dangerous_host_path("./data")` passes (relative, no `..`), but at runtime `docker compose up` runs in the repo checkout (`current_dir`), Docker follows the link, and mounts `/` into the container — exposing other tenants' `.env.temps`, the encryption key, and host secrets. `cap_drop:ALL`/`no-new-privileges` don't stop reads of a mounted FS. Same gap for `configs.*.file`/`secrets.*.file`.
**Fix:** after writing the compose files and before `docker compose up`, `canonicalize()` each bind source and confirm it resolves inside the project dir — or ban host bind mounts entirely and require named/anonymous volumes. Keep the lexical check as a first pass, not the only guard.

Follow-ups (don't block, but worth tracking):
- **[MEDIUM]** `runtime:` not rejected — on hosts with an alternate OCI runtime installed, a compose file can select it and sidestep the override. Add `runtime` to `reject_present`.
- **[LOW]** `oom_kill_disable`, `shm_size`, `tmpfs`, `ulimits` unbounded → single-tenant host DoS. Reject `oom_kill_disable`; bound the others.
- **[LOW]** `validate_compose_override` parses without `apply_merge()`, so a `<<: *anchor` importing a forbidden key is only caught by the value-level policy that runs first (defense-in-depth weakened, not a standalone bypass). Call `apply_merge()` in `parse_compose_yaml`.

Once the symlink bypass is fixed and verified, re-request sign-off.

dviejokfs:
### Security-auditor sign-off: **APPROVED FOR MERGE** ✅

Re-ran the gate against the actual branch code (worktree at head `45879119`) — the earlier BLOCK was against a stale diff. All prior findings are confirmed remediated:

- **[HIGH] symlink confinement** — closed. `validate_compose_filesystem_confinement` + `canonicalize_confined_existing_path` enforce `std::fs::canonicalize` + `starts_with(canonical_root)` across bind sources, named-volume `driver_opts.device`, `configs`/`secrets .file`, and `build.context`/`dockerfile`; `confined_write_path` refuses to traverse committed symlinks on every Temps-written file and its parent dirs; fail-closed on non-existent paths. Live symlink→`/` and →`/etc/passwd` tests included.
- **[MEDIUM] `runtime:`** — closed in all three layers (value policy + interpolation guard + inline-override blocklist), merge-key expansion covered.
- **[LOW] `oom_kill_disable`/`shm_size`/`tmpfs`/`ulimits`** — closed across five enforcement layers including `build.*` and long-form `type: tmpfs`.

Previously-cleared invariants still hold: braceless/`$$` interpolation coverage, security override applied last in `-f` order, `apply_merge` before validation, validated==executed content, preflight-before-teardown.

Four residual items are all LOW/INFO and non-blocking (`use_api_socket`/`gpus`/`extends` not in the inline-override key list but caught by the value-level policy; intra-project `pid: service:` sharing; `volumes.external:true`; sync FS calls in the async write path) — worth a follow-up but not a merge blocker.

Merging.

PR #172, opened 2026-06-29
feat(deployments): per-project Docker image retention with nightly pruning

bherila:
## Summary

Built deployment images accumulate indefinitely because ordinary Docker cleanup cannot prune tagged `temps-*` deployment images. This adds configurable global and per-project retention, while preserving rollback and live-service safety.

## Behavior and safety

- Adds migration `m20260816_000001_add_image_retention_hours` and exposes `image_retention_hours` through project responses and `PATCH /projects/{id}/settings`.
- Adds global `image_retention.enabled` and `image_retention.default_hours` settings. The default is 336 hours (14 days); project overrides accept 1–8760 hours, explicit `null` resets to the global default, and omission leaves the value unchanged.
- Keeps the newest five deployment records per project/environment regardless of age.
- Removes at most 500 candidates per nightly pass, oldest first.
- Removes an image only when every deployment reference is expired, preserving images reused by newer rollbacks or promotions.
- Protects currently active images, images belonging to remote nodes, uploaded/external images that cannot be rebuilt, and non-Temps registry tags.
- Uses batched, non-forced Docker removal, so Docker retains any image still referenced by a running or stopped container as a final safety net.
- Regenerates the canonical OpenAPI document and both CLI/web clients from the live unified API.

## Validation

- `cargo check --workspace --all-targets`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- Focused Docker cleanup suite: 20 passed, including real-Docker removal/retention coverage
- Project PATCH omitted/null/value regression test passed
- Disposable TimescaleDB startup applied all migrations; `projects.image_retention_hours` verified as `integer`
- Canonical OpenAPI regeneration: 707 paths
- CLI TypeScript check passed
- Web React-version check and TypeScript check passed
- Web tests: 280 passed
- Web production build passed (existing CSS/import-meta warnings only)

GitHub Actions is running on the rebased head.


dviejokfs:
@bherila I reviewed this and pushed the fixes to your branch (`28101f8d`) rather than leaving you a long list — shout if you'd rather I'd only commented. The conflicts with `main` are also resolved (merge, not rebase, so your commits are untouched); PR is `MERGEABLE` again.

The design was sound — AND-ing eligibility across every reference, restricting to local `temps-` tags, non-forced removal, and the omitted/null/value PATCH semantics were all right. Three things would have caused real damage in production though.

### Data loss (the important one)

`is_temps_managed_image` matched any `temps-`-prefixed local tag. Uploaded images are tagged `temps-{slug}-{env}:upload-{ts}` by the upload endpoint and the tarball isn't kept, so after the retention window the image was deleted with **no source to rebuild it from** — rollback and promotion both hard-fail with "image no longer exists locally" (`services.rs:1408`) and there's no way back. Now matched on deployment provenance (`context_vars.trigger`, `metadata.externalImageRef/Id`, non-git `source_type`) rather than tag text, since the upload endpoint accepts a caller-supplied `tag`.

Also protected: whatever each environment is currently serving (`environments.current_deployment_id`), regardless of age. Non-forced removal only protects an image while a container object still exists, so it isn't sufficient on its own. And images whose containers are on a worker node — this pass only talks to the local Docker daemon, so on a multi-node cluster it was trying to delete tags it doesn't have while worker disks grew unchecked.

The protection queries now abort the pass on error rather than failing open. A skipped night costs disk; a wrong deletion costs someone their deployment.

### The 48h default

Since rollback needs the image, retention *is* the rollback window. At 48h any project that didn't deploy over a long weekend silently lost the ability to roll back to anything. I moved the default to **336h (14 days)**. That's a judgement call on my side — easy to change, and it's now an operator setting rather than a constant.

### Unbounded query

`find_also_related(projects)` with `.all()` materialised every deployment row ever created, including `deployment_config`, `context_vars`, `commit_json` and `metadata`, plus a duplicated project row each. Now a 4-column partial select with retention overrides fetched separately into a map.

### Also in the push

- **`AppSettings.image_retention`** (`enabled` + `default_hours`, clamped to 1..=8760) — `with_default_image_retention_hours` was never called from `plugin.rs`, so the "system default" was a recompile-only constant. Settings row rather than env var, per CLAUDE.md.
- **UI + CLI.** The feature was API-only. Added a card in project settings and `temps projects settings --image-retention-hours / --reset-image-retention`, both warning below 48h. Argument validation runs before the network call.
- **Removed vs retained counts.** A run where all 400 removals were refused logged `"No expired deployment images to remove"` — reads as healthy when nothing is happening.
- **Audit.** `image_retention_hours` added to `ProjectSettingsUpdatedFields`; shortening retention is destructive and should be attributable.
- **Migration redated** `20260629` → `20260803`. It sorted into the middle of already-applied history and collided with `m20260629_000001_otel_metrics_full_fidelity`. It now applies last (verified: `163/163`).
- One Docker connection per batch instead of per image; `skip_serializing_if` on the double-`Option` field; `needs_preview_update` → `needs_project_row_update`.

### Evidence

`cargo check --workspace --all-targets` clean, `cargo clippy --workspace --all-targets -- -D warnings` exit 0, `cargo fmt --check` clean, `tsc --noEmit` clean in `web/`. Tests: 13 in `docker_cleanup_service` (all pass), plus 528/283/74/58 in `temps-deployments`/`temps-core`/`temps-config`/`temps-projects`.

Migration against a real Postgres:
```
→ [163/163] m20260803_000001_add_image_retention_hours … ✓ (11ms)
✓ 163 migration(s) applied in 2.18s.
projects.image_retention_hours | integer |
```

The retention rule is only as good as the non-forced removal underneath it, and that lives in bollard, so there's now a Docker-backed test (skips gracefully, no `#[ignore]`) that builds two real images, holds one with a container, and asserts the unreferenced one is actually gone while the in-use one is refused:
```
test test_real_docker_removes_unused_and_retains_in_use_image ... ok
```
Worth noting what that test caught: my first version tagged both images off a shared busybox base and the "in use" assertion **failed** — `remove_image` on a tag sharing an image ID with another tag just untags it without consulting container references. Temps builds one unique `temps-{slug}:{deployment_id}` tag per deployment so the shipping path is fine, but it's a sharp edge worth knowing about.

Two things I could not verify and am not claiming:
- `down()` was reviewed but not executed — `temps migrate` has no down command.
- The SDK deltas in `web/src/api/client/types.gen.ts` and `apps/temps-cli/src/api/types.gen.ts` were applied **by hand**, not by a real regen (that needs a running server + minted key). They typecheck, but please confirm with a full `bun run openapi-ts` before merge.

### On the "what else should be pruned" question

`prune_builder_cache` already exists (7d), as do dangling-image prune, stale static chunks, CAS blob GC, and TimescaleDB retention on `proxy_logs`/`otel_*`/`events`/`status_checks`. Real gaps, roughly by disk impact:

1. **Stopped containers** — no `prune_containers` anywhere in the workspace. Every failed deploy and replaced container leaves an exited container holding its writable layer *and* pinning its image. This is also why non-forced removals here will often be refused, so it's arguably a prerequisite for image retention reclaiming anything.
2. **Orphaned volumes** — ~20 `remove_volume` call sites, all on explicit-delete happy paths. Nothing reconciles volumes whose owning service/sandbox row is gone. Should reconcile against the DB, never `prune_volumes`.
3. **`<data_dir>/backups/tmp`** — staging dir for dumps, no reaper. A backup that dies mid-dump leaves a full database dump forever.
4. **`/tmp` transfer artifacts** — `temps-image-{uuid}.tar`, `temps-static-*`, `temps-bundle-*`, `temps-sourcemaps-*`. Cleaned on success, leaked on panic/restart; the image tarballs are gigabytes.
5. **Pipeline/deployment logs** — `pipeline_logs_dir()` has no age cap.
6. **Unused networks** — small on disk, but per-preview-env networks hit the bridge-network limit and then *new deployments fail*.

I'd suggest doing those as a follow-up that restructures `perform_cleanup` into a list of named reapers each returning `(items, bytes, errors)` with its own age setting, rather than bolting them on one at a time. Happy to open that issue if useful. I deliberately kept this PR to the image-retention scope.

dviejokfs:
@bherila heads-up — `main` moved while CI was running and PR #546 landed `m20260803_000001_add_template_slug_to_projects`, colliding with my redated migration on the same `20260803_000001` prefix. Merged main again and renumbered mine to `m20260803_000002_add_image_retention_hours` so the ordering is unambiguous rather than resolved by module-name sort.

Re-verified against a real Postgres:
```
→ [163/164] m20260803_000001_add_template_slug_to_projects … ✓ (5ms)
→ [164/164] m20260803_000002_add_image_retention_hours … ✓ (6ms)
✓ 164 migration(s) applied in 2.71s.
```
`cargo fmt --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` exit 0, 13/13 cleanup tests pass. PR is `MERGEABLE` again on `d5f88c93`; CI re-running.

The full check matrix was green on the previous head (`28101f8d`) — all unit shards, all integration suites including `migrations` and `docker-deployments`, MariaDB PITR E2E, Web TypeScript. The only delta since then is the merge and the migration renumber.

dviejokfs:
Reviewed this and pushed a follow-up commit (`c76b1a44b`) addressing the blocking/major/minor findings from review — summary for @bherila:

**Merge conflicts (blocking)** — the branch had drifted from `main` (rebase-ordering conflicts in the migration registry and a test-block adjacency in `handlers/types.rs`, plus pure re-export churn in the two generated client `index.ts` files). Merged current `main` in with a merge commit (no history rewritten) and resolved these; branch is now `CLEAN`/`MERGEABLE`.

**Unbounded query (major)** — `prune_old_deployment_images` was loading every deployment row with a non-null `image_name` into a `Vec` on every nightly run — full history, unbounded. Replaced with three targeted queries that push the actual work into Postgres:
- `expired_image_candidates` — per-image eligibility (`BOOL_AND` across every reference, so a newer rollback/promotion still protects an image) and oldest-reference, computed via `GROUP BY` so only distinct image names cross into the process.
- `recently_protected_image_names` — the keep-newest-N-per-project+environment floor, via a window function, bounded by (project, environment) pairs × N.
- `actively_served_image_names` — whatever each environment is currently serving, via a direct join, bounded by environment count.

Net effect: memory is now bounded by image/environment/project counts instead of total deployment history.

**Test coverage (major)** — added a DB-backed integration test (`test_prune_old_deployment_images_end_to_end`) exercising the full pass against a real Postgres: eligibility, the reused-image protection (old+new reference to the same tag), unrebuildable/upload protection, active-serving protection, and non-Temps-image exclusion, all in one seeded scenario. Also added a test asserting the per-project `image_retention_hours` PATCH rejects out-of-range values (0, 8761) rather than clamping — that's intentionally different from the global setting's clamp behavior, and only the clamp path had coverage before.

**Stale docs (minor)** — `image_retention_hours`'s doc comments said "NULL means use the system default (48 hours)"; the actual default is 336h (14 days), raised earlier in this PR's own history specifically because 48h was destroying rollback windows. Fixed both occurrences.

**Audit trail (minor/security)** — the audit log for `image_retention_hours` changes only recorded the new value. Since shortening this window is a permanently destructive action (deleted images can't be un-deleted), added `previous_image_retention_hours` so an incident review can see what was lost. Verified end-to-end against a live server: old→new value both land correctly in `audit_logs`.

Everything above is covered by `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings` (bypassing the rtk proxy per repo convention), `cargo check --workspace --all-targets`, and the full `temps-deployments`/`temps-projects` test suites — all green. Also manually verified the new "Built Image Retention" settings card against a live local server (screenshot + a real save round-trip confirmed via the API and the audit log), and the CLI/web typechecks pass.

CI is running on the new head.

PR #188, opened 2026-07-03
feat(analytics-sdk): exclude sensitive paths from session replay by default

bherila:
## Summary

Session replay (`@temps-sdk/analytics-core`'s `SessionRecorder`) is opt-in and already masks password/email inputs by default, but page-level exclusion (`excludedPaths`) defaulted to an empty array. An integrator who enabled replay would record every page — including login, checkout, and payment flows — unless they remembered to configure exclusions manually.

This PR ships a safer default: a built-in list of sensitive routes is now excluded out of the box.

Refs bherila/temps#28 (cross-repo reference, informational only).

## What changed

- Added `DEFAULT_EXCLUDED_PATHS` (exported from `constants.ts`):
  ```
  /login, /log-in, /signin, /sign-in, /logout, /log-out, /signup, /sign-up,
  /register, /checkout*, /payment*, /billing*, /reset-password*,
  /forgot-password*, /mfa*, /2fa*, /verify*
  ```
  Matching is against `window.location.pathname` only (case-sensitive, `*` → `.*`, anchored at both ends) — the same hand-rolled matcher `SessionRecorder` already used, unchanged.
- **Merge behavior**: user-supplied `excludedPaths` are merged additively with `DEFAULT_EXCLUDED_PATHS` (defaults + user paths), not replaced. This follows the same "merge defaults with user overrides" convention already used in this file for `samplingConfig` (spread defaults, then apply user values), rather than the full-replace (`||`) convention used for scalar options like `maskInputOptions`/`slimDOMOptions` — since dropping a security default silently on override felt like the wrong choice here.
- **Opt-out**: added `useDefaultExcludedPaths?: boolean` (default `true`) to `SessionRecordingConfig`. Setting it to `false` disables the built-in defaults entirely, so only the paths explicitly passed in `excludedPaths` are excluded (or none, if `excludedPaths` is also omitted).
- Deduplicated the two copies of the path-matching regex logic (`shouldRecord()` and `checkPath()`) into a single `matchesAnyPath()` helper.
- Bumped `@temps-sdk/analytics-core` from `0.0.1` → `0.0.2` (patch bump — no CHANGELOG convention exists anywhere in `sdks/node`, and other packages in this workspace bump patch versions for both features and fixes at this pre-1.0 stage).

## Behavior change note for existing integrators

If you already use `SessionRecorder`/`useSessionRecording` with your own `excludedPaths`, more paths will now be excluded than before (the defaults are added on top of your list). If you need the old exact behavior (only your explicit paths excluded), pass `useDefaultExcludedPaths: false`.

## Test plan

- [x] Added `src/SessionRecorder.test.ts` covering:
  - Default sensitive paths are excluded with zero configuration
  - A non-sensitive path is still recorded normally
  - User-supplied `excludedPaths` are additive (defaults still apply)
  - `useDefaultExcludedPaths: false` disables the built-in defaults (only user paths excluded)
  - `useDefaultExcludedPaths: false` with no `excludedPaths` records everything (full opt-out works)
- [x] `bun run test:run` in `sdks/node/packages/analytics-core` — 6/6 passed
- [x] `tsc --noEmit` — zero type errors
- [x] `bun run build` — builds cleanly

https://claude.ai/code/session_01MmJbBkcFSnk9ER6WA66gF2

### DIFF
--- a/sdks/node/packages/analytics-core/package.json
+++ b/sdks/node/packages/analytics-core/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@temps-sdk/analytics-core",
-	"version": "0.0.1",
+	"version": "0.0.2",
 	"description": "Framework-agnostic core for Temps analytics, engagement, Web Vitals, and session replay.",
 	"private": false,
 	"type": "module",
--- a/sdks/node/packages/analytics-core/src/SessionRecorder.test.ts
+++ b/sdks/node/packages/analytics-core/src/SessionRecorder.test.ts
@@ -1,0 +1,112 @@
+import { describe, it, expect, afterEach } from "vitest";
+import { SessionRecorder, type SessionRecorderOptions } from "./SessionRecorder";
+import { DEFAULT_EXCLUDED_PATHS } from "./constants";
+
+/**
+ * `shouldRecord()` is private, but it is the single source of truth for
+ * whether a given pathname is recorded. We flip `enabled` on directly
+ * (bypassing `start()`, which would spin up rrweb/network calls) so these
+ * tests exercise the real exclusion logic without needing to mock rrweb or
+ * fetch.
+ */
+function shouldRecordAt(path: string, options: SessionRecorderOptions = {}): boolean {
+  const recorder = new SessionRecorder({ ...options, enabled: false });
+  const internal = recorder as unknown as { enabled: boolean; shouldRecord(): boolean };
+  internal.enabled = true;
+  window.history.pushState({}, "", path);
+  return internal.shouldRecord();
+}
+
+describe("SessionRecorder excludedPaths defaults", () => {
+  afterEach(() => {
+    window.history.pushState({}, "", "/");
+  });
+
+  it("ships the documented list of default excluded paths", () => {
+    expect(DEFAULT_EXCLUDED_PATHS).toEqual([
+      "/login",
+      "/log-in",
+      "/signin",
+      "/sign-in",
+      "/logout",
+      "/log-out",
+      "/signup",
+      "/sign-up",
+      "/register",
+      "/checkout*",
+      "/payment*",
+      "/billing*",
+      "/reset-password*",
+      "/forgot-password*",
+      "/mfa*",
+      "/2fa*",
+      "/verify*",
+    ]);
+  });
+
+  it("excludes sensitive routes out of the box with zero configuration", () => {
+    const sensitivePaths = [
+      "/login",
+      "/log-in",
+      "/signin",
+      "/sign-in",
+      "/logout",
+      "/signup",
+      "/sign-up",
+      "/register",
+      "/checkout",
+      "/checkout/step-2",
+      "/payment/confirm",
+      "/billing",
+      "/billing/invoices",
+      "/reset-password",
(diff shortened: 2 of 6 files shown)


dviejokfs:
Heads up: this fix only reaches `@temps-sdk/analytics-core` (and transitively `vue-analytics`, `svelte-analytics`, and `analytics-browser`, which all import from `analytics-core`).

`@temps-sdk/react-analytics` doesn't pick this up. `sdks/node/packages/react-analytics/src/SessionRecorder.tsx` is a fully separate, hand-duplicated implementation — it imports `record`/`pack` from `rrweb` directly, not from `@temps-sdk/analytics-core` — and has its own three copies of the same `excludedPaths.some(...)` regex matcher, still defaulting to `[]`. `Provider.tsx` renders this local component (`./SessionRecorder`), so React integrators (likely the most common integration path) get no benefit from `DEFAULT_EXCLUDED_PATHS` / `useDefaultExcludedPaths` — they'll still record `/login`, `/checkout`, etc. by default after this ships.

Confirmed by checking imports: `vue-analytics`/`svelte-analytics` do `export * from "@temps-sdk/analytics-core"` and `analytics-browser` imports `SessionRecorder` from it directly, so those three inherit the fix for free. `react-analytics` is the outlier.

Suggest having `react-analytics/SessionRecorder.tsx` delegate to (or at least reuse the constants/flag from) `analytics-core`'s implementation instead of maintaining a fourth independent copy of this logic — otherwise the safer default only covers 3 of 4 framework packages, and the PR title/description overclaim what's actually fixed.

dviejokfs:
Pushed a fix for react-analytics in f4d437f3.

`react-analytics/SessionRecorder.tsx` now imports `DEFAULT_EXCLUDED_PATHS` and `matchesAnyPath` from `@temps-sdk/analytics-core` (already a dependency) instead of maintaining its own copy of the exclusion logic, and gains a `useDefaultExcludedPaths` prop with the same merge-by-default / opt-out-via-`false` semantics as `analytics-core`. `Provider.tsx` forwards `sessionRecordingConfig.useDefaultExcludedPaths` through. `matchesAnyPath` is now exported from `analytics-core` so all four framework packages (`analytics-core`, `react-analytics`, and the two that re-export it) share one wildcard matcher instead of four independent copies.

Added a `default excluded paths` test block to `react-analytics`'s `SessionRecorder.test.tsx` mirroring `analytics-core`'s coverage (defaults excluded, non-sensitive path still recorded, additive merge with user-supplied paths, opt-out via `useDefaultExcludedPaths: false`, full opt-out with no `excludedPaths`).

Bumped `@temps-sdk/react-analytics` 0.0.3 → 0.0.4 (patch, behavior change).

Verified:
- `tsc --noEmit` clean in both packages
- `bun run build` clean in both packages
- New tests pass (5/5)
- Diffed the full react-analytics test suite before/after: identical set of 21 pre-existing failures (unrelated flaky tests in `useSessionRecording`/`useTrackEvent`/`utils`/timing-sensitive retry tests) — no regressions introduced

All four framework SDKs (`analytics-core`, `react-analytics`, `vue-analytics`, `svelte-analytics`, `analytics-browser`) now ship the safer session-replay default.

PR #189, opened 2026-07-03
feat(auth): audit concurrent sessions and allow requiring MFA for admins

bherila:
## What

- Adds `AuthService::count_active_sessions` and emits a `ConcurrentSessionDetectedAudit` audit entry when a login occurs while another session for the same account is already active. Purely observational — never blocks the login, and a lookup failure degrades gracefully (logged, not fatal).
- Adds a `require_mfa_for_admins` app setting (default `false`, DB-backed per the project's runtime-config convention, not an env var). When enabled, `Admin`-role users without MFA enrolled are rejected on the **password**-login path with a typed `MfaRequiredForRole` error (403). SSO/OIDC logins are a separate code path and are intentionally unaffected — federating to a hardened IdP is an acceptable alternative to local TOTP MFA.

## Why

Closes part of the identity/access-control gaps tracked in bherila/temps#24 and bherila/temps#32 (a self-hosting hardening pass, not specific to any one company's compliance program — same fixes apply to anyone running Temps for a production app).

Note: the DB-level unique constraint on `users.email` (the other half of #24) already shipped upstream via d24006c4 — this PR only adds the concurrent-session audit visibility on top of that.

## Testing

- `cargo check --lib -p temps-auth`, `-p temps-core`: clean, zero warnings.
- `cargo test --lib -p temps-auth`: 8 new tests added for this change, all passing (`count_active_sessions_*`, `login_allows_mfa_enrolled_admin_when_required`, `login_blocks_admin_without_mfa_when_required`, `login_allows_admin_without_mfa_when_setting_disabled`, `login_never_blocks_non_admin_without_mfa_when_required`, `register_user_maps_concurrent_duplicate_insert_to_email_already_registered`).
- `cargo test --lib -p temps-core`: 3 new tests for the settings default/round-trip/backward-compat behavior, all passing.
- Note: 66 pre-existing `temps-auth` tests fail in the sandbox this was developed in because that sandbox has no Docker daemon reachable at all (these tests use `TestDatabase`/testcontainers, which CI's `ubuntu-latest` runners support natively per `rust-tests.yml`'s unit-tests job). Verified this is a sandbox-only limitation, not a regression: none of the new tests above depend on Docker, and the 66 failures are 100% pre-existing tests unrelated to this change.

### DIFF
--- a/crates/temps-auth/src/audit.rs
+++ b/crates/temps-auth/src/audit.rs
@@ -10,6 +10,20 @@ pub struct LoginAudit {
     pub login_method: String,
 }
 
+/// Recorded when a user completes a successful login while at least one
+/// other session for their account is already active (non-expired). This is
+/// purely observational -- it does NOT block the login (bherila/temps#24) --
+/// but gives an auditor/operator a trail to spot suspicious concurrent
+/// access (e.g. a stolen session token being used from a second location).
+#[derive(Debug, Clone, Serialize)]
+pub struct ConcurrentSessionDetectedAudit {
+    pub context: AuditContext,
+    pub login_method: String,
+    /// Number of other active sessions that already existed for this user
+    /// at the moment this login completed (not counting the new session).
+    pub existing_active_session_count: u64,
+}
+
 // User management audits
 #[derive(Debug, Clone, Serialize)]
 pub struct UserCreatedAudit {
@@ -198,6 +212,29 @@ impl AuditOperation for LoginAudit {
     fn user_agent(&self) -> &str {
         &self.context.user_agent
     }
+    fn serialize(&self) -> Result<String> {
+        serde_json::to_string(self)
+            .map_err(|e| anyhow::anyhow!("Failed to serialize audit operation {}", e))
+    }
+}
+
+impl AuditOperation for ConcurrentSessionDetectedAudit {
+    fn operation_type(&self) -> String {
+        "CONCURRENT_SESSION_DETECTED".to_string()
+    }
+
+    fn user_id(&self) -> i32 {
+        self.context.user_id
+    }
+
+    fn ip_address(&self) -> Option<String> {
+        self.context.ip_address.clone()
+    }
+
+    fn user_agent(&self) -> &str {
+        &self.context.user_agent
+    }
+
     fn serialize(&self) -> Result<String> {
         serde_json::to_string(self)
             .map_err(|e| anyhow::anyhow!("Failed to serialize audit operation {}", e))
--- a/crates/temps-auth/src/auth_service.rs
+++ b/crates/temps-auth/src/auth_service.rs
@@ -951,8 +951,8 @@ mod tests {
     use axum::http::HeaderMap;
     use chrono::{Duration, Utc};
     use temps_database::test_utils::TestDatabase;
-    use temps_entities::{magic_link_tokens, sessions, settings, users};
     use temps_entities::types::RoleType;
+    use temps_entities::{magic_link_tokens, sessions, settings, users};
 
     struct MockEmailService {
         verification_emails_sent: std::sync::Mutex<Vec<(String, String, String)>>,
@@ -2066,10 +2066,7 @@ mod tests {
             })
             .await;
 
-        assert!(matches!(
-            result,
-            Err(UserAuthError::EmailAlreadyRegistered)
-        ));
+        assert!(matches!(result, Err(UserAuthError::EmailAlreadyRegistered)));
     }
 
     // --- Concurrent-session detection (bherila/temps#24) ---
@@ -2119,11 +2116,17 @@ mod tests {
         let other_user = create_test_user(&db.db, "other@example.com", "Password123!").await;
 
         // No sessions yet.
-        assert_eq!(auth_service.count_active_sessions(user.id).await.unwrap(), 0);
+        assert_eq!(
+            auth_service.count_active_sessions(user.id).await.unwrap(),
+            0
+        );
 
         // One active session for our user.
         auth_service.create_session(user.id).await.unwrap();
-        assert_eq!(auth_service.count_active_sessions(user.id).await.unwrap(), 1);
+        assert_eq!(
+            auth_service.count_active_sessions(user.id).await.unwrap(),
+            1
+        );
 
         // An expired session for our user must not be counted.
         let expired_session = sessions::ActiveModel {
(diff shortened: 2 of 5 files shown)


dviejokfs:
## Automated review: `feat(auth): audit concurrent sessions and allow requiring MFA for admins`

**Verdict: REQUEST_CHANGES** — not merged. In addition to the findings below, this PR currently shows `mergeable: CONFLICTING` against `main` (a trivial adjacent-field collision in `crates/temps-config/src/handler.rs` / `crates/temps-core/src/app_settings.rs` with the recently-merged `cluster_dns` field — a rebase resolves it, no semantic conflict), so it can't be merged as-is regardless of the code findings.

**Scope:** 5 files (all `crates/temps-auth`, `temps-config`, `temps-core`; no migrations, no frontend) · +623/-12. Thanks for the thorough PR description and the 11 new tests — the service-layer coverage on `count_active_sessions` and the MFA-gating branches is solid.

### Blocking (must fix before merge)

- **[`crates/temps-auth/src/handlers.rs`, `login()` error mapping] Credential oracle via 401 vs 403.** `MfaRequiredForRole` is only reachable *after* the Argon2 password check succeeds (`auth_service.rs` login, right after `if !password_valid { return Err(InvalidCredentials) }`). Today, `InvalidCredentials`/`UserNotFound` map to 401 and this new error maps to 403 with a distinct body. That difference lets anyone probing a known admin email learn "this password is correct" (and that the account is Admin) without ever completing a login — a binary-searchable oracle for admin credentials. Every other error this deep in the login path (see the existing `InvalidCredentials | UserNotFound => 401` arm) is deliberately collapsed to the same 401/message specifically to prevent this. → Return 401 with the same generic message for `MfaRequiredForRole` (or otherwise avoid signaling "password was correct" pre-authentication).

- **[`crates/temps-auth/src/auth_service.rs`, `login()`] Unconditional `self.get_settings().await?` can 500 every login, not just admins.** This call has no fallback, unlike `count_active_sessions` in the same PR (which falls back to `0` on error and logs — the stated "never blocks login" design). A transient failure reading the single `settings` row (pool exhaustion, timeout) now hard-fails **all** password logins, admin or not, contradicting the PR's own graceful-degradation intent. → `self.get_settings().await.unwrap_or_default()` (default already has `require_mfa_for_admins: false`, so this fails open, not closed).

### Major (should fix, same PR)

- **No guard against operator self-lockout.** Enabling `require_mfa_for_admins=true` with zero MFA-enrolled admins and no SSO configured permanently blocks all password-based admin logins in-app, with no warning and no documented recovery path (only a raw DB edit). Consider validating "at least one active Admin has MFA enrolled, or an SSO provider exists" before accepting the flag, or at minimum returning a 409 with guidance.

- **`is_unique_violation`/`From<DbErr> for UserAuthError` is broader than intended.** It matches by generic substring ("duplicate key", "23505", `RecordNotInserted`) rather than the specific `idx_users_email_unique` constraint, and applies to *every* function returning `UserAuthError` — not just `register_user`. `magic_link_tokens.token` also has a DB-level unique constraint (`m20250101_000001_initial_schema.rs`), and `send_magic_link` returns `UserAuthError`; a token collision there would currently be silently absorbed (the handler swallows all send errors), but any future `UserAuthError`-returning function touching a unique-constrained column will misreport as `EmailAlreadyRegistered`. → Match the specific constraint name instead of a generic substring.

### Minor / Nits

- Dropping `#[from]` from `DatabaseError(sea_orm::DbErr)` without adding `#[source]` back drops `Error::source()` chaining for that variant (thiserror only wires `source()` via `#[source]`/`#[from]`/a field literally named `source`). Nothing reads `.source()` on this type today, but it's a quiet regression for future error-chain tooling (Sentry grouping, etc.). Suggest `DatabaseError(#[source] sea_orm::DbErr)`.
- `count_active_sessions` counts every non-expired row in the shared `sessions` table, which also holds short-lived (5 min) MFA-challenge sessions with no discriminator column. A user who abandons/retries an MFA challenge within 5 minutes will trigger a false-positive `ConcurrentSessionDetectedAudit`. Non-blocking (it's audit-only, never blocks login) but worth a follow-up (e.g., a `session_kind` column) since it undermines the signal the audit is meant to provide.
- Handler-layer error mapping for the new `MfaRequiredForRole` arm doesn't have a unit test, even though `handlers.rs` already has an established pattern for exactly this (`test_login_error_mapping_invalid_credentials_constant_message`) — worth mirroring once the 401/403 fix above lands.

### Security

Ran an independent `security-auditor` pass; it confirmed the two blocking findings above (rating the 401/403 divergence **HIGH** and the unguarded settings read **HIGH**), plus the `magic_link_tokens.token` unique-constraint detail on the `is_unique_violation` finding. No SQL injection, command injection, or secret-exposure issues found — `count_active_sessions` uses parameterized Sea-ORM `QueryFilter`, and no new fields need masking.

### Migrations

None — `require_mfa_for_admins` is stored on the existing JSON `settings.data` blob via `#[serde(default)]`, consistent with the project's "runtime config as a settings row, not an env var" convention. Correctly backward-compatible with pre-existing rows (covered by `legacy_settings_json_without_require_mfa_for_admins_deserializes`).

### Verified clean

- Three-layer architecture respected (handler → service → data access), typed errors throughout, no `.unwrap()`/`.expect()` in production code paths.
- SSO/OIDC path is genuinely untouched — confirmed no other caller of `AuthService::login()` exists outside `handlers.rs`, so the MFA gate's blast radius is exactly the password-login endpoint as described.
- `require_mfa_for_admins` defaults to `false` and only gates password login for the `Admin` role (verified there's only `Admin`/`User` in `RoleType` — no missed privileged tier).
- Good service-layer test coverage: concurrent-session counting (incl. expiry/other-user isolation), all four MFA-gating branches (enrolled admin, blocked admin, disabled setting, non-admin never blocked), and the register-race unique-violation mapping.

Happy to re-review once the two blocking items are addressed and the branch is rebased onto current `main`.

dviejokfs:
Pushed a follow-up commit (`cd6955ae`) addressing the review findings above, and rebased (merged `main`) to clear the earlier conflict:

- **Credential oracle (blocking):** `MfaRequiredForRole` now returns the same 401 "Invalid Credentials" response as `InvalidCredentials`/`UserNotFound` instead of a distinguishable 403 — closes the admin-password-confirmation oracle. The real reason is still logged server-side (`warn!`). Added `test_login_error_mapping_mfa_required_same_as_invalid_credentials` to lock this in.
- **Settings-outage login failure (blocking):** `login()` now falls back to `AppSettings::default()` on a settings-lookup error instead of propagating it, so a transient DB hiccup on the `settings` row can't 500 out every password login on the instance. Added `login_succeeds_when_settings_lookup_fails` (MockDatabase-based).
- **Overly broad unique-violation matching (major):** `is_unique_violation` now matches the specific `idx_users_email_unique` constraint name instead of a generic `duplicate key`/`23505` substring, so a collision on an unrelated unique column (e.g. `magic_link_tokens.token`, which also has a DB-level unique constraint) can no longer be misreported as `EmailAlreadyRegistered`. Added `is_unique_violation_false_for_unrelated_unique_constraint`.
- **`Error::source()` chain (minor):** `DatabaseError` keeps `#[source]` (independent of the manual `From` impl) so error-chain tooling still sees the underlying `DbErr`.

Verified: `cargo check --lib`, `cargo test --lib` (all 20 auth/MFA/session tests pass individually, Docker-backed `TestDatabase` tests included — ran for real, not skipped), `cargo fmt --check`, and `cargo clippy --lib --tests -- -D warnings` all clean on `temps-auth`, `temps-core`, `temps-config`.

Not addressed in this pass (left as follow-ups, non-blocking):
- No guard against enabling `require_mfa_for_admins` with zero MFA-enrolled admins (self-lockout risk) — would need a cross-crate check from `temps-config`'s settings-update handler into user/role state; leaving this as a design decision for a separate PR rather than bundling it in here.
- MFA-challenge sessions inflating `count_active_sessions`' audit signal — would need a `session_kind` discriminator (schema change), also left for a follow-up.

PR #190, opened 2026-07-03
fix(providers): bind provisioned service ports to 127.0.0.1 instead of 0.0.0.0

bherila:
## What

Managed services provisioned by Temps (Postgres, Redis, MongoDB, S3/RustFS) were binding their Docker-published ports to `0.0.0.0` — publicly reachable on any host without an external firewall in front of it — while `docs/howto/set-up-managed-services` already claims "All ports are bound to `127.0.0.1` only." This closes that gap so the code actually matches the documented behavior.

Factored the binding logic into a single `crate::utils::local_port_binding` helper (with tests asserting it never returns `0.0.0.0`), used consistently by `postgres.rs`, `redis.rs`, `s3.rs`, `rustfs.rs` (both its data and console ports), and `postgres_lifecycle.rs` — a fifth call site the original audit had missed.

Also cleans up `docs/advanced/security` to remove references to a `temps firewall` CLI command and a `config/settings.yml` file, neither of which exists in this codebase (tracked as part of bherila/temps#37).

## Why

Self-hosting hardening pass (applies to anyone running Temps for a production workload, not specific to any one deployment) — tracked in bherila/temps#29 and part of bherila/temps#37.

## Testing

- `cargo check --lib -p temps-providers`: clean, zero warnings.
- `cargo test --lib -p temps-providers`: 238 passed, 0 failed (includes 2 new tests for the port-binding helper).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -37,7 +37,7 @@ When deploying Temps, please ensure:
 
 1. **Use HTTPS** — Always configure TLS certificates for your deployment.
 2. **Strong passwords** — Use strong passwords for admin accounts and database connections.
-3. **Firewall rules** — Restrict access to management ports (see `docs/server-firewall-setup.md`).
+3. **Firewall rules** — Restrict access to management ports at the network/OS level, and use the [Admin Listener](https://temps.sh/docs/admin-listener) to bind the admin/dashboard surface to a private interface with CIDR + Host allowlists.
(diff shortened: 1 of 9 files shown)


dviejokfs:
@bherila could you fix the formatting for this PR and the others that are also failing?

bherila:
yes everything will be fixed up. sorry my codespace timed out and left unfinished work

PR #191, opened 2026-07-03
feat(auth): add platform-admin role, key/token rotation, and MCP kill-switch

bherila:
## What

- **Platform-admin role** (bherila/temps#25): adds `Role::PlatformAdmin` with the full platform/user/settings-administration permission set (`SystemAdmin`, `UsersWrite`, `SettingsWrite`, etc.) but read-only on Projects/Deployments — no ability to create, modify, or delete deployable resources. Mirrors the existing `Role::User`, which has deploy permissions but no admin ones.
- **API key / deployment token rotation** (bherila/temps#33): adds a rotate operation for both — issues a fresh secret for the same record and immediately invalidates the old one, audit-logged like other write operations. Node/agent join-token rotation is explicitly out of scope for this PR (noted as a follow-up in the linked issue).
- **MCP access kill-switch** (bherila/temps#19): adds `mcp_access_enabled` (default `true`, DB-backed via `AppSettings` so it's editable at runtime without a restart, mirrored into a process-wide `AtomicBool` the same way `insecure_tls_enabled` already works in this codebase). When disabled, any API key with `Role::Mcp` is rejected with a typed 403 in the auth middleware, before the request is ever attached to a handler.

## Why

Self-hosting hardening pass — applies to anyone running Temps for a production app, not specific to any one deployment.

## Testing

- `cargo check --lib -p temps-auth`, `-p temps-deployments`, `-p temps-core`: clean, zero warnings.
- `cargo test --lib -p temps-auth`: 6 new rotation tests + 1 new MCP kill-switch test, all passing. 69 pre-existing tests fail in the sandbox this was developed in due to no Docker daemon being reachable there (same as noted in #189) — confirmed unrelated to this change; none of the new tests depend on Docker.
- `cargo test --lib -p temps-core`: new settings round-trip/default tests passing.


dviejokfs:
@bherila 
I think MCP is no longer active, it was removed in an early commit, so the permissions for MCP may be removed and the MCP kill switch can be removed too.

The recommended way of using AI is using the skills or the new AI widget that requires an AI api key (Anthropic, OpenAI, XAI or Gemini)

dviejokfs:
Hi @bherila — I pushed a commit to this branch (maintainer edit) that **merges `main`** and **removes the MCP surface entirely**. Flagging it here so you're not surprised.

### Why
Temps ships **no MCP server** for a `Role::Mcp` API key to authenticate against. `Permission::McpConnect` is defined but never enforced by any route guard, and the `temps-mcp` crate referenced in `CLAUDE.md` doesn't actually exist. So the `mcp_access_enabled` kill-switch this PR added was gating a role that enforces nothing. Rather than ship a switch for a non-existent feature, we dropped the whole MCP scaffolding. (Note: this is unrelated to the agent "MCP Servers" feature in `temps-agents` — that's the agent runtime acting as an MCP *client*, which stays.)

Bonus: the only merge conflict against `main` was in `AppSettings`/`AppSettingsResponse`, where `cluster_dns` + `require_mfa_for_admins` landed after this branch forked — exactly where `mcp_access_enabled` was being inserted. Dropping MCP resolved the conflict cleanly.

### What was removed
- The PR's `mcp_access_enabled` kill-switch: settings field, `temps_core::mcp_access` process cache (`mcp_access.rs`), the auth-middleware rejection block, and the config-service republish calls.
- The pre-existing dead `Role::Mcp` and the `McpConnect`/`McpExecute`/`McpRead`/`McpWrite` permissions (`permissions.rs`, `apikey_types.rs`, `permission_attribute.rs`) + their tests.

The rest of your PR — **platform-admin role** and **API-key / deployment-token rotation** — is untouched.

### What happens to existing API keys? (no DB migration needed)
`api_keys.role_type` is a plain `String` column and `permissions` is a JSON string array — no PG enum, so nothing to migrate. Existing rows are untouched at the DB level. At runtime:

| Existing key | Behavior after this change |
|---|---|
| Predefined **`mcp`-role** key | **Fails closed.** `Role::from_str("mcp")` now returns `None`, so `validate_api_key` errors and the middleware treats the request as unauthenticated → 401 on protected routes. These keys couldn't do anything MCP-specific anyway. |
| **Custom** key carrying `mcp:*` permissions | **Keeps working.** Unknown permission strings are silently skipped on load; the key retains all its other permissions. |
| Creating a **new** key with `mcp` role or `mcp:*` perms | Rejected by creation validation (unknown role/permission). |

### Verification
- `cargo check --lib -p temps-auth -p temps-core -p temps-config`: clean.
- `cargo test -p temps-auth -p temps-core -p temps-config` (incl. `permissions_tests` integration): **453 passed**.
- `cargo clippy --all-targets -- -D warnings` on the touched crates: clean.

bherila:
Oh thank you. Btw due to the US holiday today I have not been around too much. But I agree it can be removed!

dviejokfs:
No worries, we got a lot of gh actions going on when this goes green it will be merged

dviejokfs:
@bherila looks good for you to be merged?

bherila:
yes

bherila:
running 1 last AI review now actually

bherila:
I found one issue that keeps the new deployment-token rotation endpoint from being reachable:

`crates/temps-deployments/src/handlers/deployment_tokens.rs:89` adds `POST /projects/{project_id}/deployment-tokens/{token_id}/rotate`, but `deployment_tokens::configure_routes()` is not merged into `DeploymentsPlugin::configure_routes` (only deployments, crons, external images, remote deployments, and nodes are merged). As a result, the new endpoint will 404 in the running API and won't appear in the exposed OpenAPI schema. Please wire the deployment-token router/schema into the deployments plugin along with the new route.

dviejokfs:
> I found one issue that keeps the new deployment-token rotation endpoint from being reachable:
> 
> `crates/temps-deployments/src/handlers/deployment_tokens.rs:89` adds `POST /projects/{project_id}/deployment-tokens/{token_id}/rotate`, but `deployment_tokens::configure_routes()` is not merged into `DeploymentsPlugin::configure_routes` (only deployments, crons, external images, remote deployments, and nodes are merged). As a result, the new endpoint will 404 in the running API and won't appear in the exposed OpenAPI schema. Please wire the deployment-token router/schema into the deployments plugin along with the new route.

do you want me to tackle it?

dviejokfs:
Pushed a fix for the deployment-token endpoint wiring (`fix(deployments): wire deployment-token routes and schema into plugin`).

**Root cause** — `handlers::deployment_tokens::configure_routes()` and `DeploymentTokensApiDoc` were merged into **neither** `DeploymentsPlugin::configure_routes` **nor** `openapi_schema`. This is a **pre-existing gap on `main`** (this PR didn't touch `plugin.rs`), so it affected **every** deployment-token endpoint — create/list/get/update/delete — not just the new rotate route. They 404'd at runtime and were absent from the OpenAPI schema (confirmed: the generated web SDK had zero deployment-token operations).

**Fix** — the token router carries its own `DeploymentTokenAppState`, so I build that state (using the already-registered `DeploymentTokenService` + audit logger) and merge the router after `with_state`, plus add `DeploymentTokensApiDoc` to the schema merge. Added a regression test asserting the token paths (incl. `.../{token_id}/rotate`) now appear in the merged OpenAPI schema.

Verified: `cargo check`, `cargo test --lib -p temps-deployments` (404 passed + new test), `cargo clippy -D warnings` all clean.

bherila:
I think you did already? :) Thanks, reviewing now

dviejokfs:
Merging

PR #245, opened 2026-07-06
perf(proxy): jemalloc allocator + revert breaking sha2 bump

dviejokfs:
## Summary
- Switch `temps-cli`'s binary (runs both proxy and console) to `tikv-jemallocator` as the global allocator. The proxy's high-churn, multi-threaded per-request allocation pattern fragments badly under glibc malloc; jemalloc should meaningfully reduce RSS. Built with `profiling`/`stats` features, which are inert (zero overhead) unless enabled at process start via `_RJEM_MALLOC_CONF`.
- Revert the `sha2` 0.10.9 → 0.11.0 bump (dependabot PR #225, auto-merged as `ee38f985`), which currently breaks compilation on `main`: several crates hex-format digests with `{:x}` (relies on sha2 0.10's output type implementing `LowerHex`, which 0.11's doesn't), and `temps-proxy`/`temps-dns` pin `sha2 = "0.11"` directly, colliding with `hmac`/`rsa` (AWS Route53 signing, GCP JWT signing, password-wall HMAC check) which are still on the older `digest` major version.

Investigated capping Pingora's upstream connection pool per-host as a separate memory lever, but on reading the Pingora 0.8.0 source, `ServerConf.upstream_keepalive_pool_size` (default 128) already bounds it globally across all hosts via a single shared `Connector`/LRU — not unbounded, and not a meaningful contributor to RSS. No change made there.

## Test plan
- [x] `cargo check -p temps-cli --bin temps` — clean
- [x] `cargo check --workspace --lib` — clean (184 crates)
- [x] Pre-commit hooks (cargo fmt, cargo clippy, Conventional Commit) passed on both commits
- [ ] Measure RSS before/after under load to confirm the jemalloc delta


dviejokfs:
Superseded — everything here is now on main via #230 (and #247):

- jemalloc as the global allocator landed in 16a48f16, upgraded to tikv-jemallocator 0.7.0 / jemalloc 5.3.1 in 3a71b12a (this PR pinned 0.6), plus profiling/stats features and the aarch64 lg-page release fix.
- The sha2 0.11 revert landed on main separately in #247.

Merging this would downgrade the allocator to 0.6, and GitHub reports it as conflicting with main for exactly that overlap. Closing rather than merging; reopen if there was anything else intended to ride along.

PR #250, opened 2026-07-07
fix: repair rcgen 0.14 API breaks on main

dviejokfs:
## Summary
- `main` currently fails `cargo check --lib --workspace` and `cargo clippy --workspace --all-targets` due to leftover rcgen 0.14 API breaks that #247 (dependency-bump compilation fix) didn't cover.
- `crates/temps-core/src/node_pki.rs`: `CertificateParams::from_ca_cert_pem` was removed and `CertificateSigningRequestParams::signed_by` now takes a single `&Issuer<impl SigningKey>` instead of separate cert+key args. Rebuilt the CA issuer via `Issuer::from_ca_cert_pem(ca_cert_pem, ca_key)` and sign the CSR through it. Also `Ia5String` is no longer re-exported at the crate root — imported from `rcgen::string` instead.
- `crates/temps-proxy/src/tls_cert_loader.rs` (test helper): `rcgen::CertifiedKey`'s `key_pair` field was renamed to `signing_key` — only surfaced under `--all-targets` since it's in a `#[cfg(test)]` block.

## Test plan
- [x] `cargo check --lib --workspace` — clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean (matches pre-commit hook)
- [x] `cargo test --lib -p temps-core node_pki` — 5/5 pass, including the worker-supplied-SAN-rejection security test (`test_sign_overwrites_worker_supplied_sans`)

### DIFF
--- a/crates/temps-core/src/node_pki.rs
+++ b/crates/temps-core/src/node_pki.rs
@@ -16,9 +16,10 @@
 //! Wiring it into the agent's TLS listener, the control-plane client, and the
 //! enrollment handshake happens in the respective crates.
 
+use rcgen::string::Ia5String;
 use rcgen::{
     BasicConstraints, CertificateParams, CertificateSigningRequestParams, DistinguishedName,
-    DnType, ExtendedKeyUsagePurpose, Ia5String, IsCa, KeyPair, KeyUsagePurpose, SanType,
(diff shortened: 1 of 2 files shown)


dviejokfs:
@bherila merging this, I saw your PRs also fixed it

PR #251, opened 2026-07-07
feat(deployments): add SecretsManagerResolver OSS extension seam (ADR 0009 §1-2)

dviejokfs:
## Summary
- New `temps_core::SecretsManagerResolver` trait, modeled on the existing `ProjectEnvVarsProvider` — a generic extension point that lets an optional plugin inject deploy-time secrets without this crate depending on any external secrets-manager integration.
- Wires an optional resolver into `WorkflowPlanner` using the same two-phase slot pattern already used for `DeploymentGate`: captured in `register_services`, wired in `initialize_plugin_services` via `context.get_service()` (never `require_service`), so a build with no resolver registered is a strict no-op.
- Extends `gather_environment_variables` with a **fail-closed** step between external-service env vars and the Sentry DSN: if a registered resolver returns `Err`, the whole deployment aborts with a clear error (never silently deploys with a missing secret). Resolved *keys* are logged, values never are.

## Test plan
- [x] `cargo check --workspace --lib`
- [x] `cargo test --lib -p temps-deployments secrets_resolver` (3 new tests: fail-closed on error, success path, absent-resolver strict no-op)
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] pre-commit hooks (fmt, clippy, conventional commit) passed locally

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2561,9 +2561,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
 dependencies = [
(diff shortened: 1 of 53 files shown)


dviejokfs:
Closing — this PR was accidentally branched off an unrelated in-progress branch instead of main, dragging in unrelated commits. Reopening cleanly.

PR #253, opened 2026-07-07
ci: build test binaries once via cargo-nextest, share across all jobs

dviejokfs:
## Summary

Started as a cache-key fix, ended up as a bigger redesign once CI proved the cache-key approach didn't actually work — leaving both here for the history, but **only the final commit matters for review**.

**Root cause:** cargo's `resolver = "2"` scopes feature unification to the exact package selection of a given invocation. A `--workspace` prebuild and any narrower `-p <subset>` selection produce different, non-interchangeable fingerprints for the same shared dependency — confirmed locally by watching `serde` recompile under a new fingerprint hash from *widening* a `-p` list, with zero `--features` change involved. No cache-key scheme can fix this: as long as `unit-tests`/`integration-tests` use different `-p` selections than any shared prebuild, cargo recompiles, full stop.

**The fix:** `cargo-nextest`'s archive workflow sidesteps the problem instead of working around it.

1. One new job, `build-test-binaries`, runs `cargo nextest archive --workspace --all-targets` **twice** — once with default features, once with `temps-providers/temps-backup`'s `docker-tests` feature on — and uploads both as artifacts. Two archives, not one, because `ubuntu-latest` runners have a live Docker daemon by default: if `unit-tests` ran with `docker-tests` compiled in, the heavy Docker tests wouldn't gracefully no-op the way they do when Docker is genuinely unreachable — they'd actually execute, against none of the shared service containers `integration-tests` sets up.
2. `unit-tests`, `integration-tests`, and `mariadb-pitr-e2e` all download the relevant archive and select which tests to run via a nextest filterset expression (`-E 'package(x) or package(y) ...'`) — **no cargo invocation happens in any of these jobs**, so there's nothing to recompile, verified locally (zero `Compiling` output running a narrower selection from a wider archive).
3. All three now only depend on `build-test-binaries`, not on each other — they run in parallel. A compile failure already fails `build-test-binaries` before any of them start, so gating `integration-tests` behind `unit-tests` bought nothing once neither compiles anything.
4. `-p <crate list>` → `-E 'package(x) or package(y) ...'`; `--skip <pattern>` → `and not test(pattern)` (nextest rejects both `-p` and `--skip` alongside `--archive-file`, confirmed locally).

## Test plan

- [ ] CI on this PR: `build-test-binaries` completes once; `unit-tests` (both groups), `integration-tests` (all 4 groups), and `mariadb-pitr-e2e` start together right after and show no `Compiling` output, just test execution
- [ ] Confirm test coverage is unchanged — every `package()`/`test()` filter mirrors the exact same crate/test-name lists the old `-p`/`--skip` flags used
- [ ] Confirm `check`/`clippy`/`fmt` still pass unaffected (untouched by this PR)
- [ ] Watch total workflow wall-clock time: one compile instead of up to 7 redundant ones

### DIFF
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -146,68 +146,124 @@ jobs:
         run: cargo clippy --workspace --all-targets --all-features -- -D warnings
 
   # ---------------------------------------------------------------------------
+  # Build every test binary exactly once, archive it, share across all
+  # downstream test jobs.
+  # ---------------------------------------------------------------------------
+  # Two archives, not one: `cargo nextest archive --workspace` unifies
+  # features the same way for every crate in a given invocation, so as long
+  # as every consumer of an archive selects tests via `-E 'package(...)'`
+  # (not `-p`, which nextest rejects alongside --archive-file anyway) against
+  # THIS SAME invocation's output, there's no resolver-scope mismatch and
+  # nothing needs to recompile downstream (verified locally: archiving a
+  # subset, then `cargo nextest run --archive-file ... -E 'package(x)'` for
+  # an even narrower selection produces zero "Compiling" output).
+  #
+  # temps-providers/temps-backup's docker-tests feature can't be baked into
+  # one shared archive unconditionally, though: GitHub's ubuntu-latest
+  # runners have a live Docker daemon by default, so if unit-tests' runner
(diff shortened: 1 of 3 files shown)


dviejokfs:
@bherila CI won't take much time after this PR, merging it

PR #255, opened 2026-07-07
chore(license): adopt FSL-1.1-ALv2 — protect the hosted cloud, keep self-hosting free

dviejokfs:
## Proposal: adopt the Functional Source License (FSL-1.1-ALv2)

> **This is a proposal PR, opened for discussion — not merging until we've talked it through.**

### TL;DR — this only restricts cloud resellers. Nothing changes for anyone self-hosting Temps.

This PR relicenses the Temps platform from MIT OR Apache-2.0 to the [Functional Source License 1.1 with Apache 2.0 future grant](https://fsl.software) — the license Sentry created and uses.

**What stays exactly the same (the point of Temps):**
- Self-hosting is free, forever — personal projects, your company, your clients, at any scale, commercially.
- Full source access: read, modify, redistribute.
- Every release **automatically converts to Apache 2.0 two years after it ships**, irrevocably. Nothing is permanently proprietary.
- All previously released versions keep their original MIT OR Apache-2.0 terms.

**The single thing that changes:**
- You can no longer take Temps and sell it *as a hosted service or competing product* (i.e. launch your own "Temps Cloud"). Managed hosting is how Temps development gets funded, and this closes the door on that one free-riding scenario — and nothing else.

**What stays permissive (anything that touches *your* code):**

| Directory | License | Why |
|---|---|---|
| `sdks/` (node, python) | MIT | Embedded in user applications |
| `packages/` (@temps-sdk/api) | Apache-2.0 | Generated client, embedded in user apps |
| `apps/temps-cli/` | MIT | Runs on user machines/CI |
| `crates/temps-captcha-wasm/` | MIT OR Apache-2.0 (unchanged) | Delivered to end-user browsers |

### Changes in this PR
- `LICENSE` → FSL-1.1-ALv2 (canonical text from fsl.software)
- Per-directory permissive `LICENSE` files for `sdks/`, `packages/`, `apps/temps-cli/`
- `Cargo.toml` workspace license → `FSL-1.1-ALv2` (SPDX identifier)
- `NOTICE`, `README` (badge, tagline, comparison table, License section), `CONTRIBUTING.md` updated
- Adds a lightweight [DCO](https://developercertificate.org/) sign-off requirement for future contributions (`git commit -s`)
- Fixes the stale `LICENSE-APACHE` reference in CONTRIBUTING.md

### Precedent
Sentry, Codecov, GitBook, and Convex use FSL; n8n and Outline use similar fair-code licenses. The 2-year automatic Apache conversion is the key difference from BUSL-style licenses — it's a hard commitment that the code ends up genuinely open source.

---

@bherila — you're the largest outside contributor, so I'd really value your thoughts before this goes anywhere. To be explicit about intent: this is **only** about reserving the managed-cloud offering (Temps Cloud) as the project's funding model. Self-hosters — including commercial use — are completely unaffected, and everything converts to Apache 2.0 on a 2-year clock. Your existing contributions stay under the MIT/Apache terms you made them under; this applies going forward only. Does this seem fair to you? Any concerns with the FSL specifically vs. alternatives (AGPL, BUSL), or with the DCO requirement?


bherila:
How does this compare with the competition?

dviejokfs:
> How does this compare with the competition?

In what areas? The only thing that changes is that you can self host it but you cannot offer temps as a service. The rest can do it but most of the people I know self host it on their own.

bherila:
Coolify is Apache for example

https://github.com/coollabsio/coolify/blob/v4.x/LICENSE

dviejokfs:
Most people use it to self host their apps and databases

Will something about the above change if the license went from Apache to FSL or even APGL?

bherila:
I like Apache/MIT because it's very simple to think about. For my use case, this is a nice widget for my personal website where I host a bunch of "for fun" products like free HTML games and food/cooking recipes. My actual day job is at a big company and it's not web development.  So, the license change doesn't affect me because I am not in the web hosting business, and I don't intend to be. I _like_ (emotionally) the fact that I'm "free" from companies like Vercel that could change the terms at any time.

From a business standpoint if I may suggest, I feel the cloud value prop is _probably_ more about your own control panel / infra support (For example, Hetzner partnership where folks don't need to manage the servers) vs. the temps panel itself (where server upkeep is still required). 

On the SMB side, there may be an opportunity to partner with additional cloud providers (besides Hetzner) such as AWS Lightsail, to offer a prefab temps VM image on their service. i.e. AWS Lightsail users can get one-click deployment of Amazon Linux + Temps Control Panel. They previously partnered with Cpanel and Plesk to offer this btw, although that seems to be mostly deprecated since Plesk discontinued the free 3-domain version. 

On the enterprise side, I saw you're offering services like a SOC2 package, but I feel like that's probably unnecessary because the SOC2 package could probably be vibe coded for less than the cost of the Enterprise plan. The part that can't be vibe coded is the services. I do think the services add-on is an interesting take, but I also feel like your customer/audience may be mistargeted. For example: Many startups who need SOC2 are likely _already_ going to use a vendor like Vanta, so maybe you'd want to work with those SOC2 vendors to integrate vs. try to sell SOC2 directly.

Hope this $.02 is helpful!

bherila:
I will also say one more thing about SOC2, which is that often folks do not like control panels when they need SOC2, cyber insurance, etc. because it's a large surface area to audit and maintain. Less infra, closer to bare metal, is often preferred there.

PR #258, opened 2026-07-08
feat(proxy): hot-path metrics with node dashboard and default alerts

dviejokfs:
## What

Proxy observability end-to-end: the Pingora proxy now measures its own traffic, the console charts it, and the alert evaluator watches it — with zero API changes (reuses the existing `GET /nodes/{id}/metrics` endpoint and generated SDK binding).

- **Hot path** (`temps-proxy/src/metrics.rs`): lock-free `ProxyMetrics` — status-class counters (1xx–5xx) + fixed duration-bucket histogram. Recording = 3 relaxed atomic adds; no locks, no allocation, no I/O; bounded cardinality by construction (no per-path/host/project dimensions).
- **Sampler** (`temps-proxy/src/service/metrics_sampler.rs`): background task snapshots counters every `monitoring.scrape_interval_secs`, computes deltas (saturating, restart-safe), writes ~11 `SourceKind::Node` points per interval via one `write_batch`. Store selection mirrors the console (ClickHouse when configured+selected, else TimescaleDB) so writes land where reads look — works in single-process and split ADR-017 topologies.
- **Record sites**: `finalize_response` (proxied), `log_static_request` (all static responses, incl. assets excluded from logs), `fail_to_proxy` (failures).
- **Metrics**: `proxy.requests{,_1xx..~_5xx}` (interval deltas), `proxy.error_rate_percent`, `proxy.request_duration_avg/p50/p95/p99_ms` (bucket-interpolated).
- **Alerts**: migration `m20260708_000001` adds `node_id` as a third `monitoring_alert_rules` target (CHECK widened to exactly-one-of-three, partial + unique indexes). Evaluator groups node rules and fires `NodeMetricThreshold` alarms. Default proxy rules seeded idempotently at startup: error rate >20% (warn, 5m) / >50% (crit, 2m), p99 >5s (warn, 5m).
- **UI**: new `/proxy` page ("Observe" nav section, between Monitoring and Proxy Logs) charting requests by status class, error rate, latency percentiles, avg duration, with 1h/6h/24h/7d range selector, skeleton loading, empty/503 states. Generated SDK only.

## Load / saturation behaviour (per the new scalability requirement)

- Expected load: 100k+ req/s on the hot path → ~300ns of relaxed atomic adds per request; memory is constant (16 `u64` counters).
- Storage cost is constant regardless of traffic: ~11 points per scrape interval.
- On store-write failure: warn + drop; counters keep accumulating so the next successful cycle writes one wider delta (no data loss, no backpressure onto the proxy).

## Tests

- `temps-proxy`: 11 new unit tests (classification, bucketing, delta/saturation, percentile interpolation, point mapping) — pass.
- `temps-monitoring`: 94 tests pass incl. new node alarm-type + proxy-seed tests.
- `cargo check --lib` clean workspace-wide; `bunx tsc --noEmit` clean in `web/`.

## Notes for review

- Node id `0` = synthetic control-plane node (mirrors `CONTROL_PLANE_NODE_ID`); no FK on `monitoring_alert_rules.node_id` for that reason.
- Metric names deliberately avoid `_total`/`_count` suffixes because the read path treats those as raw cumulative counters, while these are pre-computed deltas.
- Follow-ups (not in this PR): per-worker-node proxy metrics if proxies ever run off the control plane; UI management (create/edit) for node-scoped alert rules — seeded defaults are visible to the evaluator but the alert-rule UI still targets services/deployments.


dviejokfs:
Tested end-to-end on a local instance (fresh DB, ports 18080/18081): traffic through the Pingora proxy → sampler flush → `service_metrics` rows (`proxy.requests=76`, `proxy.requests_2xx`, p50/p95/p99) → `GET /api/nodes/0/metrics` returns the series → `/proxy` page renders.

Two findings from testing, both addressed in 1869f3ea:
- The `(node_id, metric_name)` unique index means **one rule per metric** — the third seed (critical error-rate on the same metric) was a silent ON CONFLICT no-op. Seeds reduced to one per metric + a test guarding against duplicate seed metrics.
- **Pre-existing issue (out of scope):** the same applies to the service seeds — e.g. `postgres_default_seeds()` has warning *and* critical rules on `pg.connections_active`, so the critical variants have never actually inserted (`uidx_monitoring_alert_rules_service_metric` is `(service_id, metric_name)`). Worth a follow-up: either widen the unique indexes to include severity, or dedupe those seed lists.

dviejokfs:
UI iteration (21da064f): summary stat cards (req/s, total requests, error rate, p95), sortable per-project traffic table via `/analytics/general-stats?include_project_breakdown=true`, full-width layout matching /monitoring, and solid theme-token chart tooltips.

**Follow-up found:** chart tooltips across the app that use `hsl(var(--popover))` (e.g. ServiceMonitoring.tsx) are broken — the Tailwind v4 theme tokens are `oklch(...)` values, so wrapping them in `hsl()` yields an invalid color and the tooltip renders transparent. Fixed on /proxy by using `var(--popover)` directly; the other pages need the same one-line fix.

dviejokfs:
## Efficiency review @ 10k req/s (f2f90186)

**Empirical**: `ab -n 20000 -c 100` against the full local instance (proxy + console fallback + log pipeline + metrics recording): **12,550 req/s sustained, 0 failed, p50 8ms / p99 15ms** on a dev laptop. Counter integrity verified under load: `proxy.requests = 20,000` and the destination split summed exactly (console 20,000 + project 0 + other 0).

**Per-component analysis**:
- **Record path** (`logging()` hook): 4 relaxed atomic adds + one `Instant::elapsed` + one ≤10-byte string compare, zero allocation, zero I/O. At 10k req/s that's 40k uncontended-class atomic ops/s — noise; headroom extends well past 100k req/s. Recording moved to Pingora's `logging()` hook (verified in pingora-proxy source that it fires exactly once on all three completion paths: early request_filter return, normal finish, fail_to_proxy) — this both fixed missed short-circuited responses (redirects, ACME, password walls) and guarantees the sum invariant with a single record site.
- **Sampler**: constant cost — 14 points per 30s interval regardless of request rate, one `write_batch`.
- **Read path**: the /proxy page issues ~15 small indexed series queries per 30s refresh against a table growing 14 rows/30s — negligible.
- **Alert evaluator**: +2 seeded rules → one extra batched `query_latest` per 30s cycle.

**Pre-existing pressure points at 10k req/s (not introduced by this PR, flagged for awareness)**:
1. `ProxyLogBatchWriter` builds a `CreateProxyLogRequest` per request (header map clone + `serde_json::to_value`) before the bounded channel — the dominant per-request cost at high rates. The channel is bounded with a counted-drop shed policy, so it degrades instead of stalling, but at sustained 10k req/s operators should run the **ClickHouse** log backend; row-based TimescaleDB inserts at 10k rows/s will shed.
2. Filtered (per-project) charts aggregate `proxy_logs` time buckets — at 10k req/s that table grows ~864M rows/day; retention/TTL settings matter far more than the chart queries themselves.

PR #276, opened 2026-07-10
fix(proxy): track real body bytes for request/response bandwidth

dviejokfs:
## Summary
- The Proxy dashboard's "Bandwidth" panel (`/proxy`, per-project filtered view) always showed a flat 0 B line despite real traffic.
- Root cause: `total_request_bytes`/`total_response_bytes` were derived solely from the `Content-Length` header (`log_request`/`fail_to_proxy`). Chunked/streamed responses carry no `Content-Length` — a case this same file already explicitly detects and preserves (`is_chunked_response` in `response_filter`) — so for any backend that streams, bytes were always logged as `NULL` and summed to 0 in ClickHouse.
- There was also a dead scaffold for this: `ProxyContext::upstream_body_bytes_received` was declared and zero-initialized in three places but never incremented or read anywhere.
- Fix: accumulate actual body bytes per-chunk in `request_body_filter` (new override) and `response_body_filter` (now a thin wrapper around the existing logic, renamed `response_body_filter_inner`, so counting applies uniformly across every exit path — passthrough, SSE/WebSocket, buffered Markdown conversion). `log_request`/`fail_to_proxy` now prefer these accumulated counters, falling back to `Content-Length` only when a body never reached the filter (e.g. HEAD responses).
- Per-request counters on `ProxyContext`, no locks/atomics/extra I/O — Pingora already hands the chunk to these hooks regardless, so this is a couple of integer additions against work already happening on the hot path.

## Test plan
- [x] `cargo check --lib -p temps-proxy`
- [x] `cargo test --lib -p temps-proxy` (399 passed, 4 ignored)
- [x] `cargo clippy -p temps-proxy --lib --all-targets -- -D warnings` (clean, run via resolved binary, not rtk)
- [x] `cargo fmt -p temps-proxy -- --check` (clean)
- [ ] Manual: confirm the Bandwidth panel on `/proxy` (project-filtered view) shows nonzero bytes against a streaming/chunked backend

### DIFF
--- a/crates/temps-proxy/src/proxy.rs
+++ b/crates/temps-proxy/src/proxy.rs
@@ -389,6 +389,11 @@ pub struct ProxyContext {
     /// unlike the `Content-Length` header, this is always populated even for
     /// chunked-encoded request bodies.
     pub client_body_bytes_received: usize,
+    /// Proxy log entry built in `log_request` (response-header time), held
+    /// here rather than sent immediately because `upstream_body_bytes_received`
+    /// isn't fully accumulated until the response body finishes streaming.
+    /// The `logging` hook patches in the final byte count and sends it.
+    pub pending_proxy_log: Option<CreateProxyLogRequest>,
     /// Whether the client requested a Markdown response via `Accept: text/markdown`
     pub wants_markdown: bool,
     /// Accumulated body bytes for HTML-to-Markdown conversion
@@ -1348,7 +1353,7 @@ impl LoadBalancer {
         &self,
         _session: &PingoraSession,
         upstream_response: &ResponseHeader,
-        ctx: &ProxyContext,
+        ctx: &mut ProxyContext,
     ) -> Result<()> {
         // Skip logging for internal temps API routes
         if ctx.path.starts_with(ROUTE_PREFIX_TEMPS) {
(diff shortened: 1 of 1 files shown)


dviejokfs:
Ran the `security-auditor` before merging. It caught a real correctness bug: `log_request` runs from `response_filter` (response *header* time), which is before `response_body_filter` streams any chunks — so `upstream_body_bytes_received` was always 0 when the log entry was built, making the response-side fix a no-op for exactly the chunked-response case this PR targets. (Request-side counting was unaffected and confirmed correct — the client body finishes streaming before the upstream replies.)

Pushed a follow-up commit: the log entry is now stashed on `ctx.pending_proxy_log` at header time and sent from the `logging` hook instead, which Pingora calls once per request after the response body has fully streamed (`finish()`). That hook patches `response_size_bytes` with the final accumulated byte count before enqueueing, falling back to the Content-Length value already captured if no body ever streamed.

Also flagged (non-blocking, LOW): for Markdown-converted responses (`Accept: text/markdown`), the byte count reflects bytes actually forwarded to the client (post-conversion), not raw upstream HTML bytes received — that's intentional given the panel measures egress bandwidth, and is now documented on the field.

Re-verified after the fix: `cargo check --lib`, `cargo test --lib` (399 passed), `cargo clippy -- -D warnings` (clean, resolved binary not rtk), `cargo fmt --check` (clean), plus the pre-commit hooks (workspace clippy, Conventional Commit) on both commits.

PR #278, opened 2026-07-11
perf(proxy): serve proxy-log stats from a 1-minute continuous aggregate

dviejokfs:
## Problem

The proxy dashboard's **Traffic by project** panel and the project-filtered charts hang on high-traffic installs (observed 20s+ on an instance logging ~1.7M proxy-log rows/hour for a single project). Both are backed by request-time aggregation over raw `proxy_logs`:

- `GET /proxy-logs/stats/projects-health` — `COUNT(*)` / `SUM(status>=500)` / `AVG(response_time_ms)` per project over the window
- `GET /proxy-logs/stats/time-buckets` — `time_bucket_gapfill` over the same raw rows

The `(project_id, timestamp DESC)` index (m20260528) only narrows *which* rows to read — every matching row still needs a heap fetch for `response_time_ms` / `status_code`, so the cost is O(requests in window) no matter what.

## Fix

New continuous aggregate `proxy_logs_stats_1m`: 1-minute buckets keyed `(project_id, environment_id, is_bot)` storing counts and sums (never averages, so buckets roll up losslessly to any coarser interval; `avg = sum/count` reproduces raw `AVG` exactly incl. NULL handling).

- **Freshness**: real-time aggregation (`materialized_only = false`) + every-minute refresh policy (end offset 1 min) — queries union materialized buckets with at most ~1–2 min of raw tail.
- **Routing**: `get_projects_health_summary` uses the aggregate whenever it exists; `get_time_bucket_stats` uses it only when every set filter is one of its grouping columns (`project_id`/`environment_id`/`is_bot`/`has_project`) and the interval is a whole-minute multiple — all other filter shapes (method, status class, deployment, …) keep the raw path unchanged. ClickHouse storage path untouched.
- **Backfill**: the existing post-migration hook (`run_post_migration_backfill`, same as `events_hourly`) refreshes historical data outside the migration transaction; until it completes, real-time aggregation still answers correctly at raw-scan cost.
- **Retention**: 30 days on the aggregate, matching `proxy_logs` (m20260225); dashboard max window is 7d.

## Load / memory bound (per hot-path policy)

Reads become O(minutes × active series) instead of O(requests): a 1h window over 12 projects is ≤ ~60 × series rows regardless of traffic. Write amplification is constant: ~1 aggregate row per active `(project, environment, is_bot)` series per minute, refreshed incrementally (invalidation-driven). At saturation the refresh job lags; real-time aggregation then covers the gap with a bounded raw scan.

## Drive-by fix

The cross-path test caught a live bug: raw-path `total_request_bytes`/`total_response_bytes` are `SUM(bigint)` → Postgres `numeric`, which `try_get::<i64>` silently failed to parse (`unwrap_or(0)`), so the filtered view's **Bandwidth chart always reported 0 bytes**. Fixed with explicit `::bigint` casts.

## Testing

- `test_stats_served_from_cagg_match_raw_path` — full pipeline against real TimescaleDB (raw inserts → aggregate via real-time aggregation → health + time-bucket queries), cross-checking aggregate-path results against the raw path, incl. `is_bot` filtering, error-rate status, and the unknown-project case
- Unit tests for the filter-eligibility and interval-alignment routing helpers
- `cargo test --lib -p temps-proxy` (403 passed), `-p temps-database` (25 passed); clippy `--all-targets -D warnings` clean on touched crates
- The 4 failing tests in `migration_tests.rs` fail identically on `main` (pre-existing, environmental)

### DIFF
--- a/crates/temps-database/src/connection.rs
+++ b/crates/temps-database/src/connection.rs
@@ -558,16 +558,57 @@ pub async fn get_pending_migration_names(db: &DbConnection) -> ServiceResult<Vec
 ///
 /// Run this on a long-lived runtime (e.g. via `tokio::spawn`) so it never blocks
 /// startup; it is decoupled from `establish_connection` for exactly that reason.
+/// One continuous-aggregate refresh window: optional start (`None` = NULL,
+/// i.e. unbounded) and end, both as SQL expressions.
+type RefreshWindow = (Option<String>, String);
+
 pub async fn run_post_migration_backfill(db: &DatabaseConnection) -> ServiceResult<()> {
-    // (view name, refresh window end). Each window end mirrors the
-    // aggregate's own refresh-policy end_offset so the backfill never
-    // overlaps the region the policy is responsible for.
-    let aggregates: [(&str, &str); 2] = [
-        ("events_hourly", "NOW() - INTERVAL '1 hour'"),
-        ("proxy_logs_stats_1m", "NOW() - INTERVAL '1 minute'"),
+    // Refresh windows per aggregate, executed in order. `None` as a bound
+    // means unbounded (NULL). Window ends mirror each aggregate's
+    // refresh-policy end_offset so the backfill never overlaps the region
+    // the policy is responsible for.
+    //
+    // `proxy_logs_stats_1m` is chunked into 1-day windows, NEWEST FIRST,
+    // instead of one unbounded call, for two reasons on large installs
+    // (100M+ retained rows):
+    //   * Each `CALL refresh_continuous_aggregate()` is one transaction; a
+    //     single 30-day refresh over that volume is a multi-minute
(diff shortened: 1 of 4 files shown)


dviejokfs:
Follow-up commit for large-install deploys (b4606728): the startup backfill now refreshes in **1-day windows, newest first**, instead of one unbounded `refresh_continuous_aggregate` call.

Why: with ~100M retained rows the single call was one multi-minute transaction, and once the every-minute refresh policy advanced the watermark, real-time aggregation stopped covering older un-materialized regions — 24h/7d dashboard windows would transiently under-report until the backfill reached them. Newest-first chunking makes 1h/6h/24h correct after the first chunk (seconds to ~a minute), then works backwards; each chunk is a small independent transaction, and per-window failures are logged and skipped.

Migration apply itself stays instant (all metadata: `WITH NO DATA` view, empty index, policy registrations).

PR #279, opened 2026-07-11
feat(otel,proxy): parameterize ClickHouse retention via per-row retention_days

dviejokfs:
## Summary
- Adds a `RetentionResolver` extension point (`temps-core`) so a plugin can override the fixed ClickHouse TTL for `spans`/`proxy_logs` on a per-project basis. OSS registers `FixedRetentionResolver` by default, which is behaviorally a no-op: it always returns the current hardcoded constants (90 days for spans, 30 for proxy_logs), so plugin-free instances are unaffected.
- Adds additive ClickHouse migrations (`retention_days UInt16` column + `MODIFY TTL` to reference it instead of a fixed `INTERVAL`) for both tables. Existing rows get a `DEFAULT` equal to the prior fixed interval, so no existing data is re-dated.
- `ChSpanRow`/`ChProxyLogRow` gain a trailing `retention_days` field (column order is load-bearing for positional binary serialization — appended last to match `ADD COLUMN` ordering).
- Ingest paths resolve the effective value per project at write time via the injected resolver; unrouted proxy requests (no `project_id`) use the fixed table default directly rather than a fabricated project ID.
- Updated the `ch_backfill_domains` CLI backfill tool's struct literals to populate the table default for historical rows.

## Test plan
- [x] `cargo check --lib` clean across `temps-core`, `temps-otel`, `temps-proxy`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean (workspace-wide, matches pre-commit hook)
- [x] `cargo test --lib -p temps-core -p temps-otel -p temps-proxy` — 399+ passing, 0 failed
- [x] Pre-commit hooks (fmt, clippy, conventional commit) passed on the actual commit

### DIFF
--- a/crates/temps-cli/src/commands/ch_backfill_domains.rs
+++ b/crates/temps-cli/src/commands/ch_backfill_domains.rs
@@ -550,6 +550,7 @@ mod proxy_logs {
             error_group_id: r.error_group_id,
             // Source timestamp ms as the dedup version — stable across re-runs.
             _version: ms as u64,
+            retention_days: temps_core::RetentionTable::ProxyLogs.default_days(),
         }
     }
 
@@ -671,6 +672,7 @@ mod traces {
             events: r.events.clone().unwrap_or_else(|| "[]".into()),
             // Source start_time ms as the dedup version — stable across re-runs.
             _version: r.start_time.timestamp_millis() as u64,
+            retention_days: temps_core::RetentionTable::Spans.default_days(),
         }
     }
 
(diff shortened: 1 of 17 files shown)


dviejokfs:
Follow-up commit (f85e3ea7): the first commit's `context.get_service::<dyn RetentionResolver>()` call inside `register_services` would never actually find a plugin-registered resolver, since `OtelPlugin`/`ProxyPlugin` register (and construct their ClickHouse storage) before any plugin added via `extra_plugins` gets a chance to register one — same phase, earlier position. Fixed by introducing `RetentionResolverSlot` (ArcSwap-backed, mirrors `DeploymentGate`'s `deployment_gate_slot` two-phase handoff) so the actual lookup happens in `initialize_plugin_services`, which runs only after every plugin has finished registering.

PR #284, opened 2026-07-12
feat(web): world map of web vitals by country on the speed page

dviejokfs:
Follow-up to #282, which added the geo dimensions, `country_code`, and segment filters this builds on.

## What

A choropleth world map on the Speed page (between the trend chart and the breakdown):

- Countries filled by the same **good / needs-work / poor** thresholds as the p75 tiles — e.g. US red, Spain green
- Metric selector: TTFB / LCP / FCP / INP / CLS
- Hover tooltip with the country's average value + sample count; legend; "N samples without location" note
- **Clicking a country adds the same `filter_country` chip** as the breakdown, rescoring the score/tiles/trend/breakdown/map together
- Respects the environment, time-range, Desktop/Mobile, and active segment filters

## How

- `react-simple-maps@3` (was already a dependency, previously unused) + `world-atlas` countries-110m TopoJSON **bundled into the repo** (~105KB) so self-hosted installs make no runtime CDN calls
- Country rows are matched to geometries via the ISO alpha-2 `country_code` from `/performance/page-metrics` (added in #282), converted to the topojson's numeric ids with the new `iso-3166` dependency — no fragile name matching
- Verified react-simple-maps works under React 19 (its peer range says ≤18)

## Verification

Live against seeded US/ES/DE data on a local stack: 176 countries render; exactly one red (US, 2.55s avg TTFB), one amber (Germany, 1.0s), one emerald (Spain, 264ms); hover tooltip correct; clicking the US chips the filter and the map itself rescopes to the US cohort. `tsc --noEmit` and `rsbuild build` clean.

screenshot 

## Note

With real temps-landing data the US will show red partly due to the datacenter-bot cohort (~28% of samples averaging ~10s TTFB) — ingest-side bot filtering is the next planned step so the map colors reflect real users.


dviejokfs:
Screenshot from the live verification (seeded US/ES/DE data, dark theme):

PR #294, opened 2026-07-12
fix(migrations): skip empty visitor deduplication rewrites

bherila:
## Summary

- check whether `visitor` actually contains duplicate `(visitor_id, project_id)` pairs before running the deduplication rewrites
- skip all eight foreign-key `UPDATE` statements when there is nothing to rewrite
- reuse the unique-index SQL in both the fast path and deduplication path

## Root cause

TimescaleDB can eagerly decompress hypertable chunks for an `UPDATE` even when its join source produces no rows. On an installation with no duplicate visitors, the migration's logically no-op rewrites still exceeded `timescaledb.max_tuples_decompressed_per_dml_transaction` and blocked the upgrade.

## Impact

Upgrades with already-unique visitor data now create the unique index directly without touching compressed analytics tables. Installations that do contain duplicates retain the existing deduplication and foreign-key repointing behavior.

## Validation

- `cargo check -p temps-migrations --lib`
- `git diff --check`
- applied successfully against the affected TimescaleDB installation; all seven pending migrations completed and the upgraded server started healthy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


dviejokfs:
@bherila is it ready to be merged? Since it's in draft

LGTM

PR #295, opened 2026-07-13
fix(email): harden delivery pipeline end-to-end

clau1902:
## Summary

Hardens the email-sending subsystem across the Rust backend, CLI, MCP tools, and web UI:

- **Suppression list**: refuses to send to previously hard-bounced/complained addresses, checked across `to`/`cc`/`bcc`, dropping only the suppressed addresses rather than blocking the whole send.
- **Multi-provider failover**: ordered fallback chain per domain, with a per-provider circuit breaker, rate limiter, and retryable-error classification (HTTP status for Scaleway, SDK error kind for SES, SMTP reply code for SMTP) driving bounded retries before moving to the next provider.
- **SNS signature verification**: real RSA signature verification for SES bounce/complaint webhooks (previously unverified), plus an orphaned-event fix and unified event-type naming.
- **DMARC record generation**: shown as a recommendation, deliberately excluded from auto-DNS-creation and from domain verification status, since publishing a DMARC policy isn't purely additive like SPF/DKIM/MX.
- **DNS verification precision fix**: removed a bidirectional substring check that could false-positive "Verified" on an unrelated short TXT record.
- **SSRF fixes**: click-redirect and SMTP-prober hardening, including a DNS-rebinding TOCTOU fix (resolve once, validate the resolved address, connect to that pinned address instead of letting a second independent lookup happen at connect time).
- **GDPR retention job**: batched deletion of tracking-event rows (which carry IP/user-agent) past a 90-day window, on a 24h scheduler.
- **CLI/MCP/web**: new/fixed email tooling; MCP `get_email` truncates and labels email bodies as untrusted content before it flows into an LLM's context; CLI strips control/ANSI escape sequences before printing email bodies to the terminal; web UI gets search, pagination, dedup, and per-domain deliverability stats.

## Test plan

- [x] `cargo check --workspace --lib` — full workspace builds clean
- [x] `cargo test -p temps-email -p temps-email-tracking` — 183 + 46 tests pass (includes new tests for suppression filtering, circuit breaker/rate limiter, retention purge, SSRF rejection)
- [x] `bun run typecheck`/`type-check` clean for CLI, MCP, and web (no new errors; pre-existing unrelated errors in CLI left untouched)
- [x] Secret scan across the full diff — clean
- [x] Two independent adversarial safety reviews (Rust backend, TS/CLI/MCP/web) — findings addressed: DNS-rebinding SSRF bypass, suppression list not checking cc/bcc and over-blocking on partial match, unbounded retention DELETE, N+1 query in failover chain lookup, prompt-injection risk in MCP tool output, terminal escape-sequence risk in CLI output
- [ ] Manual smoke test of the full send → bounce → suppress → failover flow against a real provider (not done in this session — no live provider credentials available)


clau1902:
Closing to split into smaller, focused PRs — one per issue — for easier review. Follow-up PRs incoming.

PR #296, opened 2026-07-13
feat(email): add bounce/complaint suppression list

clau1902:
## Summary

Refuses to send to addresses that previously hard-bounced or filed a spam complaint — repeatedly emailing one is exactly what gets a sending domain's reputation downgraded by receiving mail providers.

- New `suppressed_recipients` table + `SuppressionService` (suppress/unsuppress/`suppressed_among`), normalized (trim+lowercase) on both write and lookup.
- Checked in `EmailService::send` after the email row is inserted (still visible for debugging) but before any domain/provider work.
- Checks `to`/`cc`/`bcc` together, not just `to` — a suppressed address left in `cc`/`bcc` would otherwise still receive mail.
- Drops only the suppressed addresses rather than capturing the whole send — a suppressed address mixed into `to` alongside legitimate recipients must not deny delivery to everyone on the email.
- If every `to` address ends up suppressed (or `to` was already empty), the email is captured (not sent) with a clear `error_message`.

## Test plan

- [x] `cargo check --lib -p temps-email -p temps-migrations -p temps-entities` — clean
- [x] `cargo test -p temps-email --lib` — 172 tests pass, including new unit tests for the filter logic (drops only the suppressed cc address, keeps other `to` recipients, case/whitespace-insensitive matching, empties `to` when all suppressed) and suppression-service integration tests


dviejokfs:
Closing — this feature already shipped to `main` via #297. #296's branch (`fix/email-suppression-list`) was merged into #297's branch (`fix/email-sns-verification`) during its development, and came along when #297 was squash-merged today (commit 2210cf3f). `main` now has a more complete version than this branch: the SNS webhook actually calls `SuppressionService::suppress()` on bounce/complaint (this branch never wired that in), `suppress()` uses an atomic `ON CONFLICT` upsert instead of a racy find-then-insert, and suppression is scoped per-domain instead of global. Merging this branch as-is would regress `main` back to the older, unwired, race-prone version.

One thing this PR's review did surface that's still live on `main` and not yet fixed: suppressed addresses are logged at `info!` and written verbatim into `emails.error_message`, which is a minor enumeration/PII exposure risk. Following up on that separately.

PR #306, opened 2026-07-13
fix(cli): type email command output and sanitize rendered bodies

clau1902:
## Summary

Independent of the other email-hardening PRs — touches only `apps/temps-cli` and `mcp/`.

- Fix CLI email commands (send/list/get/stats); strips control/ANSI escape sequences from email bodies before printing to the terminal — bodies are attacker-controlled (anything sent through the platform's send API), and a crafted body could otherwise manipulate the cursor, hide/spoof output, or set the terminal window title.
- New MCP email tools (`list_emails`, `get_email`, `get_email_stats`). `get_email` truncates and clearly labels `html_body`/`text_body` as untrusted content before returning them: this tool's output typically flows straight into an LLM's context, so an admin's AI assistant calling it on an attacker-composed email could otherwise ingest prompt-injection text disguised as an email body.

## Test plan

- [x] `bun run typecheck` (CLI) / `bun run type-check` (MCP) — clean, no new errors

### DIFF
--- a/apps/temps-cli/src/commands/emails/index.ts
+++ b/apps/temps-cli/src/commands/emails/index.ts
@@ -8,10 +8,11 @@ import {
   getEmailStats,
   validateEmail,
 } from '../../api/sdk.gen.js'
+import type { EmailResponse } from '../../api/types.gen.js'
 import { withSpinner } from '../../ui/spinner.js'
 import { printTable, statusBadge, type TableColumn } from '../../ui/table.js'
 import { promptText, promptConfirm } from '../../ui/prompts.js'
-import { newline, header, icons, json, colors, success, info, warning, keyValue } from '../../ui/output.js'
+import { newline, header, icons, json, colors, success, info, warning, keyValue, formatDate } from '../../ui/output.js'
 
 interface ListOptions {
   json?: boolean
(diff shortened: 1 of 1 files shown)


dviejokfs:
We've decided to sunset the `@temps-sdk/mcp` package (context: skills + `bunx @temps-sdk/cli` cover our own dev workflows, and we don't want to carry a second, separately-versioned tool surface for end users long-term — see follow-up PR for the deprecation).

Given that, could you drop the `mcp/` changes from this PR (`mcp/src/tools/emails.ts` and the `mcp/src/tools/index.ts` registration) and keep just the `apps/temps-cli` email fix (typed `EmailResponse`/`EmailStatsResponse`, ANSI-stripping)? That part looks good on its own and I'd rather land it independently of the MCP surface we're about to remove.

dviejokfs:
@clau1902 I removed mcp completely from the project please remove it too from this PR

clau1902:
Done — dropped the `mcp/` changes entirely (both `mcp/src/tools/emails.ts` and the `mcp/src/tools/index.ts` registration). Rebased onto current `main`, so the PR is now just the `apps/temps-cli` email change:

- typed `emails list/show/stats` output with the generated `EmailResponse`/`EmailStatsResponse` (real fields: to/from/cc/bcc, `sent_at`, tags, open/click tracking, and the `total/sent/queued/failed/captured` stats), replacing the `Record<string, unknown>` casts
- ANSI/control-escape stripping of the email body before printing (bodies are attacker-controlled via the send API)

Updated the title accordingly. `emails/index.ts` typechecks clean.

PR #309, opened 2026-07-13
fix(email): minimize aged tracking connection metadata

clau1902:
## Summary

Email events can retain IP addresses and user-agent strings indefinitely. This change minimizes that connection metadata after 90 days while preserving event facts and their email association for analytics and deliverability. It does not claim to anonymize events.

- Redacts only `ip_address` and `user_agent` in ordered 5,000-row batches.
- Uses `FOR UPDATE SKIP LOCKED` so multiple nodes cooperate safely.
- Builds a matching partial `(created_at, id)` index concurrently, repairs invalid interrupted builds, and gates redaction until the index is valid and ready.
- Serializes index maintenance per PostgreSQL schema on one pinned session, including max-connections=1 migration setups.
- Runs immediately, then every 24 hours; failures retry after 5 minutes.
- Manages the scheduler lifecycle without an Arc cycle.
- Rejects zero and unrepresentable retention periods with typed errors.

## Test plan

- [x] `cargo test --lib -p temps-email` — 166 passed with isolated PostgreSQL
- [x] Exact cutoff and recent/future preservation
- [x] 5,001-row multi-batch redaction
- [x] Zero and `u32::MAX` validation
- [x] Typed database failure and index-not-ready behavior
- [x] Valid/ready partial index and invalid-index repair
- [x] Max-connections=1 migration pool regression
- [x] Scheduler shutdown lifecycle
- [x] `cargo check --lib -p temps-email -p temps-database -p temps-cli`
- [x] Strict Clippy for the changed crates

Security, Rust, and test reviews approve the final design.


dviejokfs:
Closing this because the proposed solution adds substantial operational complexity for a retention problem we have not established in practice. The scheduler lifecycle, concurrent index creation and repair, advisory locking, pinned database sessions, batching, and retry behavior expand the database and startup surface considerably, while enforcing a fixed global 90-day policy that users did not request or control.

If tracking-metadata retention becomes a demonstrated requirement, we should revisit it with a smaller, user-controlled design: retention disabled by default, configuration stored on the relevant database record, and reuse of an existing maintenance/job mechanism where possible. We should only add specialized index-management machinery once production evidence shows it is necessary.

PR #331, opened 2026-07-13
feat(cli): add first-class reverse-proxy routes

bherila:
## What

Adds a first-class `temps route` command group for mapping an external hostname to an unmanaged upstream `host:port`:

- `temps route add`
- `temps route list` / `ls`
- `temps route show`
- `temps route rm`

## Why

Temps already supports these entries through `custom_routes` and `LbService`, but operators currently need hand-written SQL. This exposes the existing data-plane capability as a validated, scriptable CLI surface.

## Design

- Uses `LbService` and the same direct-database pattern as `temps domain import`.
- Keeps this as top-level `temps route`; `temps proxy` remains the standalone proxy process.
- Supports HTTP TLS termination and TLS SNI passthrough.
- Makes no data-plane changes.
- Documents rationale and follow-ups in `docs/adr/020-reverse-proxy-routes.md`.

## Validation

The branch was first validated against the fork: formatting, Cargo Check, Clippy, TypeScript, unit tests, integration tests, MariaDB PITR, build-test-binaries, and E2E deployment tests all passed. The clean upstream-based branch will run the upstream matrix again.


dviejokfs:
The CLI is in typescript in apps/temps-sdk

The separation is made to have separated release processes since the client and the server may have different release schedules.

I'd say these commands are already added to the temps sdk cli, but probably worth checking it again.

dviejokfs:
btw dependabot broke main, because of the automatic merge

dviejokfs:
As I said, the routes are in bunx @temps-sdk/cli , there's no need to add them to the rust cli, rust cli must be for the platform, backups and restore.

PR #351, opened 2026-07-14
fix(auth): restrict audit:read to administration roles

clau1902:
## What

Restricts the `audit:read` permission to the administration roles. The audit log is a platform-wide view (it records every user's and admin's actions, IP addresses, and emails), and its list endpoint accepts an arbitrary `user_id` filter with no per-caller scoping. `AuditRead` was granted to the low-privilege `User` and `Reader` roles in addition to `Admin`/`PlatformAdmin`, so far more principals could read the full trail than the data warrants.

## Change

- `permissions.rs`: remove `AuditRead` from the `User` and `Reader` role arms. It stays with `Admin` and `PlatformAdmin`, and — being permission-based — can still be granted to a specific API key when an operator wants to delegate audit access. The handler already gates on the permission, so no handler change is needed.
- Add a regression test asserting only the administration roles hold `audit:read`.

**Behavior change:** `User` and `Reader` roles no longer read the audit log.

## Tests

`cargo test --lib -p temps-auth` passes, including the new matrix test.

### DIFF
--- a/crates/temps-audit/Cargo.toml
+++ b/crates/temps-audit/Cargo.toml
@@ -25,4 +25,5 @@ axum = { workspace = true }
 serde_json = { workspace = true }
 
 [dev-dependencies]
+# `mock` powers MockDatabase for the handler authorization tests.
 sea-orm = { workspace = true, features = ["mock"] }
--- a/crates/temps-audit/src/handlers/handlers.rs
+++ b/crates/temps-audit/src/handlers/handlers.rs
@@ -42,6 +42,7 @@ pub fn configure_routes() -> Router<Arc<AppState>> {
     responses(
         (status = 200, description = "List of audit logs", body = Vec<AuditLogResponse>),
         (status = 401, description = "Unauthorized"),
+        (status = 403, description = "Insufficient permissions"),
         (status = 500, description = "Internal server error")
     ),
     security(("api_key" = []))
@@ -93,6 +94,7 @@ async fn list_audit_logs(
         (status = 200, description = "Audit log details", body = AuditLogResponse),
         (status = 404, description = "Audit log not found"),
         (status = 401, description = "Unauthorized"),
+        (status = 403, description = "Insufficient permissions"),
         (status = 500, description = "Internal server error")
     ),
     security(("api_key" = []))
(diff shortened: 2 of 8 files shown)


dviejokfs:
@clau1902 the UI must be updated to not show audit logs if the user doesn't have permissions to access it

dviejokfs:
The backend permission change looks correct, and the existing role-matrix unit test plus green CI are a solid base. Before merge, I think this needs a small consistency pass:

- Add handler/router authorization tests for both audit endpoints (`list_audit_logs` and `get_audit_log`): `User` and `Reader` should receive 403; `Admin` and `PlatformAdmin` should pass the guard; and a custom API key carrying `audit:read` should pass.
- Add the missing 403 response to both OpenAPI endpoint declarations.
- Gate all audit UI entry points to `admin`/`platform_admin`: the sidebar, command palette, direct `/audit-logs` route, and the audit queries/stats/section on `UserDetail`. Unauthorized roles should not issue those queries and then surface 403 errors.

A real-database integration test is not needed here because the PR only changes the static role-to-permission mapping; handler-level authorization tests are the appropriate boundary coverage.

Also, #350 and #351 currently conflict in `crates/temps-auth/src/permissions.rs`, so whichever lands second will need a rebase that retains both permission-matrix tests.

clau1902:
Thanks for the review — addressed the consistency pass:

- **Handler authorization tests** (`temps-audit`): `list_audit_logs`/`get_audit_log` now have tests asserting `User` and `Reader` get 403, while `Admin`, `PlatformAdmin`, and a custom API key carrying `audit:read` pass the guard. Built with `MockDatabase` (added the sea-orm `mock` dev-dependency), no real DB needed.
- **OpenAPI**: added the `403` response to both endpoint declarations.
- **UI gating**: new `useCanViewAuditLogs` hook (`admin`/`platform_admin`, mirroring the matrix) applied to all four entry points — sidebar item, command-palette entry, the `/audit-logs` route (redirect + disabled queries), and the audit stat cards + per-user audit section on `UserDetail` (queries disabled so unauthorized roles never issue a 403 request).

On the #350/#351 conflict in `permissions.rs`: leaving it as-is per the plan — it's the two permission-matrix tests landing at the same spot, so whoever merges second keeps both test functions.

`cargo test --lib -p temps-auth -p temps-audit` is green; the edited web files pass `tsc`/`eslint` cleanly.

clau1902:
Also went ahead and removed the #350/#351 conflict you noted: moved this PR's `audit:read` matrix test next to the deployment-token tests (away from the `test_role_has_permission_method` anchor that #350 also uses). Verified with `git merge-tree` that #350/#351/#352 now all merge cleanly in any order, so no manual conflict resolution is needed for whoever lands second.

PR #354, opened 2026-07-15
fix(auth): require MFA on magic-link login

clau1902:
## What

Password login already withholds a real session when the user has MFA enabled — it issues a short-lived `mfa_session` challenge and returns `mfa_required: true` until the TOTP code is verified. The magic-link verify handler skipped that branch and called `create_session` unconditionally, so a user with MFA enabled who followed a magic link was fully logged in **without the second factor**.

## Change

- `handlers.rs` (`verify_magic_link`): when the user has MFA enabled, start the same first-factor challenge as `login` (short-lived `mfa_session` cookie, `mfa_required: true`) instead of creating a session. Non-MFA users are unaffected.
- Adds a Docker-gated integration regression test asserting a magic-link verify for an MFA-enabled user sets an `mfa_session` challenge cookie and does **not** mint an authenticating session cookie (skips gracefully without Docker).

This complements the existing `mfa_pending` session discriminator already on `main`: magic-link now routes MFA users through `create_mfa_session` (which sets `mfa_pending = true`), so the resulting token cannot authenticate a real request.

## Tests

`cargo test --lib -p temps-auth` passes (275 tests); the new integration test ran end-to-end against a test database.

### DIFF
--- a/crates/temps-auth/src/handlers.rs
+++ b/crates/temps-auth/src/handlers.rs
@@ -2117,7 +2117,8 @@ async fn disable_mfa(
 mod tests {
     use super::{
         assign_role, authorize_admin_target, authorize_role_assignment, delete_user, remove_role,
-        verify_magic_link, AdminTargetDenied, AssignRoleRequest, RoleChangeDenied, VerifyTokenQuery,
+        verify_magic_link, AdminTargetDenied, AssignRoleRequest, RoleChangeDenied,
+        VerifyTokenQuery,
     };
     use crate::auth_service::UserAuthError;
     use crate::context::AuthContext;
--- a/crates/temps-auth/src/handlers.rs
+++ b/crates/temps-auth/src/handlers.rs
@@ -874,7 +874,48 @@ pub async fn verify_magic_link(
 ) -> Result<impl IntoResponse, temps_core::problemdetails::Problem> {
     match state.auth_service.verify_magic_link(&query.token).await {
         Ok(user) => {
-            // Create session
+            // If MFA is enabled, do not issue a real session from the magic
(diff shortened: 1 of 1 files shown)


dviejokfs:
Looking at this closer: rather than patching the MFA bypass, **magic-link login should be removed entirely — nothing uses it.**

Where it stands in the codebase:

- **No first-party consumer.** `web/src/pages/Login.tsx` and `CliLogin.tsx` never reference it — the login screen has never offered magic link. The only frontend mention is descriptive copy in `AuthSettingsPage.tsx`. The `magic_link_available` flag in the auth-config response exists only in generated types; no component reads it.
- **No backend consumer.** No crate outside `temps-auth` calls `send_magic_link` / `verify_magic_link`. The CLI uses the device flow.
- **It's a live unauthenticated endpoint** (`POST /auth/magic-link/request`, `GET /auth/magic-link/verify`) with no product surface behind it — pure attack surface, as this very PR demonstrates.

Removal scope: `temps-auth` routes/service/email template, `magic_link_tokens` entity + drop migration, SDK/openapi regeneration, ADR-015 amendment (it currently names magic link as a permanent Community login method and the IdP-down fallback for SSO accounts — password reset needs to cover that recovery path), and the `AuthSettingsPage` copy.

If removal lands, this PR becomes moot and can be closed. If magic link ships in even one more release, this fix should merge first since the bypass is live on `main`.

clau1902:
Superseded by #375, which removes magic-link login entirely — there is no longer an endpoint to require MFA on, so this hardening is moot. Suggest closing in favour of #375.

dviejokfs:
superseeded by #375

PR #359, opened 2026-07-16
feat(analytics): add raw event entries drill-down with JSON props view

dviejokfs:
## Summary

The event detail page (`EventDetail`) only showed aggregates (total/visitors/sessions, referrer/country/browser breakdowns) and per-visitor rollups — there was no way to see the individual occurrences of an event type, or to inspect the custom data (`props` JSON) a specific occurrence carried.

This PR adds:

- **`GET /analytics/event-entries`** — paginated raw occurrences of a given `event_name` (time-bounded query over the `events` hypertable, joined with `visitor` and `ip_geolocations`), each entry including its `props` jsonb. Guarded by `AnalyticsRead` + project scope/access guards like the sibling `event-detail`/`event-visitors` endpoints. Pagination capped at 100 per page.
- **"Events" card in `EventDetail`** — lists each occurrence (time, visitor link, page, inline data preview); clicking a row expands a pretty-printed JSON view of the custom data with a copy button. Rows without custom data show a dash and don't expand. Skeleton loading, mobile-responsive columns.
- Generated API client entries for `getEventEntries` (hand-added in the generator's exact output style; the next `openapi-ts` regen against this branch should be a no-op).

**Deliberately excluded** from the AI read-tools allowlist in `console.rs`: this endpoint exposes developer custom-property values, which are held pending explicit review there (same as `get_property_breakdown`).

## Load & memory bound

Control-plane read path. Both queries are time-bounded on the hypertable, parameterized, and hard-capped at 100 rows per page — constant memory per request.

## Testing

- New integration test `test_get_event_entries_returns_props` against a real TimescaleDB container: props JSON round-trip, ordering (most recent first), pagination, environment filter. All 24 temps-analytics tests pass.
- `cargo check --lib` (workspace) and `cargo clippy --all-targets -D warnings` (crate) clean.
- Web: `tsc --noEmit`, eslint, prettier clean.
- Not yet exercised in a live browser.


dviejokfs:
Added two follow-ups: the custom-data JSON view now uses the shared `CodeBlock` component (syntax highlighting, copy/wrap/line-number controls), and the entries table gained a **Columns** picker that auto-discovers the props structure from the entries on screen (nested paths like `items[0].sku`, with sample values) plus a manual path input as an escape hatch. Up to 4 pinned columns, persisted per project + event in localStorage. Frontend-only — no API changes.

PR #362, opened 2026-07-16
ci(compose-security): cache Docker toolchain layers and prebuild with fast profile

dviejokfs:
## Why

**Compose Security is the slowest CI check** — ~74 minutes on the last main run (next slowest job finishes in ~35). The whole cost is building the production Docker image from scratch, uncached, on every run:

1. `cargo install wasm-pack` + `wasm-bindgen-cli` compiled from source on Alpine (~20-30 min)
2. Full 55-crate workspace `--release` build with `codegen-units = 1` (LTO-grade link times)
3. The `--mount=type=cache,target=/build/target` mount never persists across CI runners

## What

- **Dockerfile**: split the builder into a source-independent `toolchain` stage (apk deps, bun, wasm tooling, wasm32 target, zlib-static). Added `ARG TEMPS_BUILD_PROFILE=release` so a build can pick a different cargo profile; default is unchanged, so release/production images are identical to before.
- **Workflow**: the job now warms a GHA layer cache for the `toolchain` stage (only invalidates when the Dockerfile changes, and `cache-to` never exports the churning source-dependent layers), then pre-builds the image with the existing `fast` profile (inherits release, codegen-units=16, no LTO) and tags it `temps-compose-security:ci`.
- **Script**: uses the existing `COMPOSE_SECURITY_OVERRIDE` hook plus a new `COMPOSE_SECURITY_PREBUILT` env var to point the `temps` service at the prebuilt image and skip the uncached in-compose `--build`. Running the script locally with no env vars behaves exactly as before.

## Expected impact

- First run on this branch: pays the toolchain build once, but the `fast` profile alone should cut the rust build substantially.
- Steady state: toolchain layers come from cache (~1-2 min restore) and only the WASM/web/cargo-fast build runs — roughly **25-30 min instead of 74**.

## Not done (deliberately)

- Persisting the cargo `target/` cache mount (buildkit-cache-dance) would make the rust build incremental, but a multi-GB cache entry risks evicting other workflows' caches (that eviction already bit e2e-tests once). Can be a follow-up if the steady-state time is still too high.

## Test coverage semantics

The job still builds the image from the PR's source (Dockerfile + code), still exercises all the same compose security assertions, still verifies `/readyz`, admin bootstrap, secret non-leakage, etc. The only behavioral difference in what's tested is the cargo profile of the binary under test (`fast` vs `release` — same semantics, different codegen parallelism/LTO).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -81,3 +81,11 @@ docker-compose.*.yml
 .eslintrc*
 .stylelintrc*
 thoughts/
+
+# CI compose-security prebuilt-binary pipeline
+ci-build-cache/
+**/node_modules/
+
+# CI compose-security prebuilt-binary pipeline
+ci-build-cache/
+**/node_modules/
--- a/.github/compose-security.prebuilt.yml
+++ b/.github/compose-security.prebuilt.yml
@@ -1,0 +1,6 @@
+# CI-only override for scripts/test-compose-security.sh: point the temps
+# service at the image pre-built (with layer caching) by the Compose Security
+# job instead of rebuilding it from source inside docker compose.
+services:
+  temps:
+    image: temps-compose-security:ci
--- a/.github/workflows/rust-tests.yml
+++ b/.github/workflows/rust-tests.yml
@@ -19,7 +19,39 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v7
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      # The toolchain stage (apk deps, bun, wasm-pack/wasm-bindgen-cli compiled
+      # from source) only changes when the Dockerfile does, so cache it as
+      # image layers. Kept as a separate build so cache-to never exports the
+      # huge source-dependent layers of the full image on every commit.
+      - name: Warm Docker toolchain layer cache
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          target: toolchain
+          cache-from: type=gha,scope=compose-security-toolchain
+          cache-to: type=gha,scope=compose-security-toolchain,mode=max
+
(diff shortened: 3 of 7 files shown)


dviejokfs:
**v2 — binary now compiled outside Docker.** The first approach still took 1h15m: layer caching fixed the toolchain cost, but the cargo target dir inside a Docker build can never persist across runners, so the workspace build was cold every time.

New shape of the job:
1. Toolchain stage builds/loads from the GHA layer cache (~2s warm) and doubles as the **build container**.
2. WASM + web UI + `cargo build --profile fast` run inside that container (same musl toolchain, same steps as the Dockerfile builder stage) with `target/` and the cargo registry mounted from `actions/cache` keyed on `Cargo.lock` → **incremental rebuilds**.
3. The production runtime stage is assembled around the prebuilt binary via a new `TEMPS_ARTIFACTS` build-arg stage selector (~1 min). Default `docker build` / `docker compose build` from source is unchanged and never requires `ci-prebuilt/`.

Also fixed the failed run's telemetry assertion: the ENABLED/DISABLED notice is logged before the tracing subscriber initializes, so it never appears in `docker logs` — the check now inspects the container env for `TEMPS_TELEMETRY=0` instead. (All other security assertions passed on the failed run, including the fast-profile binary serving `/readyz`.)

Expected: cold run ≈ same as before; warm runs ≈ **10-20 min** (bun installs + incremental cargo + image assembly + validation). Trade-off: the cargo cache entry is a few GB — if it starts evicting other workflows' caches we can scope it down.

Verified locally: `bash -n` on the script, both Dockerfile variants pass `buildx --call=check`, and the prebuilt path was smoke-built with a stub binary (correct ownership, perms, GeoLite2 placement, uid 1001).

PR #364, opened 2026-07-16
perf(metrics): time-bound external-service latest-metric queries

dviejokfs:
## Problem
`GET /external-services/{id}/metrics/latest` and `/metrics/by-database` were very slow for services with metrics history. `query_latest`, `query_latest_by_label`, and `min_label_key_count` scanned the source's **entire retention history**: the label-key-count ordering (and the `labels ? key` predicate) prevent the planner from satisfying `DISTINCT ON` with a backwards index scan, so every chunk is read and the `jsonb_object_keys` correlated subquery is evaluated per row. At a 10–30s scrape interval that's millions of rows per request.

## Fix
Time-bound the scan to a 15-minute window anchored to `service_metrics_status.last_received_at` — an O(1) row upserted on every `write_batch` — so TimescaleDB chunk exclusion applies while stale/paused scrapers still return their last values. `COALESCE(..., now())` covers a missing status row.

## Verification
Seeded a source with **3.6M rows** (30 days @ 15s, 13 metrics, 8 per-`datname` series) in a local TimescaleDB:

| Query | Before | After |
|---|---|---|
| `query_latest` (EXPLAIN ANALYZE) | 5628 ms | 6 ms |
| `query_latest_by_label` (EXPLAIN ANALYZE) | 1916 ms | 1.3 ms |
| End-to-end via `TimescaleMetricsStore` | — | 13 ms / 11 ms |

Identical result sets before/after (13 metrics; 8 labelled series). All 99 `temps-metrics` unit tests pass.

## Load/saturation note
Read-path only; no new writes. Worst case per request drops from O(full retention) rows to O(15 min window).

### DIFF
--- a/crates/temps-metrics/src/store/clickhouse.rs
+++ b/crates/temps-metrics/src/store/clickhouse.rs
@@ -431,7 +431,7 @@ impl MetricsStore for ClickhouseMetricsStore {
             // read path expects.
             format!(
                 "SELECT \
-                   toUnixTimestamp(bucket) * 1000 AS bucket_ms, \
+                   toInt64(toUnixTimestamp(bucket)) * 1000 AS bucket_ms, \
                    greatest(bucket_max - lagInFrame(bucket_max, 1, bucket_max) \
                             OVER (ORDER BY bucket ASC), 0) AS avg_value \
                  FROM ( \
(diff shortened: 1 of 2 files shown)


dviejokfs:
Added ClickHouse coverage in 67f99696:

- Same time-bound applied to the ClickHouse store's `query_latest` / `query_latest_by_label` / `min_label_key_count` — `FINAL` over an unbounded WHERE dedup-merged up to the full 90-day TTL per request. Bounded to 15 min anchored on the source's `max(time)` (no status table in CH; the anchor is a cheap primary-key-prefix scalar subquery).
- Fixed a pre-existing `query_range` bug found while verifying: `toUnixTimestamp(...) * 1000` is `UInt64`, which the clickhouse crate refuses to deserialize into `bucket_ms: i64` — every raw-tier chart query errored. Cast via `toInt64()` in both gauge and monotonic branches.

Verified on an ephemeral `clickhouse-server:24.8` seeded with 3.6M rows: `query_latest` 375ms → 12ms, `query_latest_by_label` 197ms → 10ms (identical result sets), 30ms/12ms end-to-end through `ClickhouseMetricsStore`, and `query_range` returns 57 points instead of a schema-mismatch error.

PR #368, opened 2026-07-16
feat(proxy): trust CF-Connecting-IP from verified Cloudflare egress ranges

dviejokfs:
## What

Cloudflare-fronted deployments currently record the **Cloudflare edge IP** for every request: Pingora extracts the TCP peer and overwrites `X-Forwarded-For` with it, and nothing reads `CF-Connecting-IP`. That collapses rate limiting onto a handful of shared edge IPs, degrades audit forensics, and breaks geo attribution. Follow-up to #363, which made the console side trust-aware.

## Change

- New `cloudflare_ips` module in `temps-proxy`: a process-wide trust store of Cloudflare's published egress ranges (`ArcSwap<Vec<IpNetwork>>`, same pattern as `ip_access_control_service`).
  - **Vendored builtin list** (15 v4 + 7 v6 CIDRs from cloudflare.com/ips, 2026-07-16) seeds the store — works offline, self-hosted instances never phone home unless Cloudflare traffic is actually observed.
  - **Lazy daily refresh**: on the first connection whose *peer* is inside the (builtin) ranges, a background task starts fetching `api.cloudflare.com/client/v4/ips` every 24h (1h retry on failure). A failed fetch or malformed/empty payload **never replaces the last known-good set** — the builtin list is the permanent floor. Detection is keyed on the peer address, never on headers.
- Request path: both client-IP extraction sites now share one helper. If the peer is a verified Cloudflare address and `CF-Connecting-IP` parses as a bare `IpAddr`, that's the client; otherwise the peer. Direct callers sending the header are ignored — same trust model as the loopback-gated XFF handling in `temps_core::resolve_client_ip`.
- The real client IP then flows into the `X-Forwarded-For` we set upstream, so the console's audit/rate-limit/metadata paths (#363) pick it up with no changes.
- **Root-cause fix for IPv6 peers**: the old `split(':')` port-stripping mangled IPv6 addresses (`[2001:db8::1]:443` → `[2001`). The shared helper uses `as_inet()/ip()`.

## Hot-path budget (per CLAUDE.md)

Per request: one lock-free `ArcSwap::load` + linear scan over ~22 networks + one header parse — no locks, no allocation beyond the final `String` (which existed before), no I/O. Refresh is one HTTPS request/day in a background task. At saturation the refresher degrades to the last known-good (or builtin) list; stale-list failure mode is recording the edge IP, i.e. today's behavior.

## Tests

- 7 new unit tests: builtin list parses fully, known CF IPs match (v4+v6), spoofed header from untrusted peer ignored, valid header from CF peer honored, malformed headers (comma lists, ports, XSS strings) fall back to peer, endpoint JSON shape parses, degenerate refresh payloads rejected.
- `cargo test --lib -p temps-proxy`: 417 passed. Clippy `--all-targets -D warnings` and rustfmt clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12616,6 +12616,7 @@ dependencies = [
  "rand 0.8.6",
  "rcgen",
  "regex",
+ "reqwest 0.12.28",
  "rustls",
  "rustls-pemfile",
  "scraper",
--- a/crates/temps-proxy/Cargo.toml
+++ b/crates/temps-proxy/Cargo.toml
@@ -61,6 +61,7 @@ arc-swap = "1"
 moka = { version = "0.12", features = ["future"] }
 ipnetwork = "0.21"
 regex = { workspace = true }
+reqwest = { workspace = true }
 once_cell = { workspace = true }
 woothee = "0.13"
 rustls = "0.23"
--- a/crates/temps-proxy/src/cloudflare_ips.rs
+++ b/crates/temps-proxy/src/cloudflare_ips.rs
@@ -1,0 +1,332 @@
+//! Trust store for Cloudflare egress IP ranges.
+//!
+//! SECURITY: `CF-Connecting-IP` is honored ONLY when the immediate TCP peer is
+//! inside Cloudflare's published egress ranges. The header itself must never
+//! influence any decision (including starting the refresher) — anyone can send
+//! it directly to the listener; only the peer address is trustworthy. When the
+//! peer is not a Cloudflare address the header is ignored entirely and the
+//! socket IP is used, exactly like the `X-Forwarded-For` handling on the
+//! console side (`temps_core::resolve_client_ip`).
+//!
+//! The trust set is seeded with `BUILTIN_RANGES`, vendored from
+//! <https://www.cloudflare.com/ips/> at build time, so the check works offline
+//! and self-hosted instances never phone home unless Cloudflare-fronted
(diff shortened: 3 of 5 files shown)


dviejokfs:
## Security audit (security-auditor agent): APPROVE-WITH-CONDITIONS

Addressed in `57ab923c`:
- **[MEDIUM] IPv4-mapped IPv6 canonicalization** — `::ffff:1.2.3.4` now collapses to `1.2.3.4` for both peer and header, so v4 blocklist/rate-limit rules can't be evaded via the mapped form.
- **[LOW] Overbroad refresh payloads** — fetched CIDRs broader than /8 (v4) or /19 (v6), or lists >200 entries, are rejected wholesale; last known-good set is kept.
- **[LOW] Unbounded body read** — refresh response is streamed with a 256 KB cap.

Open condition (decision needed before merge):
- **[HIGH] Always-on trust of shared Cloudflare egress ranges.** Any Cloudflare customer can point their own zone/Worker at a directly-reachable origin and arrive from a CF egress IP, so `CF-Connecting-IP` is honored for foreign zones too. Mitigating factors: Cloudflare itself sets the header (customers cannot forge `CF-*` headers via Workers/Transform Rules), so the recorded value still reflects the true connecting client — arguably strictly better than today's edge-IP recording. Auditor recommends gating behind an operator opt-in settings column (default off) and/or Authenticated Origin Pulls binding. Needs a product call: default-on (better data, larger trust surface) vs opt-in toggle (+ migration, settings UI).

PR #399, opened 2026-07-20
feat(web): add metrics storage backend selector to monitoring settings

dviejokfs:
## Problem

The **Settings → Metrics Monitoring** page showed the metrics storage backend as a read-only badge, with copy claiming ClickHouse "activates only when the `TEMPS_CLICKHOUSE_*` env vars are set". Both halves of that were a problem:

1. **The copy is wrong.** The runtime uses ClickHouse for metrics only when the env vars are set **and** `monitoring.store == click_house` (`with_effective_store`, `serve/console.rs`). An operator who just enabled ClickHouse via env vars (which flips traces/request logs immediately) sees the metrics badge stubbornly say TimescaleDB with an explanation that implies their env vars didn't work.
2. **There was no control.** `monitoring.store` could only be changed via a raw settings PATCH — no console path existed. The page even ships a "Configured backend is not active" mismatch alert for a state the UI itself could never produce or resolve. Self-hosted users debugging alone had no way forward from that screen.

## Change

- Replace the read-only badge row with a **TimescaleDB / ClickHouse select** bound to `monitoring.store` (same `setValue` pattern as the scrape-interval select; saved by the existing form submit).
- Keep an explicit **"Active: …"** badge alongside, still driven by the server-reconciled `effective_metrics_store`, so configured-vs-actual stays visible.
- Fix the copy: ClickHouse is used when **this setting selects it and** the env vars are set; note that changes take effect **after a server restart** (the metrics store is wired at serve startup).
- Row layout follows the mobile pattern (`flex-col` → `sm:flex-row`).

The existing mismatch alert now has a real lifecycle: user selects ClickHouse without env vars → alert explains → user either sets the env vars or switches back — all from the UI.

## Testing

- `eslint` clean on the changed file; `tsc --noEmit` clean (repo-wide pre-existing lint errors on main are untouched).
- Frontend-only: the settings PATCH already persists `monitoring.store` (existing handler tests set it), and `effective_metrics_store` reconciliation is unchanged.

### DIFF
--- a/crates/temps-config/src/handler.rs
+++ b/crates/temps-config/src/handler.rs
@@ -1009,19 +1009,23 @@ fn validate_monitoring_settings(monitoring: &MonitoringSettings) -> Result<(), P
             .detail("monitoring.retention_daily_years must be between 1 and 10")
             .build());
     }
-    if monitoring.store == MetricsStoreKind::ClickHouse {
-        match &monitoring.clickhouse_url {
-            None => {
-                return Err(ErrorBuilder::new(StatusCode::BAD_REQUEST)
-                    .detail("monitoring.clickhouse_url is required when store is ClickHouse")
-                    .build());
-            }
-            Some(url) if url::Url::parse(url).is_err() => {
-                return Err(ErrorBuilder::new(StatusCode::BAD_REQUEST)
-                    .detail("monitoring.clickhouse_url is not a valid URL")
-                    .build());
-            }
-            _ => {}
+    // `monitoring.clickhouse_url` is legacy, optional config: the runtime
+    // builds the ClickHouse metrics store from the server's TEMPS_CLICKHOUSE_*
+    // env configuration (`build_ch_metrics_store`), never from this setting.
+    // Requiring it when store == ClickHouse made the store unswitchable from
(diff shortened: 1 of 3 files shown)


dviejokfs:
Live-tested end-to-end on a local server started from this branch (ClickHouse env vars set, web dev server serving this branch's UI):

1. Card renders the new select + "Active: TimescaleDB" badge with the corrected copy ✅
2. Selecting **ClickHouse** → Save initially returned **400 `monitoring.clickhouse_url is required when store is ClickHouse`** — a stale validation from when the URL lived in settings. The runtime builds the metrics store from `TEMPS_CLICKHOUSE_*` env config only (`build_ch_metrics_store`), so the requirement made the store unswitchable from the UI. Fixed in `1e1efdf7` (URL now validated only when supplied; +2 tests).
3. After the fix: select ClickHouse → Save → toast, `PUT /api/settings` 200, reload shows **"Active: ClickHouse"** (server-reconciled `effective_metrics_store`), no mismatch alert since env vars are set ✅
4. Switched back to TimescaleDB → Save → badge returns to "Active: TimescaleDB" ✅

PR #400, opened 2026-07-20
feat(sandbox): Firecracker microVM backend alongside Docker (ADR-029)

dviejokfs:
## Summary

Adds a **Firecracker microVM sandbox backend** that coexists with Docker behind the same `SandboxProvider` seam and `/v1/sandboxes` API (ADR-029). Docker remains the default everywhere; Firecracker is opt-in per host and selectable per sandbox — existing integrations are unaffected.

## What's included

**Provisioning** — `temps firecracker setup` (preflight, pinned firecracker+jailer/kernel with sha256 verification, bridge+NAT+TAP pool, jailer uids, smoke-test boot; `--check`/`--network-only`/`--uninstall`).

**Runtime** — `temps-vm-agent` (static-musl PID-1 guest agent over vsock); `FirecrackerSandboxProvider` (OCI image → ext4 conversion with digest-keyed cache, per-VM lifecycle, TAP networking + guest DNS, exec/fs over vsock); `RoutingSandboxProvider` dispatching on a typed `SandboxHandle.backend`.

**API / data** — optional `backend` on create (400 on unavailable); `sandboxes.backend` column + effective image recorded; per-sandbox `disk_size_mb` with **live grow-only resize** (`POST /…/resize`, ~2s via stop→grow→start, data preserved); rootfs GC keeping only cache entries backing a live VM (`GET/POST /…/rootfs[/gc]`); operations **timeline** (`sandbox_events` + `GET /…/events`) of lifecycle ops only (never shell).

**Web** — backend badge in the sandbox header, Identity shows image/disk, Timeline card, disk Resize control; DESIGN.md switched pages to full-width and SandboxDetail made responsive.

## Testing

Verified live on a self-hosted instance: full sandbox lifecycle (create/exec/fs/stop-start/destroy) on real KVM microVMs, egress through TAP/NAT, `backend`/image/disk shown in UI, rootfs GC reclaiming unreferenced caches, timeline recording ops, and live disk resize (2048→4096 MB in 1.6s with a marker file surviving). KVM-gated e2e test in `crates/temps-agents/tests/firecracker_e2e.rs`.

## Notes for reviewers

- Deliberately deferred (documented in ADR/code): jailer hardening, ADR-013 egress-proxy parity, snapshot-based pause, OCI ENV/ENTRYPOINT replay, no-reboot resize (needs in-guest resize2fs).
- Docker is unchanged unless an operator runs `temps firecracker setup`; the routing provider is a pass-through with a single backend.


dviejokfs:
Addressed the review in `eaaf84cd`:

**CI (was red):** fixed the `preview_auth`/`temps-config` test helpers and several `--all-targets`-only test literals missing the new `backend`/`disk_size_mb` fields; `cargo fmt`; and cfg-gated `temps-vm-agent` (Linux-only AF_VSOCK/reboot) so `cargo check --workspace` is green on macOS. `--workspace --all-targets` + `fmt --check` now pass locally.

**Backend guard:** added `SandboxProvider::supports_backend`; the service now returns 400 for a backend the host can't provide instead of silently downgrading to Docker.

**OpenAPI/SDK:** registered resize/events/rootfs/rootfs-gc in the ApiDoc (paths + components + `ToSchema`), extended the guardrail test, regenerated the web client, replaced the hand-rolled fetches in `SandboxDetail` with the generated SDK, and added `resize()`/`events()` + fields to the node SDK.

**Security:** `env.json`/VM dir now 0600/0700; `/rootfs` + `/rootfs/gc` require admin; `network_mode: "restricted"` fails closed until ADR-013; setup drops guest→host + cloud-metadata (169.254.169.254) and isolates TAP ports.

**Correctness/perf:** GC now holds the cache lock across conversion→copy→marker (no delete-mid-clone race); `docker export` streams to disk + extraction on `spawn_blocking`; `/events` is bounded; routing dispatches on the typed handle backend; nits cleaned up.

PR #407, opened 2026-07-21
feat(audit): add PII scrub endpoint for audit log data payloads

### LINKED ISSUE #405: Audit log data payloads retain actor PII (username/name/email) after account deletion
## Context

Since #386, deleting a user account preserves their audit rows (`audit_logs.user_id` becomes `NULL` instead of cascading the delete). This is the right default for an audit trail — but it makes an existing property permanent: the serialized `data` payload on each row keeps whatever actor/target details the operation recorded (`username`, `name`, `email` — e.g. `UserDeletedAudit`, `UserUpdatedAudit`, `EmailVerifiedAudit`), and that content is returned verbatim by `GET /audit/logs`.

The migration doc for `m20260720_000001_audit_logs_keep_history_on_user_delete` states this retention is intentional. What's missing is a mechanism for operators who need stricter erasure (e.g. GDPR right-to-erasure requests) to comply without destroying the structural audit record.

## Proposal

A PII-scrub step for audit `data` payloads, decoupled from row deletion:

- On user deletion (or on demand, e.g. a CLI/API `scrub` action), JSON-patch the known PII fields (`email`, `name`, `username`) out of the `data` payloads of that user's audit rows, replacing them with a redaction marker.
- Keep the row itself, `operation_type`, timestamps, and non-PII context intact so the trail still reconstructs *what happened*.
- Off by default; explicit operator action (self-hosted deployments own their compliance posture).

## Notes

- Flagged by security review of #386 (severity: major — policy gap, not a code bug).
- Any plugin that fingerprints or hashes audit row content needs to account for scrubbing being a legitimate post-hoc mutation of `data`.

dviejokfs:
Closes #405.

## What

Since #386, audit rows survive user deletion — which also means the `data` payloads keep the deleted account's `email`/`username`/`name` indefinitely and return them via `GET /audit/logs`. This adds the erasure mechanism that issue calls for:

- **`POST /audit/logs/scrub`** — explicit operator action (never automatic). Takes `{email?, username?, name?}` (at least one, each ≥3 chars) and replaces every payload string value **exactly matching** an identifier (case-insensitive, full-value match) with `[REDACTED]`, recursing through nested objects/arrays. Covers both rows the user authored and rows *about* them (e.g. the `USER_DELETED` event recorded under the acting admin). Structural record — operation type, timestamps, non-matching context — is preserved.
- **New `audit:write` permission**, granted to `admin` and `platform_admin` only.
- **The scrub audits itself**: an `AUDIT_DATA_SCRUBBED` entry records which identifier kinds were provided and row counts — never the values being erased.
- Non-JSON payloads are skipped with a warning rather than destroyed. Rows are scanned in id-ascending pages of 500: constant memory, and concurrent inserts (higher ids) can't shift the scan.
- Regenerated web SDK committed (`scrubAuditLogs` + types); `tsc --noEmit` clean.

Note for content-integrity tooling: a plugin that fingerprints audit payloads will report scrubbed rows as modified — an in-place redaction is exactly that, and the `AUDIT_DATA_SCRUBBED` entry documents why.

## Evidence (live dev instance, branch build)

Inserted a synthetic post-deletion row (`user_id = NULL`, payload with `ghost.gdpr@example.com`), then:

```
POST /api/audit/logs/scrub {"email":"ghost.gdpr@example.com","username":"ghost.gdpr","name":"Ghost Gdpr"}
→ 200 {"rows_scanned":109,"rows_scrubbed":1}

SELECT data FROM audit_logs WHERE id = 109;
→ {"context":{"user_id":1,"user_agent":"evidence-agent"},"target_user_id":9999,
   "username":"[REDACTED]","email":"[REDACTED]","name":"[REDACTED]"}

SELECT operation_type, data FROM audit_logs ORDER BY id DESC LIMIT 1;
→ AUDIT_DATA_SCRUBBED | {"identifier_fields":["email","username","name"],"rows_scanned":109,"rows_scrubbed":1}
```

Failure paths:

```
{"username":"ab"}  → 400 "Identifier of length 2 is too short to scrub safely (minimum 3 characters)"
{}                   → 400 "At least one identifier value to redact is required"
reader-role API key  → 403 "This operation requires the audit:write permission"
```

Unit tests: recursive redaction (case-insensitivity, arrays, substring non-matches, non-string values), validation rejections, and a MockDatabase scan asserting only matching rows are updated and unparsable payloads are skipped. `cargo test --lib -p temps-audit`: 8 passed. Full workspace `cargo check --lib` clean; clippy clean.


dviejokfs:
**Security review (security-auditor agent): APPROVE-WITH-CONDITIONS — all conditions now addressed in `e46bba48`:**

- **Blocking (fixed):** a scrub failing mid-scan left already-rewritten pages with no audit trail. The handler now writes an `AUDIT_DATA_SCRUB_STARTED` entry before any row is mutated (failing closed if that write fails), making the `AUDIT_DATA_SCRUBBED` entry a completion receipt — a start with no receipt means an interrupted scrub. Validation moved ahead of the start entry so invalid requests leave no dangling starts.
- **Minor (fixed):** identifiers capped at 512 chars; the scrub's own meta records are excluded from the scan (prevents redacting attribution out of the records documenting a scrub); negative-assertion tests pin `audit:write` to admin/platform_admin only.
- **Minor (accepted, not implemented):** per-user rate limiting on the endpoint. Admin-only + start/receipt audit entries bound the abuse case; can be added if operators report load issues.

The auditor separately confirmed: no injection paths (Sea-ORM typed queries only), no PII in either meta entry, and correct permission wiring.

`5c9fd75f` also fixes the `Web Dependency Audit` failure — two freshly published high advisories in dev-tooling transitive deps (GHSA-3jxr-9vmj-r5cp brace-expansion, GHSA-52cp-r559-cp3m js-yaml), pre-existing on main, pinned to patched releases via the existing `overrides` block. `bun audit --audit-level=high` now exits clean locally.

dviejokfs:
Closing unmerged after a needed-ness review. The scrub endpoint was an escalation of an auditor suggestion, not a user or compliance request, and it conflicts with the tamper-evidence design of plugins that fingerprint audit payload content: an in-place redaction is indistinguishable from tampering there, so shipping this first bakes in a verification break. The dependency fix was extracted to #409. Erasure remains achievable by the operator with a direct SQL `UPDATE` against their own database; see #405 for the re-scoped follow-up.

PR #411, opened 2026-07-21
feat(audit): support events whose actor has no resolvable account

clau1902:
## Summary

Foundation for the auth-failure auditing series (1 of 3, followed by the failed-login/MFA PR and the permission-denial PR).

Changes the single `AuditOperation::user_id()` contract from `i32` to `Option<i32>`. Existing authenticated operations return `Some(user_id)` and remain behaviorally unchanged; events without a resolvable user return `None`. The audit service persists this value directly.

This lets upcoming failure events (login attempts against unknown emails, deployment-token denials) record a NULL actor instead of inventing `user_id 0`, which the audit FK rejects. There is one source of truth: no compatibility accessor or sentinel user ID.

The branch has been rebuilt on current `main`; the audit-history migration already merged in #386 is no longer duplicated here.

## Evidence

**Optional actor persistence:** targeted regressions prove both nullable and known actors are bound correctly by the audit insert.

```bash
cargo test --lib -p temps-audit test_create_audit_log_persists -- --nocapture
```

```text
cargo test: 2 passed, 4 filtered out (1 suite, 0.00s)
```

**Affected crates:** all library tests pass, including the two DNS-dependent `temps-core` tests when run with network access.

```bash
cargo test --lib -p temps-audit -p temps-core
```

```text
cargo test: 248 passed (2 suites, 0.05s)
```

**Workspace compilation:**

```bash
cargo check --lib
```

```text
cargo build: 0 errors, 5 existing workspace/toolchain warnings (957 crates)
```

**Formatting:**

```bash
cargo fmt --all -- --check
```

```text
exit 0
```


dviejokfs:
Updated this branch with current `main` (merge, no history rewrite) and pushed one follow-up commit.

**Why it was needed:** the branch was 130 commits behind `main`. GitHub reported it as `MERGEABLE`/`CLEAN` and CI was 16/16 green, but both were misleading — CI ran against the stale base, and git merges the affected files without textual conflict. Seven `AuditOperation` impls were added to `main` after this branch was cut and still returned `i32`:

- `temps-import/src/handlers/audit.rs` (crate is entirely new)
- `temps-providers/src/handlers/pg_stat_statements_handlers.rs`
- new impls inside `temps-agents/handlers/definitions.rs`, `temps-environments/handlers/audit.rs`, `temps-notifications/handlers.rs`, `temps-providers/handlers/audit.rs` (×2)

Merging as-is would have broken `main` with `E0053`. All seven got the same mechanical conversion as the rest of the series (`Some(self.context.user_id)`).

**Verification after the update:**

```text
cargo check --lib --workspace          0 errors (1018 crates)
cargo clippy --workspace --all-targets -- -D warnings    0 errors
cargo fmt --all -- --check             exit 0
cargo test --lib -p temps-audit -p temps-core    282 passed
```

**Runtime check against real Postgres** (fresh DB, migrations applied by this branch) — confirms the nullable-actor column and that a known actor still persists correctly, which the MockDatabase tests can't prove:

```text
\d audit_logs
 user_id | integer | nullable
 fk_audit_logs_user_id FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE SET NULL

POST /api/auth/login (valid credentials) -> 200
 id | user_id | operation_type
  1 |       1 | LOGIN_SUCCESS
```

PR #419, opened 2026-07-22
feat(error-tracking): source context for native stack traces (Go/Rust/all languages)

dviejokfs:
## What & why

Error-tracking source display currently only works for **JavaScript source maps** (the original source travels inside the `.map` via `sourcesContent`). Native SDKs (Go `sentry-go`, Rust `sentry-rust`, Python, …) emit frames that already carry the **original** `filename`+`lineno` but ship no source, so the Stack Trace panel shows frames without code.

This makes source context work for **any language** by storing raw source per release and attaching `context_line`/`pre_context`/`post_context` to native frames — reusing the existing symbolication pass and the `StackTrace.tsx` renderer **unchanged**.

## Design (mostly reuse)

- **`extract_source_context`, `ResolvedFrame`, candidate/suffix path matching, the frame model, and the renderer are all language-agnostic already.** The only new thing is a raw-source *provider* that skips source-map token lookup (native frames need no un-minifying) and slices raw source directly.

## Changes

- **entities/migrations**: new `source_files` table keyed by `(project, release, file_path)`; opt-in `projects.error_source_context_enabled` (default `false`).
- **service** (`SourceMapService`): `upload_source_file` / `list_source_files` / `delete_source_files_for_release` + `resolve_native_frame`. Native resolution is **gated per project** and the flag is read **once per event**, not per frame.
- **handlers**: `POST/GET/DELETE /projects/{id}/releases/{release}/source-files`, gated on the toggle so application source is only stored when explicitly opted in.
- **settings**: `error_source_context_enabled` threaded through the project-settings update path + a toggle card in `GeneralSettings`.
- **deployments (release keying)**: image deploys now inject `SENTRY_RELEASE`/`OTEL_SERVICE_VERSION` from the deployed **image tag/digest** (previously only git deploys set a release, so image deploys reported a useless/`dev` release). This makes events attributable to the exact image and gives a stable **join key** for source uploads.

## Opt-in

Off by default. When enabled, users upload source per release (CLI/API, keyed by the deployed commit/tag) and Temps renders code around each frame. Source is only accepted/stored while enabled.

## Tests

- `image_ref_release`: tag / digest / registry-port disambiguation.
- `build_native_resolved`: context slicing + out-of-range guard.
- `cargo check --all-targets`, `cargo fmt --check`, and `cargo clippy` all clean.

## ⚠️ Remaining before merge (why this is a draft)

- [x] **Regenerate the web SDK** (`bun run openapi-ts`) so the new `error_source_context_enabled` field and source-file endpoints typecheck — the `GeneralSettings` toggle references generated types not yet present.
- [x] **Integration tests** for `resolve_native_frame` against a `TestDatabase` (exact-path + suffix match + disabled-project skip).
- [x] **CLI upload command + CI docs example** (`errors source-files upload|list|delete`, single-file or recursive `--dir`).
- [x] **Auto-generate the source bundle for git-based Temps builds** — `CaptureSourceFilesJob` (gated on the opt-in), plus `delete_stale_source_files` GC.
- [x] **Guidance for compiled services (Go, Rust, …)** shipped in the docs (the compiled-binary release trap); the per-app code/CI change lives in each app repo, let the injected `SENTRY_RELEASE` win instead; bake the real commit/tag via a build arg in CI.
- [x] **ADR-033** — release-identity invariant (binary release == source key == platform-injected value).


dviejokfs:
Web SDK regenerated against a server built from this branch (`uploadSourceFile`/`listSourceFiles`/`deleteReleaseSourceFiles` + `error_source_context_enabled` on `ProjectResponse`/`UpdateProjectSettingsRequest`). `tsc --noEmit` is clean, so the GeneralSettings toggle typechecks. Marking ready for review.

Verified end to end against the local dev server (:8091): the 2 migrations apply, `source_files` table + `projects.error_source_context_enabled` column exist, and the new endpoints/field are present in the live OpenAPI spec.

Remaining as non-blocking follow-ups: DB-integration tests for `resolve_native_frame`, a CLI `source-files` upload command + CI example (and auto-bundling for git-based builds), and the app-side `SentrySetup` release change for image-deployed native SDK apps that hard-code the release.

dviejokfs:
Knocked out the in-repo non-blocking follow-ups:

- ✅ **Integration tests** (`TestDatabase`): opt-in flag reader, native resolution when enabled (exact + suffix match), and disabled-project skip. All green.
- ✅ **CLI upload command**: `temps errors source-files upload|list|delete` (alias `sf`). `upload` takes a single `--file` or a recursive `--dir` tree, keying each file by its path relative to the root — matching how `-trimpath` Go (and most native SDKs) report frame filenames.
- ✅ **CI example + docs**: new "Source Context" section in `docs/features/error-tracking` with a copy-paste CI upload snippet and the release-keying note; removed the stale source-maps WIP callout.

Still open (larger, better as its own change): auto-generating the source bundle during git-based Temps builds (zero-config upload), and the app-side `SentrySetup` release change for native SDK apps that hard-code the release.

dviejokfs:
All three remaining follow-ups are now in (commit f53905b5):

- ✅ **Auto-capture on git builds** — `CaptureSourceFilesJob` uploads raw source from the `DownloadRepoJob` checkout, keyed by the release. Native source lives in the *checkout*, not the compiled image (the binary strips it) — that's the key difference from `CaptureSourceMapsJob`, which extracts `.map`s from the image. Gated on the `error_source_context_enabled` opt-in at plan time (no-op otherwise), bounds file count (5000) and size (2 MB), skips build/dep/VCS dirs, normalizes paths to the module root to match `-trimpath` frames. Ships with `delete_stale_source_files` GC (closes the unbounded-storage gap).
- ✅ **Compiled-service guidance** — docs now cover the trap where Go/Rust bake `ClientOptions.Release` into the binary and thus ignore the injected `SENTRY_RELEASE`.
- ✅ **ADR-033** — the release-identity invariant (SDK release == source-artifact key == injected `SENTRY_RELEASE`, all = the immutable build identity), with precedence chain, image tag/digest rule, and debug-id keying weighed as the long-term alternative.

The ADR also flags a pre-existing follow-up defect: `delete_stale_source_maps`/`delete_stale_source_files` build their keep-set from `commit_sha` only, so image-tag-keyed artifacts could be wrongly GC'd. Not triggered here (the native capture job is git-only, where release == commit_sha), but worth a dedicated fix.

dviejokfs:
**Tested end-to-end against a live server** (worktree build on :8080, dev DB) — the full loop works:

1. Enabled the opt-in on a project via the API (`error_source_context_enabled` → true).
2. Uploaded a Go source file for a release (via the API and the new CLI command).
3. Ingested a native Go error event (`POST /api/{id}/store/`) with a frame pointing at that file+line.
4. Fetched the stored event — the frame came back with real source context:

```
filename    : internal/gateway/middleware.go
lineno      : 5
context_line: '\tpanic("boom from middleware")'
pre_context : ['package gateway', '', '// Middleware ...', 'func Middleware() {']
post_context: ['}']
```

The e2e also **caught a real gap**: `errors source-files upload --dir` was uploading `node_modules`/dependency source. Fixed — the CLI now skips the same build/dep/VCS dirs as the auto-capture job (commit 31c0676b).

Also added **sample apps** (`examples/go/error-tracking`, `examples/rust/error-tracking`) — minimal Go/Rust apps that report an error on `GET /boom` and show the correct ADR-033 release setup (read `SENTRY_RELEASE` from env, don't hard-code it). Both verified to build.

Not yet exercised: the auto-capture job firing inside a real git deployment (needs a full deploy run), and the browser render (the e2e event is viewable at the project's error tracking page).

dviejokfs:
**Ran a full real git deployment e2e** (worktree build on :8080). Deployed a Go sample app from a public repo ([temps-examples#10](https://github.com/gotempsh/temps-examples/pull/10)) through the actual Temps deploy workflow:

- ✅ **`CaptureSourceFilesJob` ran in a real deployment** (scheduled because the opt-in was on), uploaded `~/main.go` keyed by the commit SHA `5f489111…`. It correctly walked only the `examples/go-error-tracking` module root (the `project_directory` handling).
- ✅ **Release-identity holds end-to-end**: the deployed container's injected `SENTRY_RELEASE` = `5f489111…` = the exact release the source was captured under.
- ✅ **Resolution against auto-captured source**: a native event at that release resolved to the captured `main.go` (`context_line: 'return fmt.Errorf("failed to do work: %w"…'`).

The e2e **caught a real robustness gap** and I fixed it (commit 4b452103): a binary built *without* `-trimpath` reports build paths like `/src/main.go`, but the old suffix-match only handled bare basenames, so prefixed paths missed. Now the resolver tries progressively shorter trailing suffixes (longest/most-specific first) — with a test. Also added `-trimpath` to the sample's Dockerfile.

Note: hitting the deployed app's `/boom` returns 500 (the app *does* capture), but the event doesn't reach Temps because the injected DSN host (`localho.st`) resolves to the container's own loopback on local dev — a container→host networking artifact, not a feature gap. The equivalent event sent from the host resolves correctly.

dviejokfs:
**Security-auditor pass done + audit logging added.** Two commits:

**Audit logging** (`28774a5f`): `upload_source_file` and `delete_release_source_files` now emit `SOURCE_FILE_UPLOADED` / `SOURCE_FILES_DELETED` audit events (actor + IP + user-agent), per the audit-all-writes rule.

**Security hardening** (`83fd07e2`) — the auditor returned REJECT with real findings; all HIGH + MEDIUM fixed:
- **HIGH** — no body limit on the upload route → added `DefaultBodyLimit` (10 MiB source-files / 50 MiB source-maps); closes an authenticated memory-exhaustion DoS.
- **HIGH** — `CaptureSourceFilesJob` followed symlinks → now skips symlinks *and* confines `project.directory` via the existing `canonicalize_confined_repo_path` (rejects `../` + symlink escape); closes host-file exfiltration (e.g. a repo symlinking to `/etc`).
- **MEDIUM** — `list_source_files` loaded the full `content` blob → now projects metadata columns only.
- **MEDIUM** — unescaped `%`/`_` in the trailing-suffix `LIKE` → escaped, so an attacker-controlled frame filename can't match unrelated stored files.

New tests cover the symlink-escape guard and the LIKE-metacharacter literal match. All green locally (clippy/fmt clean).

The auditor also flagged one **LOW** (re-check the opt-in flag inside the service upload method) — I deliberately skipped it: the capture job uploads up to 5000 files/deploy, so a per-file flag query is a real DB-load regression, and the flag is already enforced at the HTTP handler + at plan time.

Auditor-confirmed SAFE: authZ/IDOR on all three endpoints, no SQL injection, source content never returned to callers, server-side opt-in gate, cascade-delete + stale cleanup, extension allowlist excludes `.env`.

PR #426, opened 2026-07-23
fix(observability): read Observe feed through ClickHouse-aware storage backends

dviejokfs:
## Symptom

On a server with `TEMPS_CLICKHOUSE_*` configured, the project Observe page shows **Requests 0, Errors 0, empty event list** ("No events match the current filters") even for a live, deployed project — while the Requests page and Traces page show data fine.

## Root cause

When the opt-in ClickHouse telemetry backends landed (`1c3a8b70`), proxy logs and OTel spans started being **written to ClickHouse** via the pluggable `ProxyLogStorage` / `OtelStorage` backends — but `ObservabilityService` (the unified Observe merge service) was never migrated: it kept querying the `proxy_logs` table via Sea-ORM and `otel_spans` via raw Postgres SQL. On CH-enabled servers those Postgres tables are empty, so the feed silently returned nothing. Not a regression in any one commit to `temps-observability` — the reader was simply left behind when the write path moved.

## Fix

- Requests are fetched through `ProxyLogService` and spans through the registered `Arc<dyn OtelStorage>` — the same backend-dispatching readers the Requests/Traces pages use, selected at startup (ClickHouse when configured, TimescaleDB otherwise). Errors/revenue always live in Postgres and keep direct queries.
- Dependencies resolve in `configure_routes` (after all plugins register) because `ProxyPlugin` registers after `ObservabilityPlugin`.
- **Backend-agnostic row identities**: request rows are identified by `request_id`, spans by `{trace_id}:{span_id}` — ClickHouse tables have no serial PK (every row came back `id = 0`, colliding React keys and breaking `/full`). The `/full` endpoint gains an optional `ts` query hint bounding the lookup (chunk exclusion on TimescaleDB, partition pruning on CH) and enforces project tenancy server-side on the request path.
- New `TraceQuery::root_only` (both span backends: `parent_span_id IS NULL` / `''` sentinel) replaces the raw root-span SQL; new `ProxyLogsQuery::exclude_bots` implements the NULL-keeping hide-bots tri-state (`is_bot = false OR is_bot IS NULL`) on both backends.
- Web SDK regenerated (`RequestRow.id`/`FullRequest.id` → string, `ts` param). No UI component changes needed — ids were already treated opaquely; `tsc --noEmit` clean.

## Evidence

Local reproduction on a fresh instance + new ClickHouse container, seeded 40 proxy_logs rows + 24 spans (8 traces) into CH:

**Before** (`origin/main` binary, same data):
```
GET /api/projects/1/observe/events?kinds=request,error,revenue,span
{"events":[],"applied_kinds":["request","span","error","revenue"]}
```
…while `/api/proxy-logs?project_id=1` and `/api/otel/trace-summaries?project_id=1` returned the seeded CH rows — pinpointing the observability reader.

**After** (this branch, same seeded data):
```
events: 8 (page 1)
 request req-obs-0 GET /api/operations 200
 span    trace0000…:span000000000000 GET /api/operations OK
 …
```
Browser walkthrough of the Observe page: Requests card **33** (bots hidden — exactly the 7 seeded GPTBot rows excluded; toggling shows **40**), Traces card **8** after enabling the kind, spans interleaved with requests in ts-DESC order, search widening and side panel verified.

## Regression tests

- `tests/clickhouse_observe_test.rs` (new): real ClickHouse testcontainer, seeds through the production `write_batch`/`store_spans` write paths, asserts the merged feed returns requests + root spans, `hide_bots`, span-search widening, `/full` composite-id and request-id lookups, and cross-project denial. Skips gracefully without Docker. **This test fails on the pre-fix code** (feed comes back empty).
- `merge_integration.rs` updated to build the service over the TimescaleDB backends (the default-path wiring), all 9 passing.
- Unit tests: composite span identity mapping, CH `exclude_bots` clause, `map_request` request_id identity.

`cargo check --lib` (workspace), `cargo test --lib -p temps-otel -p temps-proxy -p temps-observability` (833+ tests), clippy `-D warnings` on all three crates: clean.


dviejokfs:
## Review pass (rust-standards + security-auditor + test-engineer)

**Verdict: APPROVE (with notes)** — no blocking findings; all majors and the security condition fixed in b42d52f5.

**Security — APPROVE-WITH-CONDITIONS → resolved.** Tenancy checks on both `/full` paths verified correct (post-storage `project_id` filter for requests; SQL-level scoping in `get_trace` on both backends); all new SQL fragments are constants, user values bound/escaped; auth guards intact. The one MEDIUM — `RequestStore`/`TraceStore` errors leaking backend details (CH host/port, SQL text) into 500 bodies — is fixed: full error logged server-side, generic detail returned.

**Rust standards — fixed the major:** the Observe feed no longer pays `list_with_filters`' pagination COUNT (exact hypertable `COUNT(*)` / CH `count() FINAL`) per poll for a discarded total — new count-free `ProxyLogStorage::list_page`, with the Sea-ORM filter builder extracted (`build_list_query`) so counted and count-free paths can't drift. Environment-filter omission for spans now documented as intentional.

**Test coverage — gaps closed on the TimescaleDB path:** `root_only` now falsified against an actual child span (+ search widening, + `fetch_full` composite-id happy path — which flushed out and documented the summaries-window behavior of TSDB `get_trace`), and the full `hide_bots` tri-state (hide / only-bots / off) asserted against bot / non-bot / NULL-`is_bot` rows.

**Deferred (noted, non-blocking):**
- CH `get_by_request_id` with no `ts` hint has no recent-window probe (pre-existing, PR #402 surface) — worth a follow-up together with rate-limiting the unbounded-`ts` lookup QoS concern the auditor rated LOW.
- The UI has no caller of `/full` yet; when "Show full" is wired, pass the row's `ts`.
- `is_bot=false&exclude_bots=true` on `/api/proxy-logs` yields a contradictory (empty) predicate — harmless; documented tri-state.

Re-verified after fixes: CH e2e regression test, 10 TSDB integration tests, 422 proxy unit tests, clippy `-D warnings`, live endpoint on the ClickHouse env returning the merged feed via the new `list_page` path.

PR #429, opened 2026-07-24
feat(otel): store cross-project trace refs in ClickHouse when enabled

dviejokfs:
## Problem

The `cross_project_trace_refs` Postgres control table (ADR-027 Phase 0) reached **125 GB** (39 GB heap + 87 GB b-tree indexes) in three weeks on a high-ingest install. The ADR's size estimate (~225 MB) did not hold, and the `(trace_id, project_id, first_seen)` tuples are exactly the shape ClickHouse compresses best — but the CH `spans` table can't serve trace-id-only discovery because its primary index is `(project_id, trace_id, span_id)`.

## Change

Ref storage now routes through the `OtelStorage` trait (`record_trace_refs` / `get_trace_ref_projects`):

- **TimescaleDB backend (default):** unchanged — Postgres control table + daily prune.
- **ClickHouse backend:** new `cross_project_trace_refs` CH table (`0006_trace_refs.sql`), `ORDER BY (trace_id, project_id)`, `ReplacingMergeTree` over an **inverted `_version`** so the Postgres first-write-wins / earliest-`first_seen` semantics survive merges. Per-row `retention_days` is stamped from the same `RetentionTable::Spans` resolver value as span rows — **ref TTL is locked to trace TTL by construction** (requirement: cross-trace retention must equal trace retention).

## No automatic migration

Reads union ClickHouse with legacy Postgres rows, so enabling CH needs no data migration and startup is unaffected: pre-cutover refs resolve from Postgres until the daily prune drains them (one retention window). Operators who want history copied immediately can use the new opt-in backfill domain:

```
temps backfill clickhouse --domain trace-refs
```

(also included in `--domain all`; keyset walk over `(first_seen, trace_id, project_id)`, idempotent re-runs via the inverted version).

The Postgres prune job keeps running unconditionally — on the CH backend it drains legacy rows and then becomes a no-op. ADR-027 amended accordingly.

## Evidence

- `cargo check --lib` clean; `cargo clippy -p temps-otel -p temps-cli --all-targets -- -D warnings` clean
- `cargo test -p temps-otel --lib`: 417 passed (incl. new merge-semantics, inverted-version, and service filtering tests over MockOtelStorage + MockDatabase)
- Real-ClickHouse testcontainer roundtrip (`trace_refs_roundtrip_record_and_lookup`): migration 0006 applies, record → dedupe on re-record → per-trace lookup → union with (empty) PG all verified against clickhouse-server 26.2.5

### DIFF
--- a/crates/temps-cli/src/commands/backfill.rs
+++ b/crates/temps-cli/src/commands/backfill.rs
@@ -54,10 +54,11 @@ pub enum BackfillTarget {
 
 /// Which observability domain to copy from TimescaleDB into ClickHouse.
 ///
-/// `events` replicates via the live outbox too, but the other three
-/// (`proxy-logs`, `traces`, `metrics`) are backend-swap domains with no live
-/// replication — enabling ClickHouse routes *new* writes there but leaves the
-/// historical TimescaleDB rows behind. This backfill copies that history over.
+/// `events` replicates via the live outbox too, but the others
+/// (`proxy-logs`, `traces`, `metrics`, `trace-refs`) are backend-swap domains
+/// with no live replication — enabling ClickHouse routes *new* writes there
+/// but leaves the historical TimescaleDB rows behind. This backfill copies
+/// that history over.
 #[derive(clap::ValueEnum, Clone, Copy, Debug, PartialEq, Eq)]
 pub enum BackfillDomain {
     /// Analytics events (`events` → ClickHouse `events`).
@@ -68,7 +69,12 @@ pub enum BackfillDomain {
     Traces,
     /// Resource metrics (`service_metrics` → ClickHouse `service_metrics`).
     Metrics,
-    /// All four domains, in turn.
+    /// Cross-project trace refs (`cross_project_trace_refs` → ClickHouse
+    /// table of the same name). Optional: the live read path unions both
+    /// backends, so skipping this only means pre-cutover refs resolve from
+    /// Postgres until they age out.
+    TraceRefs,
+    /// All domains, in turn.
     All,
 }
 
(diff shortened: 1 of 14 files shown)


dviejokfs:
## Live E2E verification (local instance, ClickHouse enabled)

Ran the worktree binary via the start-temps recipe (dev Postgres :15433, `TEMPS_DATA_DIR` from main repo) against the pre-existing `temps-observe-ch` ClickHouse container (:18124), which had migrations 0001–0005 but **not** 0006 — a clean upgrade test.

1. **Startup + migration — PASS**: log shows `applying ch-otel migration migration="0006_trace_refs"` → `applied=["0006_trace_refs"] skipped_count=5`; `system.tables` lists `cross_project_trace_refs`.
2. **Ingest → CH write — PASS**: OTLP traces sharing `trace_id=cafecafe…` ingested into projects 1 and 2 (`POST /api/otel/v1/traces`, `tk_` key + `X-Temps-Project-Id`); both pairs visible in CH within milliseconds:
   ```
   cafecafecafecafecafecafecafecafe  1  2026-07-24 07:31:46.746
   cafecafecafecafecafecafecafecafe  2  2026-07-24 07:31:46.768
   ```
3. **No new Postgres writes — PASS**: PG `cross_project_trace_refs` count 1603 → 1603; the test trace_id has 0 PG rows.
4. **Discovery read — PASS**: `GET /api/otel/traces/cross-project/{trace_id}` resolves the sibling in both directions (project 1 ↔ project 2, correct names/slugs/first_seen) through the union read path.

Test credentials and server were cleaned up afterwards; no schema changes to the shared dev DB (branch adds no PG migrations).

PR #430, opened 2026-07-24
fix(web): bump rrweb-player to 2.1.1 (broken 2.1.0 dist → blank session replays)

yash296:
## Problem

Session replays render as a **blank frame with no error** in the analytics console.

The root cause is the pinned player version. `web` depends on **rrweb-player 2.1.0**, whose *published* dist is a broken build: the compiled `onMount` never constructs the `Replayer`. In the shipped `rrweb-player.umd.cjs` there is no `new Replayer(...)` at all, and the controller's `play`/`pause` compile to empty functions. The component mounts an empty `.rr-player__frame` and stops — silently, with nothing logged. rrweb fixed this in **rrweb-player 2.1.1** (published 2026-07-21).

## Evidence

- **The recorded data is fine.** Feeding a real session's stored events to `rrweb.Replayer` — or to rrweb-player **2.1.1** — renders the page correctly.
- **The bundled 2.1.0 player is the failure point.** The same events, passed to a fresh `new rrwebPlayer({ ... })` with the exact props from `web/src/components/session-replay/SessionReplayPlayer.tsx`, produce the blank frame.
- **Broken-build signature:** the strings `new Replayer` and `insertStyleRules` are absent from `rrweb-player@2.1.0`'s `dist/rrweb-player.umd.cjs`; both are present in `2.1.1`.

## Change

- `web/package.json`: `rrweb-player` `^2.1.0` → `^2.1.1`.
- `web/bun.lock`: bumped the rrweb-family subtree that this pulls in — `rrweb-player`, `@rrweb/replay`, `@rrweb/packer`, `@rrweb/types`, `@rrweb/utils`, `rrweb`, `rrdom`, `rrweb-snapshot` — from `2.1.0` to `2.1.1`. The whole family releases in lockstep, so the dependency-tree shape is unchanged (only internal `^2.1.0` ranges move to `^2.1.1`); the integrities are the published npm hashes. Diff is exactly 9 lines.

No console/app code changes are needed — `SessionReplayPlayer.tsx` works unchanged once the player is fixed, and existing recorded sessions replay retroactively.

> Note on the lockfile: it was updated directly to the published 2.1.1 metadata rather than via `bun install` (no bun in my environment). If you'd prefer a tool-regenerated `bun.lock`, a quick `bun install` on this branch will produce an identical result.

### DIFF
--- a/web/bun.lock
+++ b/web/bun.lock
@@ -63,7 +63,7 @@
         "recharts": "3.9.2",
         "rehype-highlight": "^7.0.2",
         "remark-gfm": "^4.0.1",
-        "rrweb-player": "^2.1.0",
+        "rrweb-player": "^2.1.1",
         "slugify": "^1.6.9",
         "sonner": "^2.0.7",
         "tailwind-merge": "^3.6.0",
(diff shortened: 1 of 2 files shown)


dviejokfs:
Thanks for the first contribution @yash296  ! as soon as CI passes I'll merge it and create another release (beta.54)

PR #436, opened 2026-07-24
feat(sandbox): track agent-run sandboxes as first-class sandbox items

dviejokfs:
## Why

AI autofix (and workflow agent runs) launched containers through the raw `SandboxProvider` with no `sandboxes` row — invisible to the standalone sandbox API, no lifecycle events, no user attribution. The autofixer couldn't consume `SandboxService` directly because `temps-sandbox` depends on `temps-agents` (compile-time cycle).

## What

Dependency inversion via the plugin system:
- New `RunSandboxService` seam in `temps-agents::sandbox::managed`; the agents' `SandboxRegistry` uses it for create/release when present, raw provider otherwise (minimal installs without the sandbox plugin keep working).
- `temps-sandbox` implements it (`AgentRunSandboxBridge` over `SandboxService::create_for_agent_run`/`release_for_agent_run`) and injects it at plugin initialize. Rows get `agent_run_id`, lifecycle events (`created`/`destroyed`), and appear in the sandbox API.
- Migration: `sandboxes.user_id` nullable (webhook-triggered runs have no user), `sandboxes.agent_run_id` (indexed, partial), `agent_runs.triggered_by_user_id` — UI-triggered autofix runs attribute the sandbox to the clicking user so it shows in their sandbox list.
- Expiration sweeper + startup recovery skip agent-run rows (the run owns its lifecycle; autofix analysis→fix→PR legitimately outlives idle timeouts while the user reviews).
- `agent_run_id` surfaced in `SandboxSummary` / `SandboxInner` API responses.

## Verification
- `cargo check --workspace --all-targets`: clean
- `cargo test --lib -p temps-agents -p temps-sandbox`: 323 passed
- `sandbox_credential_smoke_test` fails identically on clean origin/main (pre-existing Docker-image perms assertion, unrelated)

## Follow-ups
- SDK regen for the new `agent_run_id` response field + optional UI badge for agent-run sandboxes
- e2e: trigger an autofix run and confirm the sandbox appears/disappears in the API

### DIFF
--- a/crates/temps-agents/src/handlers/autofixer.rs
+++ b/crates/temps-agents/src/handlers/autofixer.rs
@@ -365,6 +365,7 @@ pub async fn start_analysis(
             request.user_context,
             requested_provider,
             run_config,
+            Some(auth.user_id()),
         )
         .await
         .map_err(Problem::from)?;
@@ -947,6 +948,7 @@ mod tests {
             prompt_text: None,
             workspace_volume: None,
             run_config: None,
+            triggered_by_user_id: None,
         }
     }
 
--- a/crates/temps-agents/src/handlers/trigger.rs
+++ b/crates/temps-agents/src/handlers/trigger.rs
@@ -783,6 +783,7 @@ pub async fn smoke_test_agent(
         let image = format!("temps-sandbox-{}:latest", global_sandbox.runtime);
         let sandbox_config = crate::sandbox::SandboxCreateConfig {
             run_id: test_run_id,
+            owner_user_id: Some(auth.user_id()),
             container_name_override: None,
             host_work_dir: work_dir.clone(),
             workspace_volume: None,
--- a/crates/temps-agents/src/plugin.rs
+++ b/crates/temps-agents/src/plugin.rs
@@ -545,6 +545,10 @@ impl TempsPlugin for AgentsPlugin {
             context.register_service(sandbox_provider.clone());
 
             let sandbox_registry = Arc::new(SandboxRegistry::new(sandbox_provider));
+            // Registered so the temps-sandbox plugin can inject its managed
+            // run-sandbox service (agent runs then get first-class
+            // `sandboxes` rows in the standalone sandbox API).
+            context.register_service(sandbox_registry.clone());
 
             let config_service = Arc::new(AgentConfigService::new(
                 db.clone(),
--- a/crates/temps-agents/src/sandbox/docker.rs
+++ b/crates/temps-agents/src/sandbox/docker.rs
@@ -2758,6 +2758,7 @@ mod tests {
 
         // 1. Create sandbox
         let create_config = SandboxCreateConfig {
+            owner_user_id: None,
             run_id,
             container_name_override: None,
             host_work_dir: work_dir.clone(),
@@ -3104,6 +3105,7 @@ mod tests {
         let _ = std::fs::create_dir_all(&work_dir);
 
         let create_config = SandboxCreateConfig {
+            owner_user_id: None,
             run_id,
             container_name_override: None,
             host_work_dir: work_dir.clone(),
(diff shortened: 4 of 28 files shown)


dviejokfs:
Fixed both reviewed lifecycle-consistency bugs in commit 6392dbb4.

**Bug 1 — zombie rows on restart:** `SandboxService::release_orphaned_agent_run_sandboxes` (new) runs in the sandbox plugin's initialize phase — after `recover_stuck_runs` (register phase) has failed every in-flight run — and releases every non-destroyed `sandboxes` row whose agent run is terminal or gone, via the existing `release_for_agent_run` seam (container destroyed by run id, rows marked destroyed, per-run failures warn-logged and skipped). Single batched run-status lookup, no N+1.

**Bug 2 — user-destroy desync:** `destroy_sandbox` on a row with `agent_run_id` now rejects with typed `SandboxError::ManagedByAgentRun` (HTTP 409, message tells the user to stop/cancel the run) while the run is active, and routes through `release_for_agent_run` once the run is terminal so the real `temps-sandbox-<run_id>` container is torn down instead of being missed by public_id recovery. `pause`/`resume`/`restart`/`resize` reject agent-run rows with the same typed error. Run status sets are now shared consts (`ACTIVE_RUN_STATUSES`/`TERMINAL_RUN_STATUSES`) in `temps-agents`.

**Evidence**

```
$ cargo check --lib -p temps-sandbox -p temps-agents
cargo build: 0 errors, 2 warnings (2 crates)   # both warnings pre-existing workspace-level

$ cargo test --lib -p temps-sandbox -p temps-agents
     Running unittests src/lib.rs (target/debug/deps/temps_agents-8a5c04f5531754a6)
test result: ok. 237 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 6.61s
     Running unittests src/lib.rs (target/debug/deps/temps_sandbox-89d154062394bb68)
test result: ok. 107 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.34s

$ $(command -v cargo) clippy -p temps-sandbox -p temps-agents --all-targets
exit: 0   # no clippy warnings/errors in either crate
```

New tests: terminal/active status-set partition (`run_service`), `ensure_not_agent_run` reject/allow, `run_status_is_terminal` lifecycle matrix (`sandbox_service`), `ManagedByAgentRun` message contract (`error.rs`).

dviejokfs:
## Live e2e verification (real data, not fixtures)

Ran the full server locally against pre-existing production-like data and drove both fixes end-to-end through the real API/UI.

**Bug 1 — zombie rows on restart**
Found genuine pre-existing zombie data: `temps-sandbox-3` (agent run 3, run status `failed`) had sat with sandbox status `running` since the run died this morning. Restarted the server and observed:
```
Released sandboxes for 1 terminal agent run(s) on startup
Sandbox plugin: released 1 orphaned agent-run sandbox(es) on startup
```
DB confirms: `temps-sandbox-3` flipped `running` → `destroyed`.

Note: `temps-sandbox-4` (run 4, status `analyzed`) correctly stayed `running` — `analyzed` isn't in `TERMINAL_RUN_STATUSES`/`ACTIVE_RUN_STATUSES` since it's a legitimate waiting-for-review state, not an orphan. Working as intended, not a bug.

**Bug 2 — destroy desync**
Called `POST /api/v1/sandboxes/sbx_312f46faedb1c658/destroy` on the still-active run-4 sandbox:
```json
409 {"detail":"Sandbox sbx_312f46faedb1c658 belongs to agent run 4 and its lifecycle is managed by that run — stop or cancel agent run 4 instead","title":"Sandbox Managed By Agent Run"}
```
Row stayed `running` — no desync, no silent container leak.

**Bonus finding, fixed in 76ce51ac:** while chasing why the release wasn't logging anything, found `temps_sandbox` was missing entirely from the tracing filter allowlist in `temps-cli/src/lib.rs` — every log line from this plugin/service was silently dropped regardless of `--log-level`. Added the directive; confirmed the release log line above only appeared after the fix.

PR #440, opened 2026-07-25
feat(import): Kubernetes importer with cost + overprovisioning analysis

dviejokfs:
## What

Adds **Kubernetes** as an import source: point temps at a cluster via a pasted kubeconfig and get a reviewable migration plan that includes a **cost + overprovisioning analysis** — what the cluster costs today, how little of it is actually used, and what the same workloads would cost on a Hetzner server running temps.

## How it works

- **New crate `temps-import-kubernetes`** implementing `WorkloadImporter`:
  - **Discovery**: Deployments / StatefulSets / CronJobs across non-system namespaces (`<ns>/<kind>/<name>` workload ids)
  - **Describe**: image, command/args, ports, volumes (PVC/hostPath/emptyDir), probes, resources; env resolved through ConfigMaps and Secret **key names** — secret *values* are never read out of the cluster, they surface as `***REDACTED***` with a re-enter manual action in the plan
  - **Project snapshot**: the workload's namespace is the project boundary — database StatefulSets/Deployments (postgres/mysql/redis/mongo/minio images) become managed-service candidates with dump/restore data implications, Ingress hosts become domains, siblings become additional deployments
  - **Cluster observation**: nodes (instance type / region / providerID labels), summed pod requests, and live usage from `metrics.k8s.io` — every fetch degrades gracefully
- **Cost analysis** (`cost_analysis` on `ImportPlan`, new types in `temps-import-types`): node inventory priced from a static instance-type table (+$73/mo EKS/GKE control-plane fee when detected), requested-vs-capacity-vs-measured utilisation, an overprovisioning verdict (severe/moderate/reasonable), and the smallest Hetzner CPX server fitting 2× measured usage + temps overhead, with monthly/yearly savings. Unknown instance types are **not guessed**; anything unmeasurable becomes an explicit note.
- **Orchestrator**: `create_plan` now routes through `describe_project`/`generate_project_plan` when an importer supports project snapshots (Docker keeps the old path).
- **Auth**: token, embedded client certs, basic auth. `exec` plugins and file refs are rejected with actionable errors (e.g. "mint a ServiceAccount token"). The API server URL goes through the same `validate_external_url` SSRF guard as other importer `base_url`s. Kubeconfigs are per-request only (in-memory sessions), never persisted.

## Test evidence

- 21 new unit tests (kubeconfig resolution incl. rejection paths, quantity parsing, DB-image detection, full project-plan generation with cost analysis, validation rules, dry-run execute) — all pass
[...]

PR #441, opened 2026-07-25
feat(import): Kubernetes, Coolify, Dokploy, CapRover, Portainer, and Kamal importers with deploy-and-verify

dviejokfs:
Migrate full setups from other platforms into temps: **Kubernetes, Coolify, Dokploy, CapRover, Portainer, and Kamal** (plus the existing Docker importer), now unified into one PR targeting main (supersedes #440).

**Full import, not per-project:** each importer discovers workloads, snapshots the whole project (apps, databases, domains, env vars), and generates a reviewable plan. Execution is real: managed services are created (namespaced per project), data is copied from reachable source databases (readiness-gated dump/restore), custom domains registered, IP-tied source domains rewritten to the preview hostname, and env vars repointed at the new services.

**An import only reports success when the app is running:** the deploy-and-verify phase links public source repos to the project, follows the real pipeline to a terminal state, and HTTP-probes the deployed app. Verified end-to-end against a live Coolify instance (600-row database copied, app serving 200).

Also: onboarding migration tiles + wizard `?source=` deep links and a wider layout, migration docs with screenshots, and pipeline fixes surfaced by honest verification (empty commit SHA checkout, service port persistence, deployment-state semantics, duplicate initial-deploy race).

Models are typed from live API fixtures with unit tests per crate (124 tests) plus env-gated live e2e tests. Follow-up (same branch, pending): E2E-encrypted migration data transport (push agent + open-source relay) for unreachable source databases.


dviejokfs:
## Live e2e verification (this session, real Coolify instance, not mocked)

**Unit tests:**
```
$ cargo test --lib -p temps-import-coolify -p temps-import -p temps-import-dokploy -p temps-import-caprover -p temps-import-portainer -p temps-import-kamal -p temps-import-kubernetes -p temps-import-docker
cargo test: 124 passed, 4 ignored (8 suites, 0.01s)

$ cargo test --lib -p temps-deployments
test result: ok. 421 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 69.35s
```

**Full import execution against a real Coolify instance** (migration-lab seeded box, app `lab-shop` = heroku/node-js-getting-started + Postgres with 600 seeded order rows):
```
POST /api/imports/plan   -> 200
POST /api/imports/execute -> 202 in 195s

status: completed | deployment: 24
OK  create-service-lab-db  - Created temps-managed postgres service 'lab-shop-lab-db' (id 11) (4.9s)
OK  populate-service-lab-db - Copied data from the source platform into managed service 'lab-shop-lab-db' (4.6s)
OK  deploy-application     - Deployment finished (completed) (10.0s)
OK  verify-application     - http://lab-shop-production.localho.st:8080 responded with HTTP 200 (0.2s)
```

**Data actually copied** (not just a green step):
```
$ docker exec postgres-lab-shop-lab-db psql -U postgres -Atc "SELECT count(*) FROM orders;"
600
```

**App actually serving** (not just 'deployment record exists'):
```
$ curl -s -o /dev/null -w "%{http_code}" http://lab-shop-production.localho.st:8080/
200
$ curl -s http://lab-shop-production.localho.st:8080/ | head -c 120
<!DOCTYPE html><html><head><title>Node.js Getting Started on Heroku</title>...
```

This is the exact real-world failure mode the deploy-and-verify phase (Phase 4 of `execute_import`) exists to catch: earlier in this same session, before the fixes in this PR, imports created rows but the app was never actually running (empty commit SHA breaking checkout, wrong deployment-state semantics, a duplicate-trigger race, service ports never persisted, DSN encoding bugs). All of those were root-caused and fixed here, not worked around.

dviejokfs:
## Review findings fixed (962211c0)

All blocking and major findings from the review pass are fixed and pushed:

- **Blocking — session IDOR**: `GET /imports/{session_id}` now checks ownership like `execute_import` always did. Extracted into a shared `check_session_ownership()` with a real, dependency-free unit test (the existing orchestrator tests were all commented out as requiring full service wiring — this one needs none).
- **Major — poisoned-lock unwraps**: the 3 production `.unwrap()`s on the session `RwLock` now go through `sessions_read()`/`sessions_write()` helpers returning a typed error instead of wedging the whole import subsystem on a panic.
- **Major — error status collapse**: every `ImportError` variant now maps to its own HTTP status (401 auth, 429 rate-limited, 404 not-found, 502 source-unreachable, 500 execution) instead of a flat 400.
- **Major — no audit logging**: added `ImportExecutedAudit`, wired through `AppState`/plugin registration; `execute_import` now logs who ran what import against which platform.
- **Major — Dokploy `discover()` N+1**: unnamed database name lookups (no batch endpoint exists on Dokploy's API) now resolve concurrently via `futures_util::future::join_all` instead of one sequential await per stub.

Verified: `cargo check --lib` clean across the whole workspace, `cargo test` 126 passed (up from 124 — the 2 new ownership tests), `cargo clippy --all-targets -- -D warnings` clean via the resolved cargo binary.

dviejokfs:
## Security fixes + real e2e verification

Fixed all 3 findings from the latest security audit pass, commit 0ff85c2a:

### 1. Plaintext credentials in plan responses
`mask_plan_secrets()` now redacts `ServicePlan.parameters["source_url"]` and every `EnvironmentVariable` with `is_secret: true` before `POST /imports/plan` / `GET /imports/{session_id}` serialize the response. The session's stored plan stays unmasked (`execute()` needs the real values).

**While fixing this, found the masking would've been a no-op for 4 of 6 sources**: Dokploy, Coolify, CapRover, and Portainer all hardcoded `is_secret: false` for every env var regardless of name — meaning `DATABASE_PASSWORD`, `API_KEY`, etc. would still leak in cleartext even after masking landed. Added a shared `looks_like_secret_env_key()` heuristic in `temps-import-types` and wired it into all four (plus deduplicated `temps-import-docker`'s inline copy of the same heuristic).

**e2e proof**: ran the actual server, discovered a real local Docker container (`DATABASE_PASSWORD=SuperSecretPW999`, `API_KEY=sk-live-...`, `PUBLIC_APP_NAME=my-cool-app`) over the live HTTP API, and inspected the raw JSON:
```
DATABASE_PASSWORD = '***' | is_secret = True
API_KEY           = '***' | is_secret = True
PUBLIC_APP_NAME    = 'my-cool-app' | is_secret = False
```
Confirmed identically on both `POST /imports/plan` and `GET /imports/{session_id}`.

### 2. No timeout on the transfer container
`run_transfer_container` now bounds `wait_container` with `tokio::time::timeout` (30 min prod / 3s under `cfg(test)`). Extracted the wait+cleanup logic into a standalone `wait_for_container()` so it's testable without the full `ResourceExecutor` DI graph.

**e2e proof**: new test spins up a real Docker container running `sleep 300`, calls `wait_for_container` with a 2s bound, and confirms: times out (not silently hangs), force-removes the container, returns in well under the sleep duration, and the container is actually gone afterward (`inspect_container` fails). Ran against real Docker, not mocked.

### 3. Kubernetes insecure-skip-tls-verify only logged server-side
`KubeClient` now exposes `skip_tls_verify()`; it flows through `source_metadata` into `PlanMetadata.warnings` so it's visible in the plan response, not just a `warn!()` log.

**Verification**: spun up a real local k3d cluster and confirmed empirically that connecting without cert verification fails, and `danger_accept_invalid_certs` is genuinely required to reach it — proving the scenario the warning describes is real. The source_metadata → warnings plumbing itself is covered by two new deterministic unit tests. A live HTTP round trip for this one specifically isn't reachable from this environment: the importer's own SSRF guard (`validate_external_url`) correctly rejects loopback, RFC1918, and unspecified addresses — i.e. every network-reachable address a local test cluster can offer — which is the guard working as intended, not a gap.

All 130 tests pass across `temps-import`, `temps-import-types`, and every importer crate. Branch synced with `main` (merged PR #436's sandbox lifecycle fixes, no conflicts).

dviejokfs:
Fixed the first blocking finding from the last review-pr pass (commit 431eb6b0): **secret env vars were being silently dropped at deploy time** for Coolify, Dokploy, CapRover, and Portainer.

`execute_plan` filtered out every `is_secret` env var before building `CreateProjectRequest.environment_variables`. Before the masking fix (0ff85c2a), `is_secret` was hardcoded `false` for these four importers, so the filter was a no-op. Once `is_secret` started correctly identifying real secrets, that same filter began silently excluding them from the actual deployment — any imported app needing a real DB password or API key to boot would fail to start or fail its health probe, with no warning anywhere in the outcome.

Kamal was already correct and untouched: its `is_secret` vars are empty placeholders by design (values live in `.kamal/secrets`, never read), and it already warns the user to fill them in manually.

Fix: stop filtering. temps encrypts every environment variable at rest via `EncryptionService` regardless of `is_secret`, and the plan API response is already masked separately by `mask_plan_secrets()`. There's no reason to drop the real value from the deployment itself.

55 tests pass across the four fixed crates (none asserted the old dropping behavior).

Remaining findings from that review (not yet fixed): `DataImplication.recommended_action` still embeds the raw source DB URL unmasked in plan responses (same class of leak, different field), the generated frontend SDK's `ImportSource` type is stale (missing kubernetes/caprover/portainer/kamal), Portainer's client unconditionally disables TLS verification with no warning, and import sessions (holding full credentials) never get evicted from memory.

dviejokfs:
Fixed the remaining findings from the last review-pr pass:

**`recommended_action` plaintext leak (`c18407ea`)** — every platform importer's `dump_command()` embeds the source DB URL (with password) verbatim into a copy-pasteable `pg_dump`/`mysqldump`/`mongodump` command in `DataImplication.recommended_action`. `mask_plan_secrets()` masked `source_url` but never this sibling field carrying the identical credential. Fixed by scrubbing any verbatim occurrence of the raw URL out of `recommended_action` text too.

**Import sessions never evicted (`c18407ea`)** — sessions hold real, unmasked `ImportCredentials` (bearer tokens, admin passwords, or a full kubeconfig + private key for Kubernetes) with no eviction at all. Added a 1-hour `SESSION_TTL` pruned opportunistically on every `create_plan` call, plus eager eviction right after a real (non-dry-run) import completes successfully. Failed runs and dry runs are left alone so a transient failure can still be retried against the same plan.

**Portainer's unconditional TLS bypass (`19fd3b97`)** — `danger_accept_invalid_certs(true)` was hardcoded with no opt-out and no warning, unlike Kubernetes's `insecure-skip-tls-verify` which is an explicit user choice surfaced in the plan. Kept the default (Portainer ships self-signed by default; flipping it would break every default install's import), but added an explicit `credentials.extra["verify_tls"] = "true"` opt-in for trusted-cert instances, and the skip now surfaces as a `PlanMetadata` warning either way.

**Stale generated SDK (`fdedf82e`)** — `ImportSource` was missing `kubernetes`/`caprover`/`portainer`/`kamal` and `ImportSourceCapabilities` was missing `supports_cost_analysis`; `ImportWizard.tsx` was working around it with `as ImportSource` casts. Regenerated via `bun run openapi-ts` against a local build of this branch; `tsc --noEmit` shows no new errors.

133 tests pass across all import crates (3 new). Branch pushed, no conflicts.

That closes every finding from the last review-pr pass. Only remaining open item is the DNS-rebinding gap in `validate_external_url`, which is pre-existing and repo-wide (affects every importer, and every other caller of this shared guard) — recommend tracking as a separate follow-up issue rather than blocking this PR on it.

dviejokfs:
## Live e2e testing against the real migration-lab fleet (Coolify, Portainer, CapRover so far)

Spun up real Hetzner instances of Coolify, Dokploy, CapRover, Portainer, Kamal, and k8s and ran real discover → plan → execute cycles through this branch's server, exactly the path a real user's browser takes. Found and fixed two real bugs neither unit tests nor the earlier local-only e2e passes could have caught:

**Proxy 503 on real imports (`34330f5e`)** — `POST /api/imports/execute` synchronously builds, deploys, and health-checks the imported app, which routinely takes over the proxy's 60s hot-path upstream timeout for a real app. The request got RST'd and the browser would see a 503 **even though the import went on to complete successfully server-side seconds later** — a real user migrating a real app would be told it failed when it hadn't, with no way to know it actually worked. Fixed by giving console/control-plane traffic (not customer-app traffic) an extended 600s timeout. Verified live: a full Coolify import that previously 503'd now completes cleanly through the real proxy port.

**CapRover git-deployed apps always failed to deploy (`4a9188e7`)** — CapRover's own repoInfo tracks the real GitHub repo, and the importer already correctly parses it (`git_info_for`, unit-tested) — but that parsed info never made it into the plan's `deployment.git`, and `execute_plan` never read it anyway, so the deploy step always failed with "repo_owner is missing" for every git-deployed CapRover app, even ones with public repos. Fixed both halves; Coolify and Dokploy already had this wired correctly, and Portainer/Kamal genuinely have no git metadata to offer (checked — not the same bug). Verified live end-to-end twice: reproduced the exact failure against the unfixed binary, then confirmed the same real CapRover app deploys and serves HTTP 200 after the fix.

Both were caught specifically *because* this was tested through the real proxy port against real platforms rather than the console port / local Docker containers used in earlier verification rounds — recording that here since it's the reason this pass was worth doing.

Coolify and CapRover (once fixed) both now complete full real migrations end-to-end: services created, real data population via dump/restore, deployment, and HTTP 200 verification. Portainer's plan correctly shows the new TLS warning against its real self-signed cert; its deploy step fails on "repo_owner is missing" too, but that's expected/documented (Portainer manages raw containers with no git source — the plan's own manual_actions already say to trigger the first deploy manually). Continuing with Dokploy and Kubernetes next.

dviejokfs:
## \`temps imports run\` — migrate from Coolify/CapRover/Dokploy without the UI

Added a working CLI path for the migration this PR builds, on \`@temps-sdk/cli\` (\`apps/temps-cli\`), not the Rust binary — that one's just the server launcher (\`serve\`/\`migrate\`-schema/\`upgrade\`), the TypeScript CLI is what's meant for this kind of user/agent-facing interaction.

\`\`\`
bunx @temps-sdk/cli imports run --source coolify --token <token> --base-url https://coolify.example.com
\`\`\`

Walks through: pick a source → prompt for credentials (masked, never echoed) → discover workloads → pick one → render the full plan (risk badge, services + data implications, domains, env vars with secrets already masked, manual actions, unsupported features) → confirm → execute with live progress → final report with a dashboard link. \`discover\`/\`plan\`/\`execute\`/\`status\` remain as separate scriptable subcommands for automation (\`--yes\`, \`--json\`).

Two things worth noting:

1. **There's a second, unrelated "migrate" implementation already in this CLI** (\`commands/migrate/\` — Vercel/Coolify/Dokploy adapters that talk to those platforms directly from the CLI, bypassing this PR's backend pipeline entirely, no CapRover support). Deliberately built this on the existing (but previously broken) \`commands/imports/\` instead, since that's a thin wrapper over the real, deploy-and-verified pipeline already live-tested this session — extending the other wizard would have meant a third reimplementation of the same logic in TypeScript.
2. **The old \`imports\` command was non-functional against the real backend** — it never collected credentials (every self-hosted source would 401) and rendered fields (\`imported_count\`, \`progress\`) that don't exist on the real response types. Also regenerated the CLI's own SDK (\`apps/temps-cli/src/api/*.gen.ts\`), which had the same kubernetes/caprover/portainer/kamal staleness as the web frontend's did, plus ~5000 lines of unrelated accumulated drift — diffed \`typecheck\` before/after to confirm the regen only introduced one new (now-fixed) error, everything else was pre-existing and out of scope here.

Verified live against the real Hetzner lab fleet through this exact command — not mocked: full real migrations for Coolify, CapRover, and Dokploy, each ending in a real deployed app verified with HTTP 200.

82 CLI tests pass, typecheck baseline unchanged.

dviejokfs:
## CLI consolidation: one migration command, not two

The TS CLI (\`apps/temps-cli\`, \`bunx @temps-sdk/cli\`) had two parallel, redundant migration systems:

- **\`imports\`** — thin wrapper over the real Rust \`temps-import\` backend pipeline (services, data population, deploy, HTTP-verify).
- **\`migrate\`** — a client-side TS reimplementation with its own Coolify/Dokploy adapters, talking directly to the source platform's API instead of the backend.

Consolidated into one command:

- Renamed the \`imports\` command group to **\`migrate\`** (\`imports\`/\`import\` kept as aliases, so nothing breaks for existing scripts).
- Deleted the client-side Coolify and Dokploy adapters + their tests — the real backend importers (already live-tested against the Hetzner lab fleet earlier in this PR) are strictly better: real services, real data transfer, real deploy-and-verify.
- Vercel has no backend importer yet, so it's handled as a special case: \`migrate run\`'s source picker gets a synthetic Vercel entry, and picking it delegates to the existing client-side Vercel wizard instead of the (nonexistent) backend flow.

Verified: \`bun run typecheck\` (only pre-existing, unrelated baseline errors remain) and \`bun test\` (63/63 pass, down from 82 — expected, since the deleted coolify/dokploy adapter test files accounted for the difference). Smoke-tested \`migrate\`/\`imports\`/\`import\` all resolve to the same command tree, and \`migrate run --help\` reflects the updated source list.

dviejokfs:
## Live-tested `temps migrate` against all 6 real import sources

Ran the renamed `migrate` command end-to-end against the actual Hetzner lab fleet (not mocks) for every source. Found and fixed 2 real gaps along the way, plus a docs gap:

**Coolify** — `migrate run` → real service created, deployed, HTTP 200 verified.
**CapRover** — same, full deploy-and-verify green (also re-confirms the earlier git-source fix still holds).
**Portainer** — real service created, real data dump/restore, deploy step honestly fails as documented (image-only, no git source — see below).
**Kubernetes** — real discover/plan/execute against the lab cluster; services + domains created for real (one domain hit a "duplicate" error from a prior test run in the same dev DB — test artifact, not a bug).
**Kamal** — found and fixed a real bug (see below), then verified full plan/execute.
**Vercel** — not live-tested (no test credentials), but confirmed the CLI branches correctly to the client-side wizard without type/runtime errors.
**Dokploy** — not re-verified with a fresh token this round (couldn't mint one without risking secret exposure or touching an existing key); already fully verified end-to-end via this exact code path earlier in this PR's testing, and the rename touched zero Dokploy-specific logic.

### Bugs found and fixed
1. **Kamal credentials were never wired up in the CLI.** `imports/credentials.ts` treated Kamal as needing no input at all, but the backend importer actually requires the raw `config/deploy.yml` contents via `credentials.extra.deploy_yml` — so `migrate run -s kamal` failed no matter what flags were passed. Added a `--deploy-yml <path>` flag / interactive prompt.
2. **No `--username` flag existed anywhere**, even though Portainer requires one — forcing an interactive prompt even in scripted runs. Added `--username` (defaults to `admin`).
3. **Kamal, Portainer, and Docker imports never warned the user in the plan** that the first deployment needs a manual trigger (they're always image-only, no git source, so the automatic deploy step is guaranteed to fail with "repo_owner is missing"). Kubernetes already did this; added the same manual-action message to the other three so users find out during plan review instead of at the end of execute.

### Docs
`temps-landing`'s CLI reference (`docs/cli`) still documented the old, pre-consolidation two-command-group structure (`imports` with a stale flag list and no Vercel; `migrate` with only vercel/coolify/dokploy behind a `--from` flag that no longer exists). Opened kfsoftware/temps-landing-new#210 to fix it.

dviejokfs:
## Full review pass (Rust standards, security, frontend/SDK, tests) — findings fixed

Ran a 4-way parallel review (Rust standards, security audit, frontend/SDK, test coverage) against the full diff, then independently re-verified every blocking/major claim against the actual code before acting on it (two findings didn't hold up on closer inspection — see below). Fixed everything that did.

### Fixed
- **[blocking] SSRF via DNS rebinding** — Coolify/Dokploy/CapRover/Portainer/Kubernetes clients only ran `validate_external_url()` (rejects literal IPs/`localhost`, never resolves a domain). A user-controlled hostname could resolve to an internal address by connection time. Wired in the existing `resolve_and_validate_domain()` (already used by the webhook service for the same problem) and pinned each client to the validated address via `resolve_to_addrs()`.
- **[blocking] Portainer defaulted to insecure TLS** — `skip_tls_verify` was `true` unless the caller explicitly opted into `verify_tls="true"`, with no UI/CLI visibility. Flipped to verify-by-default; skipping now requires an explicit `verify_tls="false"` opt-out, and a cert-shaped connect failure now returns an actionable hint instead of a raw reqwest error.
- **[major] Import execute's worst case could still exceed the proxy timeout** — `populate_services` ran each service's data transfer sequentially (each bounded by a 30-min `TRANSFER_TIMEOUT`), so N services could exceed the 600s console timeout this same PR just extended. Made transfers run concurrently (worst case is now one `TRANSFER_TIMEOUT`, not N), and raised the console timeout to 3600s with a test asserting it actually covers the real worst case.
- **[major] Secret-masking heuristic missed connection-string-shaped keys** — `DATABASE_URL`/`REDIS_URL`/`SMTP_DSN` etc. passed through unmasked into plan responses since only `SECRET`/`PASSWORD`/`TOKEN`/`KEY` were matched. Added `_URL`/`_URI`/`DSN`/`CONNECTION_STRING` to the heuristic, plus its first unit tests.
- **[major] validation.rs had zero tests in 5 of 6 crates** — added direct unit tests for every rule (source-present, database-reachability, generated-domains, secret-placeholders, multi-host) in CapRover/Coolify/Dokploy/Kamal/Portainer.

### Investigated and downgraded (didn't hold up)
- **"IDOR on `repository_id`"** — the security pass flagged `repositories::Entity::find_by_id()` having no per-user ownership check. I checked whether the *regular* (non-import) project-creation flow enforces this anywhere else in the codebase. It doesn't: `get_repository_by_id` in `temps-git` (pre-existing, unrelated to this PR) has the same lack of a filter, and `get_user_connections_paginated` — despite its name — only filters `is_active`, no `user_id` at all. This is this system's existing, established trust model for git connections (any authenticated user with the right permission can use any connection on the instance), not a regression introduced here. Filed as a note for a separate, wider-scoped issue against the git-provider subsystem rather than blocking this PR.

All fixes verified: `cargo check --lib`/`cargo test --lib` clean across every touched crate (temps-import + all 6 importers + temps-import-types + temps-proxy, 178 tests total in the touched crates alone), and `$(command -v cargo) clippy --workspace --all-targets -- -D warnings` clean on the resolved binary.

dviejokfs:
Correction to my last comment: the verification line overstated scope — I ran `cargo check --lib --workspace` (whole workspace, clean) but clippy was scoped to the touched crates only (`$(command -v cargo) clippy --lib -p temps-import-caprover -p temps-import-coolify -p temps-import-dokploy -p temps-import-kamal -p temps-import-kubernetes -p temps-import-portainer -p temps-import -p temps-import-types -p temps-proxy -- -D warnings`, clean), not `--workspace --all-targets`. Test count (178) and all other verification claims stand as described.

PR #455, opened 2026-07-28
fix(cli): add per-command --context flag to avoid silent wrong-server runs

dviejokfs:
## Summary
- The CLI's server selection lived entirely in a single shared, mutable "active context" flag (`login` / `context use`), with no way to scope one command to a specific server without changing what every other command targets.
- That flag can silently drift between commands (a background `login` completing late, an earlier `context use`, another process on the same machine) — and because there's no per-command confirmation, a command then runs against whichever server happens to be active. **This is not hypothetical**: while manually verifying the metric-alerts-config-as-code feature (#454) locally, `projects create`/`deploy` silently executed against a real production account after the active context had reverted between two CLI invocations in the same session.
- Adds a global `--context <name>` flag that scopes exactly one invocation (sets `TEMPS_CONTEXT` for that command only) without touching the on-disk active flag. An unrecognized name **hard-exits before any command logic runs** rather than falling through to `TEMPS_CONTEXT`/the active flag/the legacy single-instance store.
- Updates the `temps-cli` skill to require `--context` on every state-changing command for scripted/agent usage, instead of relying on `login`'s side effect.

```bash
# BAD — depends on mutable global state that can drift:
bunx @temps-sdk/cli login http://localhost:3000
bunx @temps-sdk/cli projects create --repo owner/name ...   # may no longer be "local" by now

# GOOD — every command names its target explicitly:
bunx @temps-sdk/cli --context local-dev projects create --repo owner/name ...
```

## Test plan
- [x] `bun test` in `apps/temps-cli` — 66 pass (12 in `contexts.test.ts`, including 3 new `contextExists` cases)
- [x] `bunx tsc --noEmit` — no new errors (pre-existing unrelated errors in `env-sync`/`environments`/`notifications`/`providers`/`openapi-ts.config.ts` confirmed present on unmodified `origin/main` too)
- [x] Manual verification: `bun run src/index.ts --context localhost:13210 whoami` correctly targets that server; `--context nonexistent-typo whoami` exits 1 with the available-contexts list printed, and does not touch `~/.temps/.contexts.json`'s on-disk `isActive` flag (confirmed via direct inspection before/after)

### DIFF
--- a/apps/temps-cli/src/cli.ts
+++ b/apps/temps-cli/src/cli.ts
@@ -4,6 +4,7 @@ import { colors } from './ui/output.js'
 import { setQuietMode } from './ui/spinner.js'
 import { handleError } from './utils/errors.js'
 import { createRequire } from 'module'
+import { listContexts, contextExists } from './config/contexts.js'
 
 // Import command modules
 import { registerAuthCommands } from './commands/auth/index.js'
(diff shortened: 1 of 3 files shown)


dviejokfs:
Ran `/review-pr` on this and found a real, reproducible regression before merging, now fixed in 68e0cab4.

**The bug:** `login`/`logout` already define their own `--context <name>` (the name to save/remove — legitimately doesn't need to exist yet). The new global `--context` flag on the root command collided with that name. Confirmed directly (checked out the branch in a scratch worktree, drove `createProgram()` with an observer hook, no monkey-patching of the actual action):

- `login --context brand-new-name http://example.com` on this PR's original commit: the value never reached `login`'s own action at all (`actionCommand.opts()` came back `{}`) — Commander resolved the colliding flag name onto the **parent** command's option, not the more specific subcommand one. My first fix attempt (skip the new validation when the leaf command owns its own `--context`) only silenced the resulting crash — it didn't restore the value, so `login --context <name>` would have silently saved credentials under the *default* name instead of the one requested. That's worse than the crash: silent wrong behavior instead of a loud failure.
- Same repro against unmodified `origin/main` (pre-PR): the value correctly reaches `login`'s action.

**The fix:** renamed the new global flag to `--target-context <name>` so the two never collide. Re-verified all three cases directly against the real `createProgram()` (still no action-handler patching):
- `login --context brand-new-name <url>` → login's own action now receives `{ context: "brand-new-name" }` correctly, and proceeds into the real device-auth flow.
- `--target-context <typo> <command>` → still hard-exits with the available-contexts list, as intended.
- `--target-context <existing> <command>` → still sets `TEMPS_CONTEXT` for just that invocation and proceeds.

`bun test` (66/66) and `tsc --noEmit` clean on the branch after the fix.

One more thing I noticed but didn't touch: the PR description mentions updating the `temps-cli` skill to require `--context` on state-changing commands, but that file isn't in the diff — worth adding (with the new `--target-context` name) before merge, or as a fast follow.

PR #468, opened 2026-07-29
fix(projects): resolve nixpacks variants to base nixpacks preset

### LINKED ISSUE #466: Nixpacks project presets don't resolve to the base nixpacks preset
### Description

This is a slug mismatch between two preset registries.

**What happens**

1. UI list comes from temps_presets::all_presets(), which includes provider-specific Nixpacks variants such as nixpacks-node:
2. On save, project creation accepts the slug from get_preset_by_slug (so nixpacks-node passes that check), then parses it into the DB enum:
3. Preset::from_str only knows plain "nixpacks", not "nixpacks-node":
So the UI can show and select nixpacks-node, but persisting it fails with Unknown preset: nixpacks-node. Same for nixpacks-python, nixpacks-go, etc.

**Why it exists (assumption)**

The runtime/build side grew provider-specific Nixpacks presets (nixpacks-*), but the persisted temps_entities::preset::Preset enum was never updated to match, it only has a single Nixpacks variant.

Note: language presets like python, go, rust, java (and nodejs) are valid enum values and wrap Nixpacks under those slugs. nixpacks-node is a parallel, unpersisted identity.

### Steps to reproduce

1. Go to your temps console
2. Create a new project for a SvelteKit application (GitHub connection)
3. Select Nixpacks (Node.js)
4. Notice the error when you click "Deploy"

### Expected behavior

The project continues to the deployment step.

### Actual behavior

The project fails to create with Nixpacks (Node.js).

Image

### Operating System

Linux

### Temps version

v0.1.0-beta.55

### Docker version

Docker version 29.6.2

### Relevant logs

```shell
{
    "detail": "Invalid preset: Unknown preset: nixpacks-node",
    "title": "Invalid Input"
}
```

### Additional context

There are a couple of approaches we could take here:
1. Preselecting the Nixpacks base with a prefix search
2. Extending the provider list with the full list shown in the UI

I'm currently working on a PR using the former approach; let me know if you have a preference. @dviejokfs 

On another note. Absolutely amazing project! Huge respect for what you're building here; it's looking amazing so far!

brunyeestudio:
### Summary

- Fixes a mismatch between UI/API Nixpacks slugs (`nixpacks-node`, `nixpacks-python`, …) and the persistable `Preset` enum, which only has a single Nixpacks variant.
- Provider-specific slugs now resolve to `Preset::Nixpacks`, with the provider key stored on `NixpacksConfig.provider`, and round-trip back to `nixpacks-{provider}` on read/deploy/telemetry.
- Updates project create/update, deployments, git detection metadata, and plugins so they use the runtime slug instead of `preset.as_str()` / `Debug` formatting.

### The Problem

`temps-presets` exposes a closed set of provider-specific Nixpacks presets (`nixpacks-node`, `nixpacks-python`, `nixpacks-static`, etc.). The DB/entity layer only persists `Preset::Nixpacks`.

Before this change, project create/update effectively did `slug.parse::<Preset>()`. That rejects provider slugs (`nixpacks-node` is not an enum variant), so selecting a Nixpacks language preset in the UI/API could fail with “Invalid preset”.

Where code did get a stored Nixpacks row, consumers often looked up presets with:

- `project.preset.as_str()` -> always `"nixpacks"`
- or `format!("{:?}", preset).to_lowercase()` -> also `"nixpacks"`

So even after a successful save, deployments, workflow planning, telemetry, and plugins lost the provider-specific identity and treated the project as auto Nixpacks.

## Type of change

- [ ] Documentation update

## Checklist

## Related issues

Closes #466


dviejokfs:
@brunyeestudio Thank you for putting this together — the core storage decision here is the right one: keep the persisted preset canonical as Nixpacks and put Nixpacks-specific selection in its preset config, rather than adding nixpacks-language variants to the database enum.

Before merging, we want to generalize and tighten a few parts of the implementation:

- Model providers as a typed, ordered list so we can support both auto-detection and Nixpacks multi-provider builds, rather than storing one unrestricted string.
- Pass that provider configuration into Nixpacks plan generation. At the moment the selected provider round-trips through the API, but Dockerfile generation still supplies every provider, so nixpacks-node does not actually force or select Node.
- Move public-slug-to-storage resolution into the preset catalog/registration layer instead of adding Nixpacks-specific parsing to the entity enum. This will give every selectable catalog entry a canonical persisted preset and default config.
- Add a catalog invariant test so every preset exposed by the API/UI is persistable and deployable. This should also cover the existing react-app versus react mismatch.
- Validate provider configuration at deserialization/service boundaries and apply preset plus preset config atomically during updates.
- Add regression coverage for multiple providers, invalid providers, legacy Nixpacks config, and the create-to-deploy path.

Directory selection will stay independent: one directory/build context per Temps project, with multiple projects for multiple deployable directories in a monorepo.

We will contribute these follow-up changes directly to this PR branch since maintainer edits are enabled, while preserving the work and attribution already here. No action is required from you unless you have concerns with this direction. Thanks again for the detailed issue and implementation.

dviejokfs:
@brunyeestudio I’ve pushed the generalized follow-up directly to this branch in `3193b133`.

The main architectural change is that public catalog variants now own their mapping to canonical storage. We still persist `preset = nixpacks`; provider choice is typed config (`providers: [...]`), not a growing set of `nixpacks-*` database enum values. This keeps the mechanism reusable for future catalog variants without making `provider` a generic field on every preset.

The provider config now:

- supports ordered, multiple Nixpacks providers and the native `...` auto-detect marker;
- reaches the actual Nixpacks build-plan generator during deployment;
- uses native provider names where they differ from UI names (`c#`, `f#`, `staticfile`);
- remains independent from the project root directory, so monorepo directory selection still scopes each project/build normally;
- round-trips public aliases such as `nixpacks-node`, while multi-provider/hidden-provider configs return canonical `nixpacks`;
- validates inline Nixpacks TOML before persistence/build;
- is documented in the OpenAPI preset-config schema.

I also centralized create/update/Git-settings resolution, made preset + config updates atomic, fixed selecting base Nixpacks so it resets provider forcing, and added a catalog-wide persistence invariant.

### Evidence

- `cargo check --lib` for the affected crates — passed
- affected-crate Clippy with `-D warnings` — passed
- `cargo test --lib -p temps-entities -p temps-presets` — 255 passed
- `cargo test --lib -p temps-projects nixpacks -- --nocapture` against isolated PostgreSQL containers — 11 passed
- `bun test src/lib/preset-selection.test.ts` — 22 passed
- generator-level regression proves stored Python selection controls an ambiguous Node/Python repository build

The exhaustive matrix is left to GitHub Actions.

dviejokfs:
@brunyeestudio Local API-first E2E is now green on head 6bc2a0c9.

- Created project via POST /api/projects with preset=nixpacks, preset_config.providers=[node], directory=./examples/node-nixpacks, and branch=test/nixpacks-node-preset.
- The persisted response resolves to runtime preset nixpacks-node while retaining canonical config {preset: nixpacks, providers: [node]}.
- Triggered the pipeline through POST /api/projects/2/trigger-pipeline at temps-examples commit fbe3355e.
- Download, Nixpacks build, container deployment, routing, and all optional jobs completed successfully.
- The deployed root returned the expected Node.js response and /health returned {status: ok}.
- Added regenerated web client types for NixpacksPresetConfig.providers and NixpacksProvider in 6bc2a0c9.

Targeted frontend test: 22 passed, 0 failed.

brunyeestudio:
@dviejokfs Thank you for taking this to the next step and adding additional improvements! A typed provider over a growing nixpacks-* enum makes sense.

On another note, it may be worth opening a discussion... While looking into Nixpacks, I noticed that the project is in maintenance mode, and the [getting started page](https://nixpacks.com/docs/getting-started) recommends using Railpack as a replacement. Are there any plans to adopt this? I had encountered build errors on my first project deploy; for example, Nixpacks pins the `pnpm` version to 10.35, which throws `pnpm install` errors when the project expects to use `pnpm` 11 or higher.

- https://github.com/railwayapp/nixpacks/issues/1419
- https://github.com/railwayapp/nixpacks/issues/1419#issuecomment-4599247022

Install error logs:

```
●
[RUNNING] [ 7/11] RUN  pnpm i --frozen-lockfile
13:09:11.112
●
 ERROR  packages field missing or empty
13:09:11.614
●
For help, run: pnpm help install
13:09:11.617
●
[ERROR] [ 7/11] RUN  pnpm i --frozen-lockfile
13:09:11.641
●
ERROR: Build failed: Docker stream error: process "/bin/bash -ol pipefail -c pnpm i --frozen-lockfile" did not complete successfully: exit code: 1
```

dviejokfs:
@brunyeestudio 

The reason nixpacks was chosen is because it can be used as a rust library. Railpack is written in go so it would require a separate binary to be installed, breaking the promise of a single binary (right now the only hard dependency is docker)

The alternative for long term would be to either fork railpack and keep maintaining it or drop it completely and develop a "nixpack" like engine in temps.

dviejokfs:
Final maintainer review is complete on `f89a9c2f`.

The branch is now merged with current `main`, conflict-free, and GitHub reports it as mergeable.

### Final fixes

- Preserved the typed `custom` Dockerfile discriminator during config-only settings and Git updates when `variant` is omitted.
- Made explicit `dockerfile` and `custom` catalog selections authoritative in both transition directions.
- Validated inline Nixpacks TOML on create, full update, settings-only update, and Git-only update paths.
- Removed full project-model, generated Dockerfile, and build-argument log dumps.
- Redacted invalid inline TOML source content from returned errors.
- Regenerated the TypeScript SDK from the branch OpenAPI contract.

### Targeted evidence

- `cargo test --lib -p temps-presets test_custom_catalog_variant_roundtrips_through_dockerfile_storage` — 1 passed
- `cargo test --lib -p temps-presets test_every_public_catalog_preset_roundtrips_through_storage` — 1 passed
- `cargo test --lib -p temps-projects custom_dockerfile_variant -- --nocapture` — 2 passed
- `cargo test --lib -p temps-projects nixpacks_invalid_inline_toml -- --nocapture` — 3 passed
- `cargo test --lib -p temps-projects nixpacks -- --nocapture` — 12 passed
- `cargo check --lib -p temps-presets -p temps-projects -p temps-deployments` — 0 errors
- `bun test src/lib/preset-selection.test.ts` — 22 passed
- `bun run build` — passed

API-first verification created a fresh project using public slug `nixpacks-node`; the database stored canonical `preset=nixpacks` with `providers=[node]`. Deployment 38 completed from `./examples/node-nixpacks`, and its deployed URL returned HTTP 200. Browser verification reconstructed and selected Nixpacks (Node.js), with `POST /api/projects/3/git` returning 200.

Independent Rust, frontend/API-contract, and security reviews found no blocking or major issues. The security gate is approved. The exhaustive matrix is intentionally left to GitHub Actions.

PR #486, opened 2026-07-30
feat(teams): add teams and project-scoped RBAC

dviejokfs:
## What

Teams, project access grants, and per-project role enforcement — plus the thing that makes them mean something: an implementation of `temps_core::ProjectAccessChecker`.

The enforcement seam from ADR-028 has shipped inert. `project_access_guard!` and `project_permission_guard!` are threaded through ~330 call sites across 19 crates, but nothing has ever implemented the trait they resolve through, so every one of those guards has been a no-op. This PR is the implementation.

Custom roles are deliberately **not** here — they sit behind a plugin seam (see below).

## The safety property

A project with **zero** grants is unrestricted and behaves exactly as today. Team gating begins the moment its first grant is added, and nowhere else. Instance admins are never scoped by team membership.

This is why it merges without a migration window: on every existing instance `project_team_access` starts empty, so behaviour is unchanged until someone deliberately grants a team access.

## Contents

| Area | Change |
|---|---|
| Schema | `teams`, `team_members`, `project_team_access` — FKs into `projects`/`users`, cascade on delete |
| `temps-teams` | New crate: CRUD, `TeamProjectAccessChecker`, plugin registration |
| `temps-core` | Defaulted `hidden_project_ids` on `ProjectAccessChecker`; new optional `MembershipPermissionResolver` |
| `temps-projects` | `GET /projects` **and** `/projects/statistics` filtered by team access |
| `temps-analytics-session-replay` | Six handlers scoped to their owning project (security fix, see below) |
| Web | Teams list/detail under `/settings/teams`, per-project **Access** tab |
| CLI | `bunx @temps-sdk/cli teams` / `access` — 13/13 endpoints covered |

## Permission model

```
(member's role in the team) ∩ (role the team holds on the project)
```

unioned across teams. A registered `MembershipPermissionResolver` may replace the left side; the right side always applies, so a plugin can only ever narrow. Intersection cannot leak privilege regardless of which column an operator populated.

## Security review — findings and fixes

A `security-auditor` pass found a **real privilege escalation I had shipped**. Reproduced against a live instance:

```
alice: admin member of team T, T holds `viewer` on project 1
PUT /projects/1                          -> 403 projects:write
POST /projects/1/access {team:T, admin}  -> 201   <- self-promotion
PUT /projects/1                          -> 200
```

`grant_project_access` was guarded by instance-wide `ProjectsWrite` (which `Role::User` holds) plus the coarse `project_access_guard!` (which passes for a `viewer`). Any member could raise the grant half and escape their own ceiling — defeating the model's core property.

**Fixed** via `authorize_grant_mutation`: gating-state changes (first grant, last revoke) require an instance admin; otherwise the caller needs `projects:write` *within* the project; and nobody can grant a role carrying permissions they don't hold. Verified after the fix — `201 → 403`, ceiling holds, admins and project-admins still work, granting `owner` refused because it carries `projects:delete`, and a non-admin can no longer seize an ungated project.

Also found and fixed:

- **`add_events`** — sixth session-replay handler keyed by session id, still on instance-wide `AnalyticsWrite`. Guarded.
- **`GET /teams/{id}/projects`** — guarded by `ProjectsRead` (held by `Role::User`/`Reader`) with no membership check and sequential team ids, letting any low-privilege account enumerate the whole access matrix. Now `UsersRead`.
- **Soft-deleted users** — deletion sets `deleted_at` rather than removing the row, so the FK cascade never fires and memberships survive. All three resolution paths now exclude them.
- **`resolve_hidden_projects`** now fails closed on unresolvable identity; **`hidden_project_ids`** no longer loads every grant row per dashboard request; FK violations no longer misreported as "already a member"; TOCTOU and process-local invalidation limits stated honestly in the module docs.

One auditor finding was **not** acted on: it claimed `project_permission_guard!` has "exactly two production call sites". There are 20, across `temps-deployments` (11), `temps-projects` (6), and `temps-environments` (3). Its stated escalation payoff (project deletion) was also wrong — instance `projects:delete` gates it and `Role::User` lacks it. The real impact was the ceiling escape above.

## Evidence

Two teams, two projects, two non-admin users, on an isolated database.

**Isolation**

```
=== alice ===          === bob ===
GET /projects   ['alpha-site']    ['beta-site']
GET /projects/1  200               403
GET /projects/2  403               200
```

**Role narrowing** — `deployer` POST env-var → 201; `viewer` → 403 (`environments:create`). Owner-only: `deployer` DELETE project → 403.

**The ceiling, before the security fix landed** — same user, same team role, only the grant changed:

```
grant=deployer → PUT /projects/1  403
grant=admin    → PUT /projects/1  200
```

**Session-replay scoping** — a replay owned by project 2, alice holding `analytics:read` instance-wide:

```
                                          alice   bob
GET /visitors/1/session-replays            403    200
GET /visitors/1/session-replays/2          403    200
GET /visitors/1/session-replays/2/events   403    200
```

**Count filtering** — `admin: 2 · alice: 1 · bob: 1`, matching each caller's visible list.

**CLI** — exercised via `bunx --bun @temps-sdk/cli`: `teams list/show/update`, `teams members add/list/set-role`, `access grant/list`. Slug and email resolution confirmed.

**Web UI** — Teams list, team detail (back nav, edit dialog, role picker showing exactly the four fixed roles), project Access tab with the restricted/open banner, grant/revoke. Verified in-browser; screenshots in the PR thread.

## Checks

- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- 228 unit tests across `temps-teams` (61), `temps-projects` (77), `temps-entities` (54), `temps-analytics-session-replay` (15), `temps-auth` guard tests (21)
- `tsc --noEmit` — clean in `web/`; 0 new errors in `apps/temps-cli` (14 pre-existing, present before the SDK regen)
- Fresh database migrates cleanly with this branch's and main's migrations

## Known limitations (disclosed, not fixed here)

- `project_permission_guard!` covers ADR-028's v1 allowlist (20 sites). Outside it, a `viewer` can still take actions their instance role permits — e.g. `POST /projects/{id}/environments`. Widening the allowlist is a separate change.
- Four pre-existing API gaps found while reviewing, unrelated to this diff but adjacent to its claims: KV (7 endpoints, no authorization at all — `project_id` taken from the request body), `get_activity_graph` (no permission guard), `list_emails` (`project_id` is a filter, not a scope), and `analytics/general-stats` + global MRR (instance-wide aggregates). Tracked separately.

## Review

Security-sensitive. `security-auditor` has run and its findings are fixed; a human reviewer is still wanted given this activates an authorization boundary across ~330 call sites.


dviejokfs:
## Review round 2 — MAJOR + minor findings fixed

All three MAJOR findings shared one root cause: `authorize_grant_mutation` authorized against an **unlocked snapshot** and only inspected the state being *written*, never the state being *overwritten*. The rules moved into the service transaction, deciding against grant rows locked `FOR UPDATE`.

| # | Finding | Fix |
|---|---|---|
| MAJOR 1 | Role ceiling didn't cap the **overwritten** role — a project-admin could demote/delete an `owner` grant and become top authority | `check_role_ceiling` now runs on both the written and the replaced/removed role |
| MAJOR 2 | TOCTOU on concurrent revokes — two revokes each read "not the last grant", both authorized, together removed every grant and silently re-opened the project | Grants locked `FOR UPDATE` inside the txn; every rule evaluated under the lock |
| MAJOR 3 | Write gate used the checker's 60 s **cached** answer — a just-revoked project admin could re-grant themselves and make the revocation non-durable | New `effective_project_permissions_uncached`; the handler now only gathers inputs (`GrantAuthz`) |
| minor | `Ok(None)`-is-unreachable comment was false (reachable via revoke on an ungated project) | Comment rewritten; `None` explicitly maps to "holds nothing" |
| minor | `add_member` accepted **soft-deleted** users — FK satisfied (delete is soft), but the checker excludes them, so the membership granted nothing while showing as live | Pre-validation filters `deleted_at IS NULL` |
| minor | Unknown-team denial claimed the caller lacked `projects:write` — misdirection for a project-admin who holds it | Plain 404; the lookup stays last, so no enumeration oracle (test pins the ordering) |

**Not implementable as suggested:** the review asked that a `Role::User` be able to gate a project they created. `projects` has no `created_by`/owner column (`repo_owner` is the git namespace), so there is no "creator" to check. That needs a schema change and is out of scope here.

### Evidence

Unit: **57 tests pass** in `temps-teams` (11 new, one per rule). `cargo clippy -p temps-teams --all-targets -- -D warnings` clean; `cargo check --lib --workspace` clean.

Live, against a running instance (slot 8, `temps_dev_s8`). Fixture: project 1 carries an `admin` grant to team 1 and an `owner` grant to team 2; `carol` is an `admin` member of team 1, so she is a project-admin with no `projects:delete`.

```
===== DENIED (project-admin carol) =====
A. demote the owner team to viewer      403  Cannot apply the 'owner' role: it carries
                                             'projects:delete', which you do not hold
B. delete the owner grant               403  (same)
C. self-promote her own team to owner   403  (same)
D. probe a team that does not exist     404  Team 98765 not found
E. gate the un-gated project 2          403  Restricting a project to teams ... requires
                                             an instance administrator

===== ALLOWED (legitimate flows still work) =====
F. admin creates team 3                 201
G. carol grants team 3 deployer         201   (at/below her ceiling)
H. carol revokes team 3                 204
I. admin demotes the owner team         201   (no ceiling for instance admins)
J. admin restores it to owner           201
```

MAJOR 3 specifically — cache warmed by reads only, then the membership stripped **directly in Postgres** so no invalidation fires:

```
1. warm both caches with reads only
   GET /projects/1/access        -> 200
   GET /projects/1               -> 200
   GET /projects/1/environments  -> 200
2. delete from team_members where team_id=1 and user_id=3    DELETE 1
3. stale-cache window: POST /projects/1/access               403
   "Changing who can reach project 1 requires the projects:write permission on it"
4. is the cache genuinely still stale-allow?
   GET /projects/1/access        -> 200   <-- yes: cache says allowed, write gate says no
```

Step 4 returning 200 while step 3 returns 403 is the fix: the guard is still serving the stale allow, and the write gate no longer trusts it. Before this change step 3 returned 201.

No API schema change, so no SDK regeneration is needed — same endpoints, same request bodies, new denial codes only.

dviejokfs:
## Review round 3 — minors fixed

### The one with substance

Round 2 moved the grant/revoke *rules* under a `FOR UPDATE` row lock, but the *input* they were evaluated against stayed outside it: the handler resolved the caller's project permissions on a separate connection and passed them in via `GrantAuthz.held`. So the decision still rested on a snapshot taken before the lock, while the comments asserted it was made under the lock — the worse half, since the next reader would have trusted them.

Rather than soften the comments, the guarantee is now real:

- Permission resolution is extracted out of the checker into `permissions_from_grants`, generic over `C: ConnectionTrait`.
- The service calls it **on its own transaction**, passing the very grant rows it holds locked — so the resolution can't even re-read a different grant set.
- `GrantAuthz` shrinks to `{ is_instance_admin: bool }`. The handler can no longer supply permissions, by accident or otherwise; `grant_authz` is now a pure four-line classification.

The residual window from round 2 (sub-millisecond, but real) is gone: the read the decision rests on and the write it authorizes are the same snapshot.

### Other findings

| Finding | Fix |
|---|---|
| A corrupt stored `role` returned `InvalidRole` → **400**, blaming the caller for a data-integrity fault | New `CorruptStoredRole { entity, id, role }` → **500**, naming the table and row. Stored roles are never caller input — the API takes roles as a typed enum — so 400 sent operators hunting a fault in their own request. Fails closed either way. |
| Cache invalidation is post-commit and non-transactional, undocumented | Documented *why*: the cache must never be invalidated for a write that then rolls back, and a crash in that window can't be escalated because the write gate doesn't read the cache at all. |
| Comment pointed at `authorize_grant_mutation`, deleted two commits ago | Retargeted. |
| PR body claimed 45 tests in `temps-teams` | Updated to the real counts (above). |
| `handlers/project_access.rs` had zero tests | Added three, now that `grant_authz` is a pure function. |

### Evidence

**Unit** — 61 in `temps-teams` (was 57). The grant/revoke tests no longer inject a permission set; they append membership rows and drive the real resolution, so they additionally cover the grant role capping the member role, and a corrupt stored role denying with a 500. `clippy --all-targets -D warnings` and `cargo check --lib --workspace` clean.

**Live** (slot 8) — full matrix re-run against the refactored path, identical results:

```
DENIED (project-admin carol)                    ALLOWED
A. demote the owner team      403               F. admin creates a team        201
B. delete the owner grant     403               G. carol grants deployer       201
C. self-promote to owner      403               H. carol revokes it            204
D. probe a nonexistent team   404               I. admin demotes owner team    201
E. gate an un-gated project   403               J. admin restores it           201
```

**The lock/cache property** — cache warmed by reads only, then the membership stripped **directly in Postgres** so no invalidation fires:

```
1. warm both caches with reads only    GET x3 -> 200
2. delete from team_members ...        DELETE 1
3. POST /projects/1/access             403   "requires the projects:write permission"
4. GET /projects/1/access              200   <-- cache still stale-allow
```

Step 4 at 200 while step 3 is 403 is the point: the read guard is still serving the stale allow, and the write gate — now reading inside the transaction, under the lock — is not fooled by it.

One correction on process: my first attempt at this test ran against the wrong Postgres (there are several daemons on this machine; `docker` resolved to a context whose `temps-dev-db` doesn't hold `temps_dev_s8`), so the membership was never actually deleted and the step-3 `201` it produced was meaningless. The run above is against the database the server is genuinely serving, confirmed by reading back rows created through the API.

No API schema change, so no SDK regeneration.

PR #487, opened 2026-07-30
fix(domains): reserve the console hostname from project domains

### LINKED ISSUE #478: Project domain can override Temps console
### Description

When configuring a project with a domain, you can use the Temps console domain. After selecting it, the Temps console becomes inaccessible, and you must use the public IP to restore the console.

Image

### Steps to reproduce

1. Configure Temps to run on a custom domain.
2. Deploy a sample project
3. Go to the domain section and add the console domain for the project.
4. Refresh your page, and you will be presented with either your application or an error.

### Expected behavior

The Temps console domain can not be selected for a project.

### Actual behavior

The project can use the Temps console domain and cause access to the Temps console to become inaccessible.

### Operating System

Linux

### Temps version

v0.1.0-beta.55 (ba60d25)

### Docker version

29.6.2, build dfc4efb

### Relevant logs

```shell

```

### Additional context

My immediate thought is to simply filter out the domains configured for the Temps console, the external URL, and the preview domain. However, that may come with some oversight. Any thoughts @dviejokfs ?

Image

dviejokfs:
Closes #478

## Problem

A project custom domain could be set to the Temps console hostname. The proxy then routed the console host at the project's deployment, so the console became unreachable and the only way back in was the server's public IP.

## Fix

Three layers, because the API guard alone leaves already-broken installs broken:

1. **`AppSettings::is_reserved_hostname`** (`temps-core`) — the console hostname (derived from `external_url`, accepting a full URL or a bare host) and the preview-domain apex are platform-owned. Case- and trailing-dot-insensitive. Inert when `external_url` is unset (IP-only installs).
2. **`CustomDomainService`** — `create_custom_domain` and `update_custom_domain` (rename path) reject reserved hostnames with `InvalidDomain`, which already maps to a 400 Problem Details response. This also covers the importers, which go through the same service.
3. **Route table** — any route resolving to the console hostname is dropped just before the atomic swap, with a `warn!`. Installs that already stored such a row self-heal on the next reload instead of requiring public-IP access to delete it.
4. **Web UI** — `DomainSelector` renders reserved entries with a "Reserved" badge and `disabled`; `DomainForm` blocks submit and explains why, which also covers the wildcard-plus-subdomain path where the reserved host is typed rather than picked.

Subdomains of the preview domain are deliberately still assignable — only the apex is reserved.

No new API endpoint, so no `@temps-sdk/cli` change is needed; the CLI surfaces the 400 from the existing endpoint.

## Evidence

New tests, all run locally:

```
$ cargo test -p temps-core --lib app_settings
cargo test: 19 passed, 253 filtered out

$ DOCKER_HOST=unix://.../colima/default/docker.sock cargo test -p temps-projects --lib custom_domain
cargo test: 26 passed, 36 filtered out
   (incl. test_console_domain_is_rejected: create rejects console/CONSOLE/preview apex,
    a normal domain still creates, and renaming onto the console host is refused)

$ DOCKER_HOST=unix://.../colima/default/docker.sock cargo test -p temps-routes --lib test_route_table
cargo test: 15 passed, 43 filtered out
   (incl. test_route_table_skips_console_hostname: a pre-existing project_custom_domains row
    for console.example.com produces no route, while app.example.com still routes)

$ cd web && bun test src/lib/reservedHostnames.test.ts
8 pass, 0 fail
```

Lint/type gates:

```
$ cargo clippy -p temps-core -p temps-projects -p temps-routes --all-targets -- -D warnings   # clean
$ cd web && bunx tsc --noEmit                                                                  # clean
```

`eslint` on `DomainForm.tsx` reports 2 `react-hooks/preserve-manual-memoization` errors — both pre-existing on `origin/main` (verified by stashing), untouched by this change.

### DIFF
--- a/crates/temps-core/src/app_settings.rs
+++ b/crates/temps-core/src/app_settings.rs
@@ -1021,11 +1021,112 @@ impl AppSettings {
             .unwrap_or_else(|| format!("http://host.docker.internal:{console_port}"));
         raw.trim_end_matches('/').to_string()
     }
+
+    /// Hostname the Temps console is served on, derived from `external_url`.
+    ///
+    /// Returns `None` when `external_url` is unset or unparsable (installs
+    /// reached by raw IP), in which case there is no console hostname to
+    /// protect.
+    pub fn console_hostname(&self) -> Option<String> {
+        let raw = self.external_url.as_ref()?.trim();
+        if raw.is_empty() {
+            return None;
+        }
+        // Tolerate a bare host ("console.example.com") as well as a full URL.
+        let candidate = if raw.contains("://") {
+            raw.to_string()
+        } else {
+            format!("https://{raw}")
+        };
+        url::Url::parse(&candidate)
+            .ok()?
+            .host_str()
+            .map(|h| h.trim_end_matches('.').to_ascii_lowercase())
+    }
+
+    /// True when `host` is owned by the platform itself and must never be
+    /// claimed by a project domain.
+    ///
+    /// Reserved hosts are the console hostname (`external_url`) and the
+    /// preview domain apex — routing either of them at a project makes the
+    /// console or every generated preview URL unreachable, and recovering
+    /// requires shell/IP access to the box (issue #478).
+    pub fn is_reserved_hostname(&self, host: &str) -> bool {
+        let host = host.trim().trim_end_matches('.').to_ascii_lowercase();
+        if host.is_empty() {
+            return false;
+        }
+        if self.console_hostname().as_deref() == Some(host.as_str()) {
+            return true;
+        }
+        let preview = self
+            .preview_domain
+            .trim()
+            .trim_end_matches('.')
+            .to_ascii_lowercase();
+        !preview.is_empty() && preview == host
+    }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    // Issue #478: a project domain must never be allowed to claim the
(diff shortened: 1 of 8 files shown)


crates/temps-core/src/app_settings.rs
```diff
@@ -1021,12 +1021,113 @@ impl AppSettings {
...
         raw.trim_end_matches('/').to_string()
     }
+
+    /// Hostname the Temps console is served on, derived from `external_url`.
+    ///
+    /// Returns `None` when `external_url` is unset or unparsable (installs
+    /// reached by raw IP), in which case there is no console hostname to
+    /// protect.
+    pub fn console_hostname(&self) -> Option<String> {
+        let raw = self.external_url.as_ref()?.trim();
```
brunyeestudio, line 1031: I assume this does not protect against cases where the operator has incorrectly set their external URL?

Image

I can envision a case where the operator configures/misspells their domain for the external URL and then is allowed to assign the console domain to a project. The UI and checks here do not account for that.

Should the external URL be configurable for custom domain setups, or can we infer it during setup and lock it?

dviejokfs: Good call.

Theres also a feature to set allowed host headers for the console, maybe that must be the main console url and extract the host from the url as the allowed one for the console

dviejokfs: it will be handled in a different PR

PR #488, opened 2026-07-30
fix(core): make Visit links reachable on sslip.io installs

jlucaso1:
## Problem

On a quick-start install, clicking **Visit** on a project opens a URL that times out:

```
http://observability-starter-1.127.0.0.1.sslip.io/   ERR_CONNECTION_TIMED_OUT
```

The app itself is healthy — the environment host serves it in ~2 ms. Two independent defects stack up to produce the dead link.

### 1. The sslip.io domain is generated with a dotted IP

Wildcard-DNS services (sslip.io, nip.io, traefik.me) locate the target IP by scanning the hostname for four numeric components joined by a **single** separator style — all dots or all dashes — and taking the **leftmost** match. That match is not anchored to the base domain, so with a dotted base any generated label ending in a number donates it to the front of the IP and pushes the real last octet out.

Measured against live DNS:

| hostname | resolves to | |
|---|---|---|
| `observability-starter-1.127.0.0.1.sslip.io` | `1.127.0.0` | ❌ not local |
| `pr-42.127.0.0.1.sslip.io` | `42.127.0.0` | ❌ not local |
| `pr-42.10.0.0.5.nip.io` | `42.10.0.0` | ❌ not local |
| `pr-42.127.0.0.1.traefik.me` | `42.127.0.0` | ❌ not local |

Deployment slugs are `{project}-{n}` and preview environments are `pr-{number}`, so on a dotted base this was **every generated hostname**, not an edge case.

**Fix: emit the dashed spelling at install time.** The separator styles then disagree — `1.127-0-0` mixes a dot and dashes and is rejected — so the scan falls through to the genuine all-dash quad:

```
observability-starter-1.127-0-0-1.sslip.io  ->  127.0.0.1   ✅
pr-42.127-0-0-1.sslip.io                    ->  127.0.0.1   ✅
foo-1-2.127-0-0-1.sslip.io                  ->  127.0.0.1   ✅
```

`deploy.sh` sets the same dashed form for the quick-start flow and passes it via `--wildcard-domain`. This PR covers `temps setup` run **without** `--wildcard-domain`, which generates its own domain — so the two installers agree.

> **Note on approach.** An earlier revision of this PR rewrote the base domain at *runtime* inside `normalize_base_domain`. That has been reverted. Mutating a stored value on read is confusing to operators, and it split generators from matchers: the on-demand TLS zone, the sandbox preview-host parser, the preview cookie scope and the ephemeral redirect target all read the raw setting, so each had to be kept in sync by hand — a class of bug that recurred twice during review. Emitting the correct value once, at the source, removes it entirely.

### 2. The project header linked to the ephemeral deployment host

`deployment.url` is derived from the deployment slug, so it is specific to one build and changes on every deploy. The environment's stable domain was already on the same payload, and `DeploymentDetails` already preferred it — `ProjectDetail` did not, and passed `lastDeployment.url` straight to the Visit button.

This half is a bug on **any** install, not just sslip.io: Visit should land on the domain that keeps working after the next deploy.

## Changes

- `crates/temps-cli/src/commands/setup.rs` — the generated sslip.io domain uses the dashed IP form. One line plus the rationale; no runtime logic.
- `web/src/lib/deployment-url.ts` — `resolvePrimaryUrl` extracted from `DeploymentDetails` and shared with `ProjectDetail`. Scheme validated by parsing rather than a `startsWith('http')` prefix test, which passed `httpfoo://` through and turned `//evil.com` into an `evil.com` origin; candidates now fall through instead of yielding null, so one bad stored domain can't suppress a valid link.
- `web/src/pages/{ProjectDetail,DeploymentDetails}.tsx` — use the shared resolver.

`crates/temps-core/src/public_hostname.rs` is **unchanged** — hostname generation is untouched.

Superseded deployments still link to their own URL, since they are not served at the environment domain.

## Not covered here

**Existing installs** provisioned with a dotted domain keep their broken hostnames; nothing rewrites them. The intended follow-up is a settings warning that detects the dotted spelling, explains why app URLs time out, and offers a one-click change — telling the operator rather than silently rewriting.

## Verification

- `cargo test --lib -p temps-core public_hostname` — 9 passed (unchanged from `main`; confirms hostname generation is untouched).
- `cargo test --lib -p temps-cli setup` — 5 passed.
- `bun test src/lib/deployment-url.test.ts` — **14 passed**, including rejection of `javascript:`, `httpfoo://`, `//evil.com`, `data:` and `file:` candidates, fall-through past a hostile domain, and http not being upgraded to https.
- `cargo fmt --all --check`, `cargo clippy -p temps-cli --all-targets -- -D warnings`, `bunx tsc --noEmit`, `prettier --check` — all clean.
- Live DNS, on the exact hostnames the installer now produces:

```
$ dig +short observability-starter-1.127-0-0-1.sslip.io
127.0.0.1
$ dig +short myapp-prod.1-2-3-4.sslip.io
1.2.3.4
```

## Known limit

A name that itself contains a dash-joined IPv4 quad (`a-10-0-0-5.<base>`) still resolves to that embedded address. That takes a deliberately hostile name and is out of scope.


dviejokfs:
really good catch thanks for the contribution @jlucaso1!

Let me take a look, the workaround is to use the environment link instead of the deployment link. That's what I usually use since deployment specific links work for static websites, but once the deployment is superseeded the deployment URL no longer works

dviejokfs:
## Switched the fix from suffixing labels to normalizing the base domain

Pushed as `07cf4e5`. The diagnosis in this PR is exactly right, and I reproduced every row of your table against live DNS. I've changed **where** the fix is applied, not what it fixes.

Previously: every generated label that would end in an IPv4 octet got an `x` appended (`pr-42` → `pr-42x`).
Now: a wildcard-DNS base written as a dotted IPv4 is rewritten to its dashed spelling (`127.0.0.1.sslip.io` → `127-0-0-1.sslip.io`), and the generated name is left completely alone.

```text
before:  pr-42x.127.0.0.1.sslip.io     ->  127.0.0.1   (works, name mangled)
after:   pr-42.127-0-0-1.sslip.io      ->  127.0.0.1   (works, name intact)
```

### Why this is better

**1. It doesn't change hostnames on installs that never had the bug.**

The clearest symptom of the old blast radius is in the diff itself: an existing assertion had to change from `preview-123.example.com` to `preview-123x.example.com`. A plain `example.com` install has no defect, yet every one of its preview hostnames shifted — forcing certificate reissue and breaking existing links, for zero benefit. The stated rationale ("a hostname is identical across installs regardless of base domain") doesn't really hold, since hostnames already differ across installs precisely *because* the base domain differs.

This also worked against ADR-029's stable-endpoint goal: webhook and OAuth redirect URIs are pinned to preview URLs, so silently moving `pr-42` → `pr-42x` is the opposite of what that needs.

**2. The old guard was not injective — a real collision.**

```rust
"pr-42"  -> trailing "42"  -> octet     -> "pr-42x"
"pr-42x" -> trailing "42x" -> not octet -> "pr-42x"
```

Two distinct environments collapse onto one hostname and shadow each other in the route table. The `guard_is_idempotent` test enshrined exactly the property that creates this. Any append-a-marker scheme has this flaw; normalizing the base has nothing to collide.

**3. It repairs already-provisioned installs, with no migration.**

Normalization happens in `normalize_base_domain`, which every hostname helper already funnels through, so it applies on read. An install created before this change starts producing working hostnames as soon as it's upgraded — no operator action, no DB migration. Changing only the installer default would have left every existing install broken.

**4. It covers nip.io and traefik.me too.**

Both have the identical defect, which the label guard also fixed, but worth recording that all three were measured:

```text
pr-42.10.0.0.5.nip.io        ->  42.10.0.0    ✗       pr-42.10-0-0-5.nip.io       ->  10.0.0.5   ✓
pr-42.127.0.0.1.traefik.me   ->  42.127.0.0   ✗       pr-42.127-0-0-1.traefik.me  ->  127.0.0.1  ✓
```

`localtest.me` is deliberately excluded — it resolves everything to loopback without parsing an IP out of the hostname, so it has no defect to fix.

### One thing this required that's worth flagging

On-demand TLS derives its zone from `external_url`, and `temps-proxy/src/on_demand_cert.rs:380` matches SNI against the `.{zone}` suffix. Normalizing hostnames without normalizing the zone would have meant every dashed app hostname failed the in-zone check and **no certificate would ever be issued** on a sslip.io install. `derive_zone` now applies the same normalization, and there's a test asserting both spellings of `external_url` derive the same zone.

Pleasingly, the codebase was already half-expecting this: `on_demand_cert.rs` and `setup.rs::is_loopback_zone` both already matched `127-0-0-1` alongside `127.0.0.1`.

### What I kept

Your second fix — `resolvePrimaryUrl`, so Visit links to the stable environment domain rather than the ephemeral `{project}-{n}` deployment host — is untouched. It's a genuine, install-agnostic bug and the fix is correct. Same for the shared extraction into `web/src/lib/deployment-url.ts`.

### Verification

- `cargo test --lib -p temps-core public_hostname` — 16 pass, including per-suffix coverage, idempotency, either-spelling equivalence, and that ordinary domains are byte-for-byte unchanged.
- `cargo test --lib -p temps-cli on_demand` / `setup` — 18 + 5 pass; three zone assertions updated to the dashed spelling, plus a new test that both spellings agree.
- `cargo check --workspace --lib`, `cargo fmt --all --check`, `cargo clippy -p temps-core -p temps-cli --all-targets -- -D warnings` — all clean.
- Live DNS confirms the exact hostname the code now emits: `observability-starter-1.127-0-0-1.sslip.io` → `127.0.0.1`.

The remaining known limit, unchanged from before and documented on the function: a name that itself contains a dash-joined quad (`a-10-0-0-5.<base>`) still resolves to that embedded address. That takes a deliberately hostile name and is out of scope.

jlucaso1:
Thank you <3

dviejokfs:
@jlucaso1 thinking about this, I think this PR is not needed

I need to modify the deploy.sh script to change the ip notation for sslip.io from "." to "-".

Modifying values at runtime based on a logic can confuse the user.

what do you think?

dviejokfs:
Agreed — reworked the PR along those lines rather than closing it, since the second half is unrelated and still needed.

**The runtime rewriting is gone.** `crates/temps-core/src/public_hostname.rs` is back to `main` untouched, and so are the TLS-zone, proxy and DNS-resolver changes that existed only to keep matchers in step with it.

You were right about the confusion risk, and the review made the case concretely: normalizing on read split *generators* from *matchers*, and the split had to be maintained by hand in six places. It was wrong twice in one day — first the configured on-demand TLS zone (which would have silently stopped certificate issuance on every existing install), then the sandbox preview parser, the preview cookie scope and the ephemeral redirect. Fixing the value at the source removes the whole class.

**One thing to keep alongside `deploy.sh`:** `crates/temps-cli/src/commands/setup.rs:1319` generates its *own* sslip.io domain when `temps setup` runs without `--wildcard-domain`. `deploy.sh` always passes one, so it isn't your main flow, but leaving it dotted means the two installers disagree. That one-line change is what this PR now contains — same fix, no runtime logic:

```rust
let sslip_domain = format!("{}.sslip.io", ip.replace('.', "-"));
```

**Kept:** the `resolvePrimaryUrl` fix (Visit linking to the stable environment domain instead of the ephemeral `{project}-{n}` build host). That's a real bug on every install including plain custom domains, unrelated to sslip.io, and it would have been lost by closing.

The PR is now **5 files, +242/-14** (was 15 files, +660/-36).

**Existing installs** are deliberately not addressed here. Your warning-plus-one-click idea is the right shape for that and matches the "show status, give instructions" rule — detect the dotted spelling, explain why app URLs time out, offer to change it. Happy to pick that up as a follow-up.

Verification and the DNS measurements are in the updated description.

jlucaso1:
> @jlucaso1 thinking about this, I think this PR is not needed
> 
> I need to modify the deploy.sh script to change the ip notation for sslip.io from "." to "-".
> 
> Modifying values at runtime based on a logic can confuse the user.
> 
> what do you think?

This can work, so it's fine

PR #489, opened 2026-07-30
feat(deployments): support worker nodes of a different architecture

jlucaso1:
## Problem

A worker node of a different CPU architecture joins the cluster without complaint and then receives deployments it cannot run.

The control plane builds an amd64 image, `docker save`s it, pushes the tar to the arm64 agent via `POST /agent/images/import`, and the container dies with `exec format error`. Nothing in the deploy log mentions architecture — the failure surfaces as a container that won't start.

The check that should have caught this is inert for remote nodes:

```rust
// crates/temps-deployer/src/remote.rs
fn get_native_platform(&self) -> String {
    // Unknown for remote — will need to query agent in Phase 2
    "linux/amd64".to_string()
}
```

`RemoteNodeDeployer::inspect_image` returns `"not supported"`, so `validate_image_platform` can never compare anything. The `nodes` table has no architecture column, the heartbeat reports only CPU/memory/disk, and `NodeScheduler` places replicas without knowing what a node can execute.

Managed services are unaffected — the agent pulls those images on the node itself (`service_handlers.rs`), so multi-arch manifests resolve natively. The gap is specific to images the control plane builds.

### Architecture discovery

The agent resolves its platform from `docker info` — the **daemon's** architecture, not the binary's, since an agent can drive a daemon over `DOCKER_HOST` (or an emulated one) that differs from the process it runs in. Reported at registration and on every heartbeat, stored on a new nullable `nodes.architecture`, surfaced in the console.

Nullable on purpose: an agent older than this change reports nothing, and such a node stays schedulable (logged as unverified) so a rolling agent upgrade doesn't drain the cluster.

`crates/temps-deployer/src/platform.rs` is new: one place that normalizes the three spellings the ecosystem uses (`x86_64` from `docker info`, `amd64` from image configs, `std::env::consts::ARCH`), replacing the `#[cfg(target_arch)]` blocks previously duplicated in `docker.rs`, `build.rs` and `remote.rs`. Unknown architectures pass through verbatim rather than being coerced to amd64 — they must fail a comparison, not fake a match.

### Scheduling

`NodeScheduler` gained a platform filter. A node enters the pool only when an image exists for its architecture, and the control plane's `Local` slot is filtered the same way — on a heterogeneous cluster it is just another node with an architecture. When nothing compatible remains the deployment fails with `NodeError::NoCompatibleNode` instead of silently falling back to `Local`, which was the old behaviour and equally broken.

### Cross-architecture builds

When the nodes a deployment could land on span several architectures, `BuildImageJob` builds one image per platform. The control plane's platform keeps the plain tag; the rest are suffixed (`myapp:latest-arm64`), so several architectures coexist in one image store without a registry or manifest list. Each replica receives the tag built for the node it lands on.

The eligible-node set is computed with the same selectors the scheduler will apply (`target_nodes` / `target_labels`), so we never pay for an emulated build targeting a node the deployment can't reach. A homogeneous cluster gets an empty platform list — one native build, same tag, same scheduling decisions as before.

Cross-builds go through the daemon's `platform` option, which needs QEMU binfmt handlers on the control plane. When they're missing the error names the platform and the command that installs them.

### Fail-fast on transfer

Before an image is shipped, its architecture is compared against the target node's. A mismatch fails the deployment naming both platforms, instead of transferring a tar that produces a container which cannot start. A node that hasn't reported a platform is asked directly via its health endpoint before falling back to "unknown, proceed".

### Legacy-builder guard

Docker's legacy builder **accepts `platform` and ignores it** — it reports success and returns an image of the host's architecture:

```
DOCKER_BUILDKIT=0 docker build --platform linux/arm64 …  →  linux/amd64
DOCKER_BUILDKIT=1 docker build --platform linux/arm64 …  →  linux/arm64
```

Left unchecked, a control plane without BuildKit would tag an amd64 image as `-arm64`, ship it, and fail on the node — with an error blaming the node rather than the build. Every cross-built image is now inspected and a mislabelled result fails the build, pointing at BuildKit as the fix.

### Image uploads

`deploy/image-upload` validated the uploaded image against the control plane's platform only, so a legitimate arm64 image was rejected with 400 even when an arm64 worker existed. It now validates against every architecture present in the cluster.

### Drive-by fix

`DockerRuntime::remove_image` bound its request future to `_stream` and dropped it: nothing was sent to the daemon and every caller received a silent `Ok(())`. Found because the new tests could not clean up the images they create.

## Testing

**Unit and integration** (~50 new tests, all in the `docker-deployments` and `migrations` CI groups):
platform normalization; the scheduler's architecture filter and `required_build_platforms`; per-platform tag selection; sanitization of the agent-reported value at the HTTP boundary; `RemoteNodeDeployer` platform resolution against a fake agent; multi-platform build with a mock that records what was requested; heterogeneous-fleet placement end to end with `MockDatabase`.

**Real Docker** (`tests/multiarch_build_test.rs`, new): drives `BuildImageJob` against a live daemon for two platforms and asserts each tag carries the architecture it claims and that the two images differ. No QEMU needed — the fixture Dockerfile has no `RUN`, so nothing foreign executes. A companion test builds through the legacy builder deliberately, pinning the guard above. Skips gracefully without Docker.

`temps-agent` was in no CI group at all — its tests were compiled and never run. It now joins `docker-deployments` and carries tests for the platform detection that decides what a node reports.

**Manual end-to-end on a mixed cluster**: a QEMU-emulated arm64 `dockerd` (`ARCH=aarch64`) driven by an x86 agent through `DOCKER_HOST`.

| | |
|---|---|
| `temps join` | `Container platform: linux/arm64` |
| `GET /internal/nodes` | `control-plane: linux/amd64`, `worker: linux/arm64` |
| build log | `Cluster spans multiple architectures — building one image per platform platforms=["linux/amd64", "linux/arm64"]` |
| images produced | `app:latest → linux/amd64`, `app:latest-arm64 → linux/arm64` |
| on the arm64 worker | only the `-arm64` tags, both `Architecture=arm64` — the amd64 image was never sent |
| running them | `-arm64` → `aarch64`, primary → `x86_64` |

The app container itself did not start on that worker: `runc` under nested emulation fails with `can't get final child's PID from pipe`, alongside the DinD-specific `nft` and cgroup-v2 issues that had to be worked around first. Those are limitations of the emulated sandbox, not of this change — but it does mean **a container built by this path has not been observed running on real ARM hardware**.

## Notes for reviewers

- **Cross-builds require BuildKit** on the control plane (Docker 23+ enables it by default). Without it the build now fails with an explicit message rather than producing a mislabelled image.
- **Emulated builds are slow.** Operators who don't want that can pin an environment to one architecture with target nodes/labels, or push a pre-built image.
- **Not covered by CI**: transferring to a node of another architecture, and a container actually starting there. That needs a second real daemon — an `ubuntu-24.04-arm` runner would close it, as a separate change.
- The migration is additive and re-runnable (`ADD COLUMN IF NOT EXISTS`); `Migrator::up`/`down` both exercise it in the `migrations` group.


dviejokfs:
**Status: changes requested. All of it lands here, in phases — no follow-up PRs.**

First, this is unusually thorough work and I want to be specific about why. Discovering that the legacy builder *accepts `--platform` and silently ignores it*, then writing a test that deliberately builds through it to pin the guard, is what most PRs in this area miss. The daemon-vs-binary reasoning in `detect_agent_platform` is correct and applied consistently. Deliberately not caching a failed `docker info` so a late-booting daemon isn't frozen at the wrong architecture is a subtle call and the right one. And volunteering that a container was never observed running on real ARM hardware is rarer than it should be.

I reviewed the full diff. The architecture-awareness half — discovery, `nodes.architecture`, `platform.rs`, the scheduler filter, fail-fast before transfer, the `remove_image` fix — I'd merge today. The `BUILDABLE_ARCHES` allowlist is load-bearing and correctly placed: it's what keeps an agent-reported string from reaching a Docker tag or `--platform` as garbage.

---

## Phase 1 — Safety gate (blocks merge)

The per-platform build loop returns `Err` on *any* platform's failure, including when the native build at index 0 already succeeded. Combined with `required_build_platforms` being driven automatically by cluster topology, and no binfmt detection anywhere in the codebase:

> An operator joins an arm64 worker to an amd64 control plane without first running the `tonistiigi/binfmt` command from the new docs. Every deployment in the cluster now fails at build. Before this PR they had broken ARM replicas; after it, no deployments at all.

Same mechanism, second route: `BUILDABLE_ARCHES` includes `s390x`, `ppc64le`, `riscv64`, `mips64le`. One node reporting any of those — buggy or compromised — forces every unpinned deployment to cross-build for a platform nobody can emulate. The allowlist correctly blocks *junk*; these are valid entries. The real issue is a node-reported field driving cluster-wide build work at all.

1. **Cross-building opt-in, default off.** `required_build_platforms` returns empty unless explicitly enabled — `AppSettings` or environment column (project rule: no env-var config). The important one: it fixes the availability regression *and* removes the trust concern, because operator config drives the build instead of an agent-reported value.
2. **Degrade, don't abort.** A non-primary platform failure drops that platform from `image_tags_by_platform` and lets the existing filter exclude those nodes — that path already yields a clean `NoCompatibleNode`. Only a primary failure aborts.
3. **`control_plane_node_response` → `discovered_platform()`**, not `get_native_platform()`. Everywhere else you're careful about confirmed-vs-fallback; here the console can show the binary's arch as the daemon's, to the one person debugging a mixed cluster. The UI already renders null as `Unknown`.
4. **Audit-log architecture changes** in `NodeService::heartbeat` — it's `tracing::info!` today, and the field now affects placement.

## Phase 2 — Scheduling policy and visibility

Same files, so it belongs in this PR rather than a follow-up.

5. **Anti-affinity shortfall becomes an error.** `schedule_round_robin` wraps modulo `pool.len()`, and the exclusion is silently relaxed when it would empty the pool — so 3 replicas onto 2 compatible nodes reports success. When anti-affinity is on and the compatible pool is smaller than `replica_count`, fail with the number named: *"3 replicas requested with anti-affinity; 2 compatible nodes available (worker-arm excluded: runs linux/arm64, no image for it). Set replicas to 2, add a compatible node, or disable anti-affinity."*
6. **Stacking stays legal when anti-affinity is off** — it's the default and what every single-node install does. Not a shortfall.
7. **Transient shortfalls keep degrading.** Health, stale heartbeat, load threshold — these self-correct, and hard-failing them trades a degradation for an outage.
8. **Surface exclusions in the deploy log.** They're `tracing::info!` today, so a user whose node was skipped sees nothing. Needs exclusion reasons returned alongside assignments so `deploy_image` can write them. Per the project rule on actionable failure paths, this is the difference between "the platform handled it" and "the platform silently skipped a node I'm paying for."
9. **Drop the vestigial** `✅ Successfully deployed {}/{} replicas` — the failure path rolls back and returns first, so it can never print a partial.

## Phase 3 — Docs and evidence

10. Mark cross-building as opt-in in `multi-node/page.mdx` and document enabling it.
11. Evidence per `pr-evidence`: opt-in off → single native build, unchanged behaviour; opt-in on with a platform that can't be emulated → that platform excluded, deployment succeeds on the rest.

---

## Not asking for

The `-arm64` tag-suffix scheme is right given Temps has no registry — I checked whether buildx or a manifest list could replace it and neither fits (the Engine API takes a single platform; a manifest list can't be `docker save`d out of the classic image store). Keep it, and keep the daemon-vs-binary distinction throughout.

## Explicitly deferred

Choosing a build host per architecture (designated builder node; emulation only where actually available) so weak ARM workers and slow emulated builds stop being the only two options. Not blocking — with cross-building off by default, nothing depends on it.

Nice catch on the `remove_image` future bound to `_stream` and dropped, by the way — verified it has no production callers, so the fix is free.

---

## How we'd like to proceed

We'd rather finish this in **your** PR than close it and reopen the work elsewhere — the design is sound and most of the diff merges as written.

Whichever you prefer:

- **You take the phases** — happy to review incrementally, phase by phase, so you're not sitting on a large rebase.
- **We push commits to your branch** — maintainer edits are enabled. Items 2, 3, 4 and 9 are a handful of lines each; item 1 is the only one needing a decision (where the setting lives), and that's ours to make.

Either way authorship stays with you. Just say which you'd like, and if there's a phase you'd rather hand off, take it off your plate and we'll pick it up.

jlucaso1:
@dviejokfs thanks for the head ups. I've implemented item 2, 3, 4 and 9. Now you can work on the item 1 since depends on your choice, feel free to submit changes :)

dviejokfs:
@jlucaso1 thanks — items 2, 3, 4 and 9 all look right, and extending the degrade rule to `verify_built_platform` (plus removing the mislabelled image) was the correct call; I'd only asked for the build-failure path.

I picked up item 1 and the rest, and pushed six commits to this branch. I also stood up a real mixed-architecture cluster to verify it, which turned up three defects worth calling out.

## Test environment

Control plane on **linux/arm64**, worker on **linux/amd64** — an emulated `docker:dind` under QEMU driven by an arm64 `temps agent` via `DOCKER_HOST`. The inverse of your setup, and it exercises the same daemon-vs-binary distinction: `temps join` printed `Container platform: linux/amd64` from an arm64 binary.

## Verified end to end

| | |
|---|---|
| `GET /internal/nodes` | `control-plane: linux/arm64`, `worker-amd64: linux/amd64` |
| audit on daemon swap | `NODE_ARCHITECTURE_CHANGED` — `from: linux/amd64 → to: linux/arm64` |
| cross-builds **off** (default) | one native build, no `-amd64` tag, though the cluster spans both |
| cross-builds **on** | `Building '…:latest' for linux/arm64` + `Building '…:latest-amd64' for linux/amd64`, both succeeded |
| per-node tag selection | `Replica 1 scheduled on node 'worker-amd64'` → `Saving image '…:latest-amd64' to tar` |
| transfer | `Image '…:latest-amd64' transferred to node 'worker-amd64' successfully` |
| exclusion in deploy log | `Skipping node 'worker-amd64' runs linux/amd64, and this deployment only has images for [linux/arm64]` |
| replica shortfall | deploy **failed**: `3 replicas requested with anti-affinity, but only 1 node(s) can run this image (…). Set replicas to 1, add a compatible node, or disable anti-affinity` |

Your `multiarch_build_test` also passes against a live daemon here — I confirmed via `docker events` that it really creates both tags rather than skipping.

## Three defects the cluster caught

1. **`523e5311`** — my first cut of the shortfall check was unreachable in the case that matters. When the architecture filter removes *every* remote node, the "no eligible nodes" branch returns `replica_count` Local assignments and returns first. On the real cluster all three replicas stacked on the control plane and the deploy reported success. Moved ahead of that branch, regression test added.
2. **`72613bf5`** — `crossArchitectureBuilds` existed on `DeploymentConfig` but was carried by neither the project deployment-config PATCH nor the environment settings PUT, so the feature could not be turned on at all. Added to both.
3. **`58bf6212`** — my docs named `cross_architecture_builds`, but `DeploymentConfig` is `rename_all = "camelCase"`, so operators must send `crossArchitectureBuilds`. Fixed, with a curl example.

## What still isn't verified

Container **start** on a foreign-architecture node — the same wall you hit. Here it fails as:

```
rosetta error: failed to open elf at -exec-root=/var/run/docker
```

I confirmed it's the sandbox, not this change, by running the transferred image with a plain `docker run` against that daemon: identical failure, no Temps involved. Closing it needs a real second machine (or the `ubuntu-24.04-arm` runner you suggested).

Item 6 (stacking still legal with anti-affinity off) is unit-tested only — the code only intervenes when `anti_affinity == true`.

## State

676 tests pass, `clippy --all-targets -- -D warnings` clean, `cargo fmt` clean. Web SDK regenerated against the live OpenAPI, so `architecture` and the new setting are typed.

Remaining from my earlier list: the audit record still has `ip_address: None`. Your own comment gives "something is impersonating the node" as a reason to audit it, and that's the one field that would distinguish it from an operator repointing a daemon — `node_heartbeat` would need `ConnectInfo` threading in. Happy to take it if you'd rather not.

jlucaso1:
The item of ip_address closed in 3a9d33dd

dviejokfs:
> The item of ip_address closed in [3a9d33d](https://github.com/gotempsh/temps/commit/3a9d33dd20b79e88b12f429ec2c8dbe021cfa646)

perfect, thanks! btw with how many nodes are you testing the cluster?

jlucaso1:
> btw with how many nodes are you testing the cluster?

Two nodes: the amd64 control plane plus one emulated arm64 worker (docker:dind under QEMU, agent driving it over DOCKER_HOST), with a single replica.

I have an arm64 vps with coolify and another x64 vps, so I tested everything emulated locally before pushing this to my real servers

dviejokfs:
Merged 🎉 — thanks @jlucaso1, this was a genuinely good piece of work.

CI came back fully green on the final head (22 checks), and the peer-address commit closes the last item from the review.

## Where it landed

Everything from the three phases is in:

- **Cross-architecture builds are opt-in, off by default** — `crossArchitectureBuilds` on the deployment config, settable per project or per environment (environment wins, and can turn it *off* for a project that has it on)
- **A secondary platform failing degrades** instead of taking the whole cluster's deployments down
- **Nodes the scheduler passes over are named in the deploy log**, so a worker going unused is visible rather than silent
- **Anti-affinity refuses impossible replica counts** when an architecture exclusion causes the shortfall, and tells you the number to set
- **Architecture changes are audited**, now with the peer address

## Please try it on your real hardware

You mentioned you have an arm64 VPS and an x64 VPS. That's the one thing neither of us could prove: **a container actually starting on a foreign-architecture node.** You hit `runc`/`can't get final child's PID from pipe` under nested emulation; I hit `rosetta error: failed to open elf` in emulated DinD — and I confirmed it's the sandbox by running the transferred image with a plain `docker run` on that daemon, same failure, no Temps involved.

Everything up to that point is verified end to end on a real mixed cluster (arm64 control plane + amd64 worker over `DOCKER_HOST`): platform discovery, the audit trail, per-node tag selection, and the tar transfer landing on the foreign-arch agent. It's only the final `docker start` that no sandbox here can do.

If you get a chance on the nightly, the run that would close it:

1. Join the arm64 VPS as a worker to the x64 control plane (or vice versa)
2. Enable cross-builds on the environment:
   ```bash
   curl -X PATCH https://<control-plane>/api/projects/<id>/deployment-config \
     -H 'Content-Type: application/json' \
     -d '{"crossArchitectureBuilds": true}'
   ```
3. Install QEMU on the control plane if it isn't there: `docker run --privileged --rm tonistiigi/binfmt --install <arch>`
4. Deploy with 2 replicas and anti-affinity on, so one lands on each node

What I'd most like to know:

- Does the container **start and stay healthy** on the foreign-arch node?
- Is the emulated build time tolerable on a real VPS, or bad enough that we should prioritise picking a build host per architecture rather than always cross-building on the control plane?
- With cross-builds **off** (the default), does an incompatible node get cleanly excluded with the `Skipping node …` line in your deploy log?

That last one matters most for the default path, since it's what every existing user gets.

## Two follow-ups worth knowing about

Neither is caused by this PR, both surfaced while testing it:

1. `download_repo` swallows clone errors — a failed clone leaves the deploy log ending mid-step with no reason given
2. `repo_owner`/`repo_name` and `git_url` can drift, and there's no API to fix `git_url` — branch resolution uses one, the clone uses the other

I'll file those separately.

Thanks again — the legacy-builder discovery and the daemon-vs-binary reasoning were the parts that made this solid, and being upfront about what you hadn't been able to verify is what let us pick up exactly where you left off.

PR #501, opened 2026-08-01
feat(deployer): attach app containers to required extra Docker networks

dviejokfs:
Recovers #88 (fork branch `elaine/required-extra-docker-networks`, conflicting with main) onto a fresh branch off `origin/main`.

Original author: @adamlevineagent / codex-elaine (preserved via `Co-authored-by`).

## Summary
- add `TEMPS_DOCKER_EXTRA_NETWORKS` for self-hosted installs that need app containers attached to external Docker dependency networks before startup
- add `DeployRequest.extra_networks` for per-request attachments; defaults to empty, backwards compatible
- fail the deployment when a configured required network is missing or cannot be attached, instead of a green deploy with dependency DNS failures
- normalize configured networks (drop empties, duplicates, primary and overlay names)
- document the operator env var in `docs/reference/environment-variables`

## Conflict resolution vs. #88
Main moved on since May, so three things were re-done rather than taken verbatim:
- `temps-deployer/src/plugin.rs` — kept main's build-limits + ADR-024 cluster-DNS setup block and chained `.with_extra_networks(server_config.docker_extra_networks.clone())` onto `DockerRuntime::new`, reusing the `config_service` main already binds.
- `docs/reference/environment-variables/page.mdx` — kept main's newer server table (admin listener rows, updated `TEMPS_CONSOLE_ADDRESS` description) and added only the new `TEMPS_DOCKER_EXTRA_NETWORKS` row.
- `temps-deployer/src/docker.rs` tests — kept main's DNS-failover test and appended the new `unique_test_name` / `alpine_deploy_request` helpers.

Four call sites added to main after #88 was written needed the new fields, which is why #88 could not just be rebased:
- 3 × `DeployRequest` literals (`docker.rs` ×2, `remote.rs`) → `extra_networks: Vec::new()`
- 1 × `ServerConfig` literal (`temps-deployments/src/jobs/take_screenshot.rs`) → `docker_extra_networks: Vec::new()`

## Verification
Run locally on this branch, on top of `324ad120`:

```
cargo fmt --all                                              # clean
cargo check --all-targets --workspace                        # no errors
cargo clippy --workspace --all-targets -- -D warnings        # no errors
```

Test suites are deferred to CI on this PR (the new extra-network integration tests require a live Docker daemon and were not executed locally).

Closes #88


dviejokfs:
Security audit (security-auditor): **APPROVE-WITH-CONDITIONS** — both merge blockers fixed in 2f0b5921.

**MEDIUM — per-request `extra_networks` was unbounded.** `attach_required_networks` merged caller-supplied networks with the operator's set and treated both identically. `POST /agent/containers/deploy` deserializes `DeployRequest` straight from the body, so a token holder could bridge a container onto any host network (another project's DB network, the control plane's). Requests may now only *narrow* `TEMPS_DOCKER_EXTRA_NETWORKS`, never widen it; regression test `test_deploy_rejects_request_network_outside_operator_allowlist`.

**MEDIUM — primary/overlay filter was name-only, bypassable by ID.** `normalize_extra_networks` compared raw strings while the attach path accepts network IDs, so the overlay's 64-hex ID sailed past the guard and attached without `install_overlay_routes`. Identifiers are now resolved to a canonical ID before both the allowlist check and the reserved-network guard; `test_deploy_attaches_required_network_by_id` now configures by name and requests by ID to cover it.

Non-blocking items also addressed: raw bollard errors no longer returned to API callers (logged, generic summary returned); failed post-error cleanup now surfaces in the returned error with the `docker rm -f` to run; docs state the setting is host-wide across all projects and therefore single-tenant only.

Explicit negatives from the audit: no CRITICAL, no HIGH, no injection (bollard typed client, no shell/path construction), no SQL changes, no new secret exposure, no dependency changes. Attach runs strictly before `start_container`, so a failed attach can never leave a running partially-networked container.

Reviewed and merging with all 22 CI checks green, including Integration Tests (docker-deployments).

PR #513, opened 2026-08-02
fix(nixpacks): temps-nixpacks crate, resolved security issues, upgrade corepack pin for pnpm@11 compatibility

brunyeestudio:
## Description

**WIP**

image

## Type of change

- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update

## Checklist

- [ ] All new and existing tests pass (`cargo test --lib`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,189 +3,6 @@
 version = 4
 
 [[package]]
-name = "actix-codec"
-version = "0.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5f7b0a21988c1bf877cf4759ef5ddaac04c1c9fe808c9142ecb78ba97d97a28a"
-dependencies = [
- "bitflags 2.11.1",
- "bytes",
- "futures-core",
- "futures-sink",
- "memchr",
- "pin-project-lite",
- "tokio",
- "tokio-util",
- "tracing",
-]
-
-[[package]]
-name = "actix-http"
-version = "3.12.1"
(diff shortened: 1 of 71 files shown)


dviejokfs:
Thanks @brunyeestudio!! I've been thinking about this, yesterday I was working on an alternative which for the long term it may be better than nixpacks/railpack.

it's called `autopack`, https://github.com/gotempsh/autopack 

The goal is to support nixpack+railpack while testing that most applications deploy with zero config to achieve zero-config deployments on as many deployments as possible

The goal is to also support LLB backend as railpack but libraries are not yet there, but developing them with AI will be possible, for now, only generating the Dockerfile is enough.

what do you think?

dviejokfs:
I think it makes sense to have it as another independent project so that other projects could use it from rust, currently, only nixpacks can be used as a rust SDK with limited support.

brunyeestudio:
> Thanks @brunyeestudio!! I've been thinking about this, yesterday I was working on an alternative which for the long term it may be better than nixpacks/railpack.
> 
> it's called `autopack`, https://github.com/gotempsh/autopack
> 
> The goal is to support nixpack+railpack while testing that most applications deploy with zero config to achieve zero-config deployments on as many deployments as possible
> 
> The goal is to also support LLB backend as railpack but libraries are not yet there, but developing them with AI will be possible, for now, only generating the Dockerfile is enough.
> 
> what do you think?

100%, sounds like a great approach!! Happy to close this off 👌

PR #521, opened 2026-08-02
feat(ai): let the assistant propose metric alert rules from a project's own telemetry

dviejokfs:
## What

Adds a **Suggest alerts with AI** button to the metric alerts page. It opens a chat that enumerates what the project actually emits, grounds candidate thresholds in real queried values, backtests anomaly detectors, and stages each proposed rule as a separate confirm-gated write action.

Most of the machinery already existed (`ai_pending_actions` propose-then-confirm, the ADR-024 read CLI, `preview_alert`). The gaps were allowlist entries, one seeding provider, and an onboarding path.

## The interesting decision: read-only POSTs

HTTP method is this codebase's structural proxy for "does this mutate?", and the AI read allowlist is GET-only for that reason. `preview_alert` breaks the assumption: it is documented read-only, guarded by `OtelRead`, persists nothing, and is a POST only because its request body is a whole detector config.

Putting it in the **write** allowlist would be wrong in the other direction — that path is propose-then-confirm, so a backtest would stage a pending action and wait for a human, turning the cheap "check before you propose" step into the very thing it exists to inform.

So there is now a second, separately-reviewed allowlist for side-effect-free POSTs (`ai_read_safe_posts()`), with the rule stated at the definition: **an operation belongs there only if it writes nothing.** Tests prove the two lists are disjoint (naming a POST in the GET list doesn't admit it, and vice versa) and that PATCH/PUT/DELETE are structurally unreachable from the read index regardless of what someone adds.

## Allowlist changes

| Operation | List | Why |
|---|---|---|
| `list_alerts`, `get_alert` | read | without them the model proposes duplicates of rules that already exist |
| `preview_alert` | read (safe POST) | backtest, so a proposal arrives with evidence rather than a guessed threshold |
| `create_alert`, `update_alert` | write | reversible, changes no running workload |
| `delete_alert` | **excluded** | deleting an alert silently removes monitoring — only noticed when the incident it would have caught happens |

## Onboarding

The button is always visible, because the interesting case is the one where the feature *isn't* set up: hiding it leaves a self-hosted user with no way to discover the capability or find the two settings that unlock it. New `GET /projects/{id}/ai/readiness` reports the three prerequisites separately — they fail for different reasons and are fixed in different places (one instance-wide, two per project), so a single "AI unavailable" would leave the user with no idea which of three places to go.

- no AI provider → dialog pointing at Settings → AI Providers
- write actions off → one-click enable behind a confirmation stating what it grants
- ready → opens the chat

CLI parity: `temps ai readiness [project] [--json]`.

## Drive-by fix (found while testing this)

When the AI provider errored mid-turn, the round loop just `break`ed. The SSE stream closed with zero events, so the chat showed the user's own message and then nothing, forever — indistinguishable from a hang, with no hint the provider key was at fault. The client already handled an `error` event; the server never sent one. It now reports the provider error and names the setting that fixes it. Regression test added.

## Evidence

Verified on a local instance (worktree slot 3, fresh DB, `alert-demo` project).

**Allowlists resolve against the real OpenAPI doc** — startup logs, no `unresolved` warning:

```
AI read tool: indexed read operations resolved_count=214 allowlist_count=214
AI write tool: indexed write operations resolved_count=22 allowlist_count=22
  resolved=[... "create_alert", "update_alert", ...]
```

**Readiness endpoint** reflects each gate independently:

```
$ curl -b cookies localhost:8110/api/projects/1/ai/readiness
{"ai_configured":false,"chat_enabled":true,"write_actions_enabled":false}
# after configuring a provider:
{"ai_configured":true,"chat_enabled":true,"write_actions_enabled":false}
```

**Browser walkthrough** (agent-browser, `/projects/alert-demo/metrics/alerts`):

1. Button renders next to "New alert" — `button "Suggest alerts with AI"`
2. No provider → `heading "Configure an AI provider first"` + working link to `/settings/ai-providers`
3. Provider present, write actions off → `heading "Enable AI write actions?"`
4. "Enable and continue" → toast `AI write actions enabled for this project`, dock opens on the `Suggest alerts` conversation

**Seed is correct** — the conversation row and system message:

```
 context_type  | context_id |     title
---------------+------------+----------------
 alert_suggest | 1          | Suggest alerts

--- Existing metric alert rules ---
None. This project has no metric alert rules at all, so nothing about it is currently being watched.
```

**Silent-failure fix**, before vs after, same forced provider error:

```
# before: stream closed with zero events, UI showed nothing
# after:
event: error
data: AI provider error: no reply was produced. AI provider error for
      'chat.alert_suggest.tools': Encryption error: ... Check the provider's key
      and model in Settings → AI Providers, then try again.
```

and the same text rendered in the chat panel in the browser.

**Tests**: `cargo test --lib -p temps-ai-chat -p temps-ai-api-tools` → 153 passed; `-p temps-cli ai_tool_allowlist` → 4 passed. `cargo clippy --all-targets -- -D warnings` clean on all three crates. `tsc --noEmit` clean in `web/`; no new errors in `apps/temps-cli` (19 pre-existing, unchanged).

## Not verified

**No live model turn.** No AI provider credential was available in this environment, so the end-to-end path — model calls `list_metric_names`, backtests via `preview_alert`, stages `create_alert` proposals, human confirms — has **not** been exercised against a real model. Everything up to the model call is verified, and the tool wiring is confirmed (`chat_stream_turn: ... tools=2` in the logs, so both the read CLI and `temps_write` were offered). The quality of the model's actual proposals is unproven and wants a run with a real key before merge.

Related: metric alerts have no CLI coverage at all today (pre-existing). Only the one new endpoint got CLI parity here; full alert CRUD in `apps/temps-cli` is worth a follow-up.


dviejokfs:
## Update: verified against a real model

The earlier "not verified" caveat is now closed. I ran the flow end-to-end against a **local Ollama model** (`qwen3:30b`, wired in as an OpenAI-compatible provider via `base_url`), against **real seeded telemetry** — a week of hourly OTLP points across four metrics, with a diurnal traffic shape, two deliberate latency/error incidents, and a slow memory creep.

Doing so found three real defects, now fixed in `4e1112ec`, plus one that remains open.

### What works

The full loop is proven:

```
temps telemetry-metrics list_metric_names
temps telemetry-metrics query_metrics --metric_name http.server.duration
temps telemetry-metrics query_metrics --metric_name http.server.errors
temps telemetry-metrics query_metrics --metric_name http.server.requests
temps telemetry-metrics query_metrics --metric_name process.runtime.memory.rss
temps_write  create_alert …          →  ai_pending_actions row, status=proposed
```

The proposal renders in the chat as a card with the full params JSON and **Confirm & run** / **Reject**, and confirming replays the real mutation through the router.

The model's reasoning was grounded in data it actually fetched, e.g.:

> Historical queries show max values consistently below 330 million (e.g. 327,225,527.88 in the 2026-07-26T18:00:00Z bucket). A threshold of 350 million (above the observed peak) avoids noise while catching critical spikes.

It also correctly reported "no existing alerts", and in one run correctly *declined* to propose anything when it distrusted the data (my seed script had wrongly declared per-second rates as cumulative counters — a fixture bug, and the model caught it).

### Defects found and fixed

1. **One missing param per round.** The validator reported only the first missing required field. `create_alert` has eight, so the model spent a round per field and hit the round cap before ever staging a proposal. Now all are reported at once — observed recovery went from 8+ rounds (never succeeding) to 3:

   ```
   Missing required parameters for operation 'create_alert':
     aggregation, detection_config, enabled, metric_name, name, window_secs.
   Provide ALL of them in the flat parameters object.
   ```

2. **`$ref` body-property descriptions were discarded.** `detection_config` resolves to a bare `object`, so its sibling description is the only place the required JSON shape is documented — and it was being dropped.

3. **`--help` didn't mark required body fields**, only path params — i.e. not the ones a write call actually fails without.

Plus two prompt fixes for behaviours seen in the runs: the model printing `temps_write …` in a code block instead of calling the tool (proposing nothing, and showing the user text with no button to accept), and asserting a metric had no data without having queried it.

### Still open — worth a follow-up, or a review call

**A structurally invalid proposal can still reach the human.** Across three runs the model consistently guessed `detection_config` as `{"type": "threshold"}` or `{"op": "gt", "value": N}` instead of `{"kind": "static", "comparator": "gt", "threshold": N}`. It never ran `create_alert --help`, so fix (2) above doesn't reach it in practice.

`prepare_write_cli` validates presence and enums but not nested body *shape*, so the bad payload proposes cleanly, the user clicks **Confirm & run**, and only then does it fail:

```
Upstream call to operation 'create_alert' returned HTTP 422:
  detection_config: missing field `kind`
```

The failure is at least reported clearly and the action is marked `failed` rather than silently lost. But a proposal that cannot possibly succeed shouldn't be offered to a human at all — the model should get that error at propose time and self-correct. Fixing it properly means validating required keys of a `$ref`'d body object at prepare time, which is a bigger change than belongs in this PR; happy to do it here if reviewers prefer.

Second, smaller: **the model never called `preview_alert`**, despite the prompt asking it to backtest. In one run it went further and *claimed* it had ("Backtested via preview_alert, this rule would have fired 0 times") without ever making the call. The tool is correctly wired and reachable — this is model behaviour. A frontier model may well do better than a local 30B, but the prompt shouldn't invite an unverifiable claim either.

### Caveat on all of the above

This is one 30B local model. It establishes that the plumbing works and surfaced real usability defects, but it is not evidence about proposal quality from a frontier model.

dviejokfs:
## Fixed: structurally impossible proposals no longer reach a human

The open item from the previous comment is closed in `dd265d59`.

### What was wrong

`prepare_write_cli` checked that required params were *present* and that enums held legal values, but never looked *inside* an object-valued body field. `detection_config` is an internally-tagged union that reached the model as a bare `object` with no hint of its contents — so it guessed `{"type": "threshold"}` instead of `{"kind": "static", …}` on three consecutive runs. That payload validated cleanly, was staged as a pending action, and failed with a 422 only after the user clicked **Confirm**.

### The fix

The index now resolves a `$ref`'d body object into its required keys, understanding both shapes that matter:

- a plain struct → its `required` list;
- utoipa's internally-tagged-enum encoding (`oneOf` of `allOf`s, each pairing a payload `$ref` with an inline object pinning the tag to a single-valued enum) → discriminator name plus per-variant required keys.

Validation then checks the discriminator and that variant's required keys. **The error carries the accepted shape, not just the complaint** — the model demonstrably does not run `--help` before a write call, so the rejection is the one place the shape is guaranteed to be read:

```
Parameter 'detection_config' has the wrong shape for operation 'create_alert':
  missing the 'kind' discriminator.
Expected object, one of:
  {"kind": "static", "comparator": …, "threshold": …} |
  {"kind": "anomaly", "deviations": …}
```

`--help` renders the same line, for when it does.

This is deliberately a **shape** check, not full JSON Schema validation: it separates calls that cannot possibly succeed from ones that can, and leaves field-level type-checking to the API. Being stricter risks rejecting calls the router would have accepted.

### Verification

A test against the **real `OtelApiDoc`** (not a synthetic fixture) asserts both directions — the exact payload the live model produced is now refused at prepare time, and a well-formed static detector is still staged:

```rust
// {"type": "threshold", "threshold": 500}  -> WritePrepareOutcome::Invalid, naming `kind` + `static`
// {"kind": "static", "comparator": "gt", "threshold": 500} -> WritePrepareOutcome::Prepared
```

Plus unit tests for each rejection case: missing discriminator, unknown variant (lists the known ones), missing variant keys, plain-object required keys, and a scalar where an object belongs — and one asserting a valid object reaches the API byte-for-byte.

`162 passed` across `temps-ai-api-tools` + `temps-ai-chat`, `5 passed` in the `temps-cli` allowlist module, clippy `-D warnings` clean on all three.

### Still model behaviour, not a code defect

`preview_alert` remains under-used: the tool is wired, allowlisted, and reachable, but the local 30B model usually skips it and in one run claimed to have used it without doing so. The prompt now asks for it explicitly; enforcing it (e.g. refusing to propose an anomaly detector that was never backtested) would be a further change and I'd rather a reviewer weigh in than assume.

dviejokfs:
## The AI was right and I was wrong: a real silent-wrong-answer bug

Adding the entry point to **Explore** (done — see below) surfaced something more important. In an earlier run the model reported that every metric returned identical ~62 million values and refused to propose anything, calling it "telemetry misconfiguration". I dismissed that as my seed fixture being wrong. It wasn't.

### The bug

The model called:

```
temps telemetry-metrics query_metrics --metric http.server.duration
```

The real flag is `--metric_name`. `parse_flags` accepts any `--name value` into the params object, and `build_request_parts` only ever *reads* parameters it knows about — so `metric` was silently dropped, `metric_name` was absent, and the query ran **unfiltered**. Reproduced directly:

```
# with the correct flag
?project_id=1&metric_name=process.runtime.memory.rss
  -> avg_value: 308109589.47   count: 1

# with the flag dropped (what actually happened)
?project_id=1
  -> avg_value: 61622034.82    count: 5
     min_value: 0.338          max_value: 308109589.47
```

`61,622,034` is the model's "~62 million": the mean of latency-in-ms, error rates and memory-in-bytes averaged together. **HTTP 200, well-formed body, completely meaningless.** Nothing downstream can detect that — and the Explore UI showed correct per-metric values the whole time (248.92 / 0.208 / 89.98 / 486.68M), which is why it took a screenshot from you to catch it.

This is the worst failure mode for this feature specifically: the number becomes a threshold someone gets paged by.

### The fix

Unknown parameters are now an error rather than a shrug:

```
Unknown parameter(s) for operation 'create_alert': 'condition'.
Valid parameters: aggregation, detection_config, dynamic_alerts, enabled,
for_duration_secs, group_by, ... Nothing was queried — re-run with the correct name.
```

Three deliberate details:

- **Suggests the intended name** on an obvious near-miss — prefix/containment (`metric` → `metric_name`) or a bounded edit distance. Below the threshold it suggests nothing, because a wrong guess sends the caller further astray than none.
- **Says "nothing was queried"** explicitly. Without it the model reads the failure as an empty result set and concludes the metric has no data — which is exactly what it did in the earlier run.
- **Lists every valid parameter**, so the correction takes one round.

### Full chain, verified live

One run against the local model, showing every fix in this PR firing in order and the loop closing:

```
1. list_metric_names                            -> 4 metrics
2. query_metrics --metric_name http.server.duration
     -> avg 245.62  min 174.31  max 316.93      ← real filtered data, not 62M
3. create_alert --condition 'max_value > 320'
     -> Unknown parameter(s): 'condition'. Valid: ...   ← NEW
4. create_alert --detection_config '{"operator":"gt",...}'
     -> Missing required parameters: aggregation, enabled, name, window_secs  ← batched
5. ... -> detection_config wrong shape: missing 'kind'. Expected {"kind":"static",...}  ← shape check
6. ... -> {"status":"proposed", "action_id":"1e702ccc..."}
```

Proposed rule, threshold grounded just above the observed peak of 316.93:

```json
{"name":"High HTTP Server Duration","metric_name":"http.server.duration",
 "aggregation":"max","window_secs":300,"for_duration_secs":300,"severity":"critical",
 "enabled":true,"detection_config":{"kind":"static","comparator":"gt","threshold":320}}
```

Confirming it returned `"status":"executed"` and created rule id 1. **The end-to-end path — discover → query → propose → human confirms → rule exists — now works with no 422.**

### Explore entry point

`SuggestAlertsButton` is now in the metrics explorer header next to Refresh, which is where the question actually occurs to you — you're looking at a chart. With a metric selected the opening prompt narrows to that metric ("I'm looking at `X`, is it worth alerting on?") instead of restarting the whole survey. The conversation stays keyed on the project, so returning resumes the same chat rather than forking one per metric.

`166 passed` across the two crates, `tsc --noEmit` clean, clippy `-D warnings` clean.

dviejokfs:
## Two more fixes, and the alert lifecycle verified end to end

### 1. Field types and enums inside object body params (`fc5e35c6`)

The shape check in `dd265d59` verified required keys were *present*, which catches only half the mistakes. The model then produced:

```json
{"kind": "static", "threshold": "0.5", "comparator": ">"}
```

Every key is there. Both values are wrong, and wrong in ways that look right — `"0.5"` is a quoted number where the API wants an `f64`, and `>` is the natural way to write a comparator while the schema accepts only `gt|gte|lt|lte` (its description even warns "NOT the SQL operators", text the model never sees). It validated, was staged, the user clicked Confirm, and got `invalid type: string "0.5", expected f64`.

`ObjectShape` now carries each field's type and enum members rather than just its name, resolving a property's own `$ref` so an enum defined as its own component still yields its values. The error shows the offending value with quotes intact, since `0.5` and `"0.5"` are indistinguishable in a message that merely echoes it. Integers are still accepted where a number is wanted, or every whole-numbered threshold would break.

Separately: `coerce` only ever attempted `i64`, so a bare `--threshold 0.5` became the *string* `"0.5"` and hit the same 422 with no object involved. It now parses floats, guarding `inf`/`NaN` (they parse as f64 but can't be represented in JSON and would serialize as `null`).

Live run, every guard firing in order and recovering to a valid call:

```
query_metrics --metric http.server.errors      -> Unknown parameter 'metric'. Valid: ...
query_metrics --metric_name http.server.errors -> real data
create_alert --detector static --threshold 0.5 -> Unknown parameter(s): 'detector', 'threshold'
create_alert --detection_config '{"type":"static",...}'
                                               -> Missing: aggregation, enabled, name, window_secs
                                               -> wrong shape: missing 'kind'. Expected
                                                  {"kind":"anomaly","algorithm": one of robust|basic|agile|ewma, ...}
create_alert --detection_config '{"kind":"static","comparator":"gte","threshold":0.5}'
                                               -> {"status":"proposed"}
```

Note the rendered shape now carries types and enum members — that's what the model needed to get `comparator` right.

### 2. Copy button on every chat message (`ab98be74`)

Added per-message copy, revealed on hover and on keyboard focus. An assistant turn copies its prose only — tool cards are UI around the work, not something anyone pastes into an issue — and a turn that ran tools but produced no prose renders no button rather than one that silently copies an empty string.

Doing it properly meant fixing `CopyButton`, which all 33 copy sites in the console go through. It called `navigator.clipboard.writeText(value)` without awaiting or catching, so a rejected write **showed the success checkmark anyway** and left an unhandled rejection. Worse, `navigator.clipboard` doesn't exist outside a secure context — not an edge case for this product, that's a self-hosted instance on a plain-http LAN address, where every copy button in the app appeared to work and copied nothing.

It now awaits the write, falls back to a selection-based copy when the API is absent, and reports failure honestly (distinct icon + toast naming the reason). All three paths verified in the browser:

| path | result |
|---|---|
| write rejected (headless perms) | `lucide-x` icon + toast: *"Failed to execute 'writeText'… Write permission denied"* |
| no clipboard API (insecure origin) | falls back to `execCommand('copy')`, checkmark, temporary textarea removed |
| correct text captured | `"I am looking at the metric \`http.server.errors\`…"` |

### 3. The alert lifecycle actually works

Worth recording since it was the last unverified link. The rules created through this flow initially showed `Unknown` — not a bug: the seeded telemetry had gone stale (newest point 9h14m old) while the rules evaluate a 300s window, so `unknown` correctly meant "no data". With fresh data:

```
ok       199.27   (both rules evaluating real values)
...spike sustained...
ok       501.13   ← over threshold 320, not yet sustained
ok       478.30
firing   501.13   ← for_duration_secs: 300 satisfied
```

Alarm row created: `Metric threshold breached: High HTTP Server Duration | firing | critical`.

So an alert the assistant proposed, that a human confirmed, has now detected a real breach and raised a real alarm — and the `for_duration_secs` the model chose correctly suppressed a single-scrape blip.

`173 passed` across `temps-ai-api-tools` + `temps-ai-chat`, `5 passed` in the `temps-cli` allowlist module, `tsc --noEmit` clean, clippy `-D warnings` clean.

dviejokfs:
## `preview_alert` now gets invoked — and turn budgeting reworked

### Why it was never called

Not discoverability — it was listed in `alerts --help` all along with complete flags. **It rejected every detector kind except `anomaly`**, and every rule the flow proposes is a static threshold. The model wasn't being lazy; the tool 400'd on the rule at hand. `f38d0492` makes static detectors backtestable through `Comparator::breaches`, the same call the evaluator makes, so a backtest can't drift from production:

```
threshold 320 on http.server.duration  ->  22 breaches / 171 points
threshold 1                            ->  171 / 171
```

The same probe surfaced a second bug: utoipa encodes `Option<String>` as `type: ["string","null"]` — a type *union* — which was being read as `SchemaType::Array` and reported as `"array"`. Every optional field in every write body was advertised to the model as wanting a list (`--start_time <array>` for an RFC 3339 string).

### Making it happen

A stronger prompt wasn't enough — I tried, and the model still skipped it. `create_alert`/`update_alert` are now **refused** unless a successful `preview_alert` ran that turn (a failed one doesn't count, or one malformed call becomes the way past the gate).

The first version of the refusal *described* the command; the model rediscovered the flags by trial and error and the turn ended with nothing staged — worse than the un-backtested proposal I was preventing. It now spells the command out. Live result:

```
create_alert (no backtest)   -> GATE FIRED
preview_alert                -> breach_count = 28 of 171     ← correct flags, first try
create_alert threshold 315   -> PROPOSED                     ← was 300 before the backtest
```

> **Backtest result**: Would have fired 28 times over the sampled window (28/171 points, ~16% breaching), balancing signal/noise.

That claim is now true and server-checkable, and the tool *changed the answer* rather than decorating it.

### Turn budgeting: rounds were the wrong governor

I initially raised the step cap 10 → 14 to fit the extra backtest. That was papering over the real problem, and @dviejo pushed back on exactly the right thing: a chat should solve the task, not stop every N turns.

Measured across three live runs — **30–50% of every budget went to rounds that produced only a rejection**, with the productive work costing 6–7 rounds:

| run | tool calls | rounds producing only an error | outcome |
|---|---|---|---|
| A | 8 | 4 | nothing staged |
| B | 10 | 4 | nothing staged (hit cap) |
| C | 10 | 3 | proposed |

One counter can't serve two opposite goals: a long productive task wants a generous budget, a stuck one should be killed fast. Ten was too tight for the first and far too loose for the second — a model looping on a validation error got ten attempts before anything stopped it.

The chat already has the right controls: every tool call is visible, there's a Stop button, and closing the panel drops the SSE stream which cancels the upstream request. So the step cap becomes a **runaway backstop (40)**, not a governor, and the real bounds are:

- **3 consecutive unproductive rounds ends the turn** — a stuck loop now dies in 3 rounds instead of 10.
- **192 KiB of carried-forward tool results.** Every result is replayed on every later round, so the *transcript* — not the step count — is what actually bounds a turn. Without this, a higher cap trades "stopped early" for "blew the context window", which is worse for being unexplained. Older results are stubbed in place with `tool_call_id` intact, since providers reject a `tool_calls` block whose answers have gone missing.
- **The user is told when a bound ends a turn**, before the salvage answer. A confident summary written under duress reads as a complete one — the same dishonesty as the backtest that never ran.

Classification parses the read CLI's defined success shape rather than sniffing for words, and an unrecognised tool's output counts as productive — misreading it as failure would cut a working turn short. The backtest gate uses the same function instead of the string matching it shipped with an hour ago.

Verification: the broad "review every metric and propose rules" run now completes **naturally in 9 rounds** — backtested, proposed, answered, no truncation and no stop-reason. Unit tests cover a stuck model (stops in ≤6 rounds and says why), an endlessly-productive one (bounded at 40, told it was cut short), oldest-first trimming, and the productive/rejection classification.

`617 passed` across the three crates, `6 passed` in the `temps-cli` allowlist module, clippy `-D warnings` clean.

### For reviewers

Two changes here are wider than the feature and worth a look: `MAX_ROUNDS` 10 → 40 with the new bounds replacing it, and the backtest gate living in the generic tool loop keyed off a small `BACKTEST_REQUIRED` table. Both are deliberate, but they affect every chat context, not just alerts.

PR #545, opened 2026-08-03
fix(proxy): keep backend latency for streaming sessions

dviejokfs:
Follow-up to #514, addressing the review findings on that PR.

## 1. Backend latency was over-excluded (the substantive fix)

#514's carve-out returned from `record` before the upstream histogram was touched, so a WebSocket or SSE session contributed **no** backend-latency sample. That was too broad. `upstream_ms` is measured at the first upstream response header — the `101`, or the SSE headers — *before* the tunnel starts carrying traffic, so it is a genuine latency even though the session's total duration is not. As merged, a slow WebSocket handshake was invisible in `proxy.upstream_duration_*`.

Streaming sessions now contribute to the backend histogram again.

**This required splitting the denominator.** `proxy.self_duration_*` divided by `upstream_count`. Once streaming sessions land in the upstream histogram, reusing that count would drag the self-time mean toward zero — the same class of distortion #514 existed to fix, pointing the other way. The self histogram now carries its own `self_count`.

A regression test pins it. Reverting just the denominator gives:

```
test test_self_avg_denominator_excludes_streaming_sessions ... FAILED
assertion `left == right` failed
  left: 5.0       <- diluted by 3 streaming sessions
 right: 20.0
```

## 2. Documented what the carve-out does not catch

#514 read as if it closed the whole class. It does not. Still landing in the latency histograms:

- HTTP/2 WebSockets — RFC 8441 extended `CONNECT` answers `200`, not `101`
- gRPC streaming (`application/grpc`), chunked long-poll / NDJSON feeds
- **Large or slow downloads** — `elapsed` covers the whole body transfer

Widening detection would mean classifying on response *duration* rather than shape, which needs a threshold nobody can pick correctly for every deploy. The two covered cases are the ones that produce hour-long sessions by design; the rest are bounded by the upstream read timeout. That reasoning is now in the code rather than implied.

## 3. Per-series observation sets + operator note

The gauges do not all divide by the same number of requests, and nothing said so. Added a table to the module docs:

| series | observation set |
|---|---|
| `proxy.requests*`, `proxy.error_rate_percent` | every request |
| `proxy.request_duration_*` | every request except streaming sessions |
| `proxy.upstream_duration_*` | every request that reached an upstream, streaming included |
| `proxy.self_duration_*` | proxied requests except streaming sessions |
| `proxy.streaming_*` | streaming sessions only |

Plus an upgrade note: the duration series drop wherever streaming traffic exists, so alert thresholds tuned before #514 should be re-checked — one that used to fire may now sit permanently below its trigger.

## 4. Nits

- Unstuck a comment that rustfmt had glued to the trailing comment of the line above, making it read as if it belonged to `skip_tracking`.
- `test_record_classifies_status_codes` recorded a `101` with `is_streaming: false` — a state the proxy cannot produce.
- Corrected the "Streaming sessions" panel description: the read path averages per collection interval, so the plotted value is not a bucket total.
- Noted on the backend-latency panel that it now includes streaming handshakes.

## Test plan

- `cargo test --lib -p temps-proxy` — **449 passed, 0 failed**.
- `cargo clippy -p temps-proxy --all-targets -- -D warnings` — clean.
- `tsc --noEmit` — clean.
- New test verified to fail against the un-split denominator (output above).

Not re-run: the live-proxy WebSocket/SSE exercise from #514. This change alters which histogram a sample lands in, not whether the session is detected, and the denominator split is covered by the unit test above.

### DIFF
--- a/crates/temps-proxy/src/metrics.rs
+++ b/crates/temps-proxy/src/metrics.rs
@@ -12,6 +12,25 @@
 //! classes and durations to a fixed bucket ladder. No per-request attribute
 //! (path, host, project) ever becomes a metric dimension — per-project traffic
 //! breakdowns come from the proxy request logs instead.
+//!
+//! # Observation sets differ per series
+//!
+//! The gauges do not all divide by the same number of requests, and an
+//! operator reading them side by side needs to know which:
+//!
+//! | series | observation set |
+//! |---|---|
+//! | `proxy.requests*`, `proxy.error_rate_percent` | every request |
+//! | `proxy.request_duration_*` | every request except streaming sessions |
+//! | `proxy.upstream_duration_*` | every request that reached an upstream, streaming included |
+//! | `proxy.self_duration_*` | proxied requests except streaming sessions |
+//! | `proxy.streaming_*` | streaming sessions only |
+//!
+//! Note for operators upgrading: the duration series exclude WebSocket/SSE
+//! lifetimes as of this change, so their values drop wherever streaming
+//! traffic exists. Alert thresholds tuned against the previous (inflated)
+//! numbers should be re-checked — a threshold that used to fire may now sit
+//! permanently below the trigger.
 
 use std::sync::atomic::{AtomicU64, Ordering};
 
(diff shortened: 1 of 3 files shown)


dviejokfs:
## Review + security-audit fixes (`9722826e`)

All Major/Minor findings from the review and the security audit are addressed.

**Major — restored upstream sample re-opened mean poisoning.** `upstream_ms` is time-to-first-header, and `upstream_peer` grants a WebSocket upgrade a 1h read timeout (60× ordinary traffic) selected from the request's own `Upgrade` header. A hung upstream could report ~3_600_000ms and drag `proxy.upstream_duration_avg_ms` into the tens of seconds — the exact distortion this series exists to close, aimed at the backend gauge instead. Now clamped to `MAX_HANDSHAKE_OBSERVATION_MS` (60s, the ordinary read timeout), so a streaming handshake can move the mean no more than any other request already can. Kept rather than dropped, so `upstream_count` stays honest.

Regression test verified against the unclamped code:

```
test test_streaming_handshake_observation_is_clamped ... FAILED
  left: 35999.9      <- unclamped, one hung handshake in 100
 right: 609.9
```

The test also asserts p99 stays *flat* at 1-in-100 — documenting why bounding the mean is the only available defence: the percentiles never react to a lone tail observation however large it is.

**Major — `record` doc contradicted the code.** It still claimed streaming sessions were kept out of *every* duration histogram after this PR added them back to the backend one. Corrected, along with the stale relaxed-atomic-add counts in the same comment.

**Minor — SSE classification was substring-matched.** The upstream `content-type` used `contains`, so `text/html; note=text/event-stream` qualified. That header comes from a tenant's own app and the flag it sets removes the request from the operator's latency histograms. Now compares the media-type essence, with tests for parameters, case, whitespace, prefix (`text/event-stream-x`) and the smuggled-parameter case. The request-side `Accept` check deliberately stays a substring match — `Accept` is a comma-separated list, so substring is correct there.

**Verification:** `cargo test --lib -p temps-proxy` → **454 passed, 0 failed**. Clippy `--all-targets -D warnings` clean. Both new regression tests verified to fail against the un-fixed code.

PR #648, opened 2026-08-13
fix(database): replace testcontainers flat-sleep with real postgres readiness wait

dviejokfs:
## Summary
Root-causes the intermittent "connection refused"/"connection reset" integration test failures reported in https://github.com/gotempsh/temps/actions/runs/31678279497/job/94381856908?pr=647 (`test_restore_postgres_from_url`, `docker-deployments` job).

- `TestDatabase`'s two TimescaleDB container startup paths (`SharedContainer::new`, `TestDatabase::new_isolated_with_config`) called `.start()` on the `timescale/timescaledb-ha` container and then just did a flat `tokio::time::sleep(5s)` before connecting — **no `testcontainers::WaitFor` readiness strategy at all**.
- The postgres/timescaledb entrypoint runs a temporary server for `initdb` + init scripts, shuts it down, then starts the real server — `"database system is ready to accept connections"` is logged **twice**, with a shutdown/restart cycle in between. A fixed 5s sleep is a guess: fine on an idle box, but under CI load (this repo runs many Docker-based integration tests concurrently by convention) the gap between the two server starts can exceed it, and a connection attempt lands mid-restart — TCP RST/refused, not a real bug in the code under test.
- This isn't a testcontainers cleanup problem (no evidence of port/container leakage between tests) — it's a missing readiness check.

## Fix
- Added a `WaitFor` that waits for the readiness log line on **stderr, twice** (once per server start) before `.start()` returns.
- Raised the container `startup_timeout` from testcontainers' 60s default to 120s for headroom under CI/dev-box contention.
- Removed the now-redundant flat sleep; kept the existing `connect_with_retry` loop as a safety net for the brief window between the log line and the TCP listener actually accepting.

## Verification
- `cargo check --lib -p temps-database` — clean.
- Pre-commit `cargo fmt` + `cargo clippy` — passed.
- Manually booted a standalone `timescale/timescaledb-ha:pg18` container and confirmed via `docker logs` that `"database system is ready to accept connections"` appears **exactly twice, both on stderr** — confirming the wait condition is correct.
- Could not get a clean end-to-end local repro of the fixed test: this dev machine currently runs ~26 unrelated Docker containers with continuous health-check traffic, and `docker events` showed no container-create activity during a local test attempt that itself timed out — i.e. the local box is Docker-daemon-contended right now, not representative. Deferring to CI (a dedicated runner, and the actual environment the original flake was observed in) to confirm the fix.

## Follow-up (not in this PR)
The same flat-sleep-instead-of-readiness-wait pattern is duplicated in several other files that hand-roll their own container bring-up rather than using this shared helper: `temps-analytics/src/testing/test_utils.rs`, `temps-dns/tests/dns_registry_test.rs`, `temps-migrations/tests/{migration_tests,normalized_managed_domain_index_test,sandbox_workspace_lifecycle_test}.rs`, `temps-network/tests/it_allocator.rs`, `temps-providers/tests/role_reconciler_integration.rs`. Out of scope here since they don't go through `TestDatabase`, but likely worth the same fix in a follow-up.

## Test plan
- [x] `cargo check --lib -p temps-database`
- [x] Pre-commit hooks (fmt, clippy, conventional commit)
- [ ] CI `docker-deployments` job — watch for `test_restore_postgres_from_url` and confirm no regression / flake resolved over a few reruns

### DIFF
--- a/crates/temps-database/src/test_utils.rs
+++ b/crates/temps-database/src/test_utils.rs
@@ -56,26 +56,20 @@ use sea_orm::*;
 use sea_orm_migration::MigratorTrait;
 use std::sync::Arc;
 use temps_migrations::Migrator;
-use testcontainers::{
-    core::{wait::LogWaitStrategy, WaitFor},
-    runners::AsyncRunner,
-    ContainerAsync, GenericImage, ImageExt,
-};
+use testcontainers::{core::WaitFor, runners::AsyncRunner, ContainerAsync, GenericImage, ImageExt};
 use tokio::sync::{Mutex, OnceCell};
 
 /// The postgres/timescaledb entrypoint starts a temporary server to run
 /// initdb + init scripts, shuts it down, then starts the real server --
 /// "database system is ready to accept connections" is logged once for
-/// each. Waiting for only the first occurrence (or worse, a flat sleep)
-/// races that restart: under CI load the gap between the two servers can
-/// exceed a fixed delay, and a connection attempt during the restart
-/// window gets ECONNRESET/ECONNREFUSED instead of a slow-but-working
-/// connection. Waiting for the second occurrence is the only readiness
-/// signal that's actually correct.
+/// each. This waits for the first (temporary-server) occurrence -- same
+/// wait condition already proven reliable elsewhere in this workspace
+/// (temps-providers/src/pg_stat_statements.rs,
+/// temps-metrics/tests/postgres_checkpoint_stats_integration.rs) -- and
+/// callers add a short buffer sleep afterward to clear the temp-server
+/// shutdown/real-server restart window before connecting.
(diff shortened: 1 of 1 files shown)


dviejokfs:
Correction: the first version of this fix (waiting for the postgres readiness log line twice, once per internal server restart) timed out in CI on the exact test it was meant to fix — a real bug, not just flakiness. Pushed a revised version that falls back to the wait condition + buffer-sleep pattern already proven working elsewhere in this workspace (`pg_stat_statements.rs`, `postgres_checkpoint_stats_integration.rs`): wait for the first readiness line, then a short sleep to clear the server-restart window, then connect-with-retry as a final safety net. Watching the next CI run to confirm before calling this done.

PR #668, opened 2026-08-15
fix(projects): stop mislabeling git connection errors as project-not-found

dviejokfs:
## Summary
- Connecting a repository could fail with a Problem Details response titled **"Project Not Found"** even though the project existed — the real issue was that the supplied `git_provider_connection_id` either didn't exist or belonged to a different user, and both cases were incorrectly folded into `ProjectError::NotFound`/`ProjectError::Other`, which the handler generically maps to "Project Not Found".
- Adds a dedicated `ProjectError::GitProviderConnectionNotFound { connection_id }` variant, uses it in both `update_git_settings` (missing-row + cross-user-ownership checks) and `update_project_settings` (missing-row check), and maps it to its own "Git Provider Connection Not Found" Problem Details title/detail.
- Updated the existing IDOR-guard test (`test_update_git_settings_rejects_connection_owned_by_another_user`) to assert the new, correctly-labeled variant instead of the old generic `NotFound`.

## Test plan
- [x] `cargo check --lib` (whole workspace) — 0 errors
- [x] `cargo test --lib -p temps-projects` — 108 passed
- [x] `cargo fmt` / pre-commit hooks clean

### DIFF
--- a/crates/temps-projects/src/handlers/handlers.rs
+++ b/crates/temps-projects/src/handlers/handlers.rs
@@ -1226,7 +1226,6 @@ pub async fn update_git_settings(
         .project_service
         .update_git_settings(
             project_id,
-            auth.user_id(),
             settings.git_provider_connection_id,
             settings.main_branch.clone(),
             settings.repo_owner.clone(),
--- a/crates/temps-projects/src/handlers/types.rs
+++ b/crates/temps-projects/src/handlers/types.rs
@@ -946,6 +946,14 @@ impl From<ProjectError> for Problem {
                     "The requested project could not be found: {}",
                     reason
                 )),
+            ProjectError::GitProviderConnectionNotFound { connection_id } => {
+                problemdetails::new(StatusCode::NOT_FOUND)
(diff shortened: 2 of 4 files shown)


dviejokfs:
Added a second commit addressing the root cause more directly: the `connection.user_id != Some(caller_user_id)` ownership check in `update_git_settings` was itself wrong, not just mislabeled. Git provider connections are scoped to the installation (workspace-wide) — a GitHub App installation is inherently shared, and PAT connections are meant to be usable by any project maintainer, not gated to whoever created them. That check would have broken GitHub App installations outright (created with a sentinel `user_id`) and blocked legitimate workspace-wide reuse of PAT connections.

Removed the ownership check and the now-unused `caller_user_id` parameter. Access to the *project* itself is already enforced by `permission_guard!`/`project_scope_guard!` in the handler — that's the correct authorization boundary, not connection ownership. Replaced the old IDOR-style tests with one asserting the opposite invariant: a connection created by a different user must still be usable.

`cargo check --lib` (workspace) and `cargo test --lib -p temps-projects` (107 tests) both clean.

PR #676, opened 2026-08-15
fix(security): harden critical trust boundaries

dviejokfs:
## Summary

Bundles the security-hardening work that can be safely applied to current `main` from 39 of bherila's security PRs. The branch was rebased onto current upstream and reviewed as one integrated change rather than merging the fork PRs independently.

### Critical fixes

- Parameterizes MariaDB data-browser filters and disables unsafe raw PostgreSQL filters.
- Restricts managed PostgreSQL images, applies the same allowlist and authoritative
  service-image checks to major upgrades and resumptions, and verifies local image
  identity fail-closed.
- Restores fail-closed project permission fallback and enforces Blob/KV project and token scope.
- Scopes internal proxy traffic to the caller's project using event-driven invalidation plus authoritative Docker ownership checks with per-project quotas, hard global concurrency, and fail-closed timeouts.
- Validates WebSocket origins, confines Docker secret paths, and safely quotes WAL-G environment values.
- Blocks unsafe webhook and custom AI endpoints with external-only resolution and redirects disabled.

### High-severity fixes

- Tightens credential rotation, AI tools, repository access, email sending, deployment-token, template, and setup-token authorization.
- Hardens node placement, registry credential forwarding, HTTPS host validation, preview subdomains, auto-deploy opt-out, CAS assets, teardown, and container inventory.
- Replaces predictable MinIO credentials, pins helper images by immutable manifest digest, removes credentialed helper build residue, binds DNS changes to verified provider zones, opts preview environments out of production secrets, and checksum-pins Bun installation.

### Source PR coverage

Critical: #48, #50, #53, #62, #70, #79, #82, #83, #85, #86, #92, #95, #96.

High: #47, #51, #52, #54, #55, #56, #57, #59, #60, #61, #63, #64, #65, #66, #67, #68, #69, #71, #72, #76, #84, #87, #88, #91, #93, #94.

#87 contributes only the owner-binding portion that is compatible with current main.

Not ported:

- #49 and #58 are superseded by #53 and #54.
- #73, #74, #75, #77, and #78 already have stronger equivalents on current main.
- #80 is incomplete when applied endpoint-by-endpoint; analytics token authorization needs a separate centralized fix.
- #81 is stale/unsafe against the current authorization model.
- #89 rejects valid Next.js runtime layouts and needs a compatibility-aware design.
- #90 can cross-wire existing Redis services without a durable allocation migration.

## Compatibility and migration notes

- PostgreSQL raw filter strings are rejected; callers must use structured filters.
[...]

PR #677, opened 2026-08-15
fix(security): harden trust boundaries with CI forward fix

bherila:
## Summary

This replacement PR carries forward the security-hardening bundle from #676 and includes the forward fixes required by the current upstream test suite.

## Forward fixes

- Updates PostgreSQL upgrade fixtures to the supported managed images: `gotempsh/postgres-walg:17-bookworm` and `gotempsh/postgres-walg:18-bookworm`.
- Makes WAL-G environment probing work when no live service container exists.
- Restores the multinode enrollment E2E after the HTTPS fail-closed hardening: the control plane now serves a private TLS listener, the worker trusts a generated one-day test certificate, and enrollment/replay checks use `https://control-plane.temps.test`.
- The multinode healthcheck verifies the certificate chain, SNI hostname, and TLS handshake before worker enrollment begins.

The original #676 security-hardening commits remain attributed to their original contributor. The focused CI compatibility commits are layered on top so the complete branch can be reviewed and merged together.

## Evidence

**Shell and Compose contracts**

```bash
bash -n tools/dev-cluster/role-control-plane.sh tools/dev-cluster/role-worker.sh
shellcheck tools/dev-cluster/role-control-plane.sh tools/dev-cluster/role-worker.sh
docker compose -f tools/dev-cluster/docker-compose.yml -p temps-dev config --quiet
docker compose -f tools/e2e-multinode-cluster/docker-compose.yml -p temps-e2e-mn config --quiet
git diff --check
```

All commands exited 0 with no diagnostics.

**Scenario TypeScript**

```bash
cd apps/temps-e2e
bun install --frozen-lockfile
bun run typecheck
```

```text
$ tsc --noEmit
```

Exit status: 0.

**Generated certificate trust and hostname verification**

An ephemeral container generated the same ECDSA certificate as the topology, installed its public certificate into the Debian trust store, started an OpenSSL TLS server, and verified the connection with:

```bash
echo | openssl s_client \
  -connect 127.0.0.1:443 \
  -servername control-plane.temps.test \
  -CAfile /etc/ssl/certs/ca-certificates.crt \
  -verify_hostname control-plane.temps.test \
  -verify_return_error
```

```text
HTTPS_TRUST_OK
```

The dedicated DinD image also built successfully with OpenSSL installed.

**Full multinode runtime**

The local ARM Docker VM exhausted memory (`rustc` received SIGKILL while compiling `temps-entities`) before the control plane could start. This path builds the full workspace inside Docker. GitHub CI uses the already-built Linux artifact, so the dedicated multinode job on this PR is the required full registration/deployment runtime proof.

## Review

- Security review: approved; the plaintext staging key lives only in `/run`, is mode `0600`, is never shared with the worker, and is removed after encrypted database import.
- Test review: approved; worker registration and consumed-token replay now both exercise the trusted HTTPS path.
- Public product documentation is unchanged because this is an isolated CI/dev topology. Inline topology and scenario documentation has been updated.


bherila:
Investigated the \`Scenario E2E / Scenario (multinode mTLS)\` failure on this run — looks like DinD networking flakiness rather than a regression from this branch.

Control-plane logs show a completely clean boot: setup completes, migrations apply, and \`temps serve\` starts listening on \`0.0.0.0:80\` within ~13s of container start, with no errors or panics afterward (Console API, proxy server, all subsystems report ready).

\`worker-1\` never gets past a bare TCP \`Connection refused (os error 111)\` trying to reach \`10.52.0.10:80\` (the control-plane's own address), across 90 retries × several full retry cycles, for the entire ~10 minute test. Since the control-plane is provably listening on all interfaces the whole time, this is a network-level failure between the two outer DinD containers, not an application issue — I checked the full diff for anything touching server bind addresses or internal-proxy routing to \`/api/internal/nodes/register\` and found nothing that would explain a network-layer refusal.

I don't have admin rights on this repo to re-run just this job. Could someone with access kick off a re-run of this check when convenient?

bherila:
Traced the multinode mTLS failure — not DinD flakiness as I first guessed (thanks for the correction, David). Root cause: `crates/temps-proxy/src/proxy.rs`'s new `production_https` default (no `external_url` configured -> assume production, force HTTPS) wasn't scoped to resolved project traffic. It also caught requests whose `Host` never matched a project domain at all — including `/api/internal/nodes/register`, which by design must work over plaintext HTTP before any TLS material exists. Since the multinode cluster never configured `external_url`, every join attempt got the "production" redirect applied against a listener with no TLS configured.

Pushed a fix (`02a2d8ce`, now merged with your harness fix as `8d5ee798`): extracted the gating into `should_apply_production_https_default()` and require `ctx.environment.is_some()` — i.e. only apply the production-HTTPS assumption to traffic that actually resolved to a project/environment, matching the feature's stated intent (CDN-terminated environments), not admin/console/internal-cluster traffic. Added unit tests covering the unresolved-host case plus the existing gates (global disable, TLS already active, ACME challenge, per-env override).

This also fixes a more general case beyond this test: any fresh self-hosted install that hasn't configured `external_url` yet would hit the same redirect-to-nowhere on its own admin/internal traffic — not just this E2E scenario.

Complementary to your harness change (real `external_url`/TLS cert + `TEMPS_DISABLE_HTTPS_REDIRECT` for the scenario) — both landed together with no conflicts.

bherila:
Update: the mTLS registration fix worked — this run got much further than before (worker joined, negotiated mTLS, project created, deployment kicked off) before failing at a new, later step:

\`worker application to GET http://production.e2e-msuwub0r-opnng5-mn.temps.local did not converge within 60s (last: {"code":1,"stdout":"","stderr":"wget: server returned error: HTTP/1.1 403 Forbidden\n"})\`

This is a different area than the redirect logic we just fixed — `TEMPS_DISABLE_HTTPS_REDIRECT=true` is set for this cluster, so the redirect path is globally short-circuited regardless of my change, and this 403 is coming from somewhere else in the request path (possibly IP access control or attack-mode challenge gating, still unconfirmed). Since the scenario never got this far before, this looks like a previously-latent issue only now exposed. Haven't dug further since I know you're actively iterating on this scenario — let me know if you want me to chase this one down too or if you've already got it.

bherila:
Update on the latest run (after `70f59fe6`, your control-plane probe placement fix): got further again — past the previous 403, through Postgres HA provisioning and DNS convergence — but now times out at a new step: **"wait for the DNS/database probe deployment"** (deployment 3, `wait for terminal state` times out at 300s, last state `running`).

Dug into the `temps-multinode-logs` artifact for this run:

- The control plane's `deploy_container` job for deployment 3 never logs a completion line. The two earlier deployments to the same worker (`whoami`, the DNS-client nginx) both logged `✅ Job 'deploy_container' completed with status: success` within 3-6s. This one just starts (`🏃 Starting job: Deploy Image`) and nothing after.
- The final `docker exec worker-1 docker ps` snapshot (~7 min after the container was created) shows every other container `Up 6-7 minutes`, but `e2e-msv7frnj-crboue-dns-probe-1` shows only `Up 32 seconds` — it's crash-looping.
- The probe app (`apps/temps-e2e/src/lib/probe-app.ts`) calls `log.Fatal` on any Postgres open/connect/create-table error, which would explain a restart loop if it can't reach the DB via the rewritten `POSTGRES_DNS_ADDRESS` host from the worker node.
- Unfortunately the CI artifact only captures the outer compose services' logs + `docker ps`/dockerd snapshots, not the deployed workload container's own stdout, so I can't see the actual Go-side error. Worth adding a `docker exec worker-1 docker logs <container>` capture step for deployed workload containers to the CI failure-log collection so the next failure shows the real crash reason.

This looks unrelated to the mTLS-redirect fix that landed — it's downstream, in the new DNS-linked-service probe step. Not chasing further since you're actively building this path; flagging in case the crash-loop symptom + missing log capture saves you a cycle.

PR #681, opened 2026-08-15
fix(providers): avoid helper image for postgres WAL probe

bherila:
### Motivation
- A prior change probed for `/var/lib/postgresql/walg.env` by pulling and running the mutable `busybox:latest` image with the Postgres data volume mounted, expanding the trust boundary and exposing WAL-G/AWS credentials to third-party image code.
- The probe must be fail-closed but avoid executing untrusted images against the Postgres volume, so we read the file via Docker's archive API instead of starting a helper container.

### Description
- Replace `compute_desired_enable_archiving()` to call a new `walg_env_exists_in_container()` helper that inspects the existing service container filesystem.
- Implement `walg_env_exists_in_container()` using Bollard's `download_from_container` archive API to read `/var/lib/postgresql/walg.env` instead of pulling/starting `busybox:latest` and mounting the volume.
- Remove the helper-container flow (image pull, create/start/wait/remove probe container) so no mutable third-party image is given read access to the data volume.

### Testing
- `cargo fmt --check`
- `cargo test -p temps-providers` (196 non-Docker tests passed; Docker-dependent tests require a local Docker socket)


dviejokfs:
Reviewed. **Requesting changes** before merge — the security motivation is good (drops a mutable \`busybox:latest\` helper container with read access to the Postgres data volume), but there's a functional regression in exactly the scenario this code's own docstring calls out as critical.

**Blocking:** \`start()\` calls \`compute_desired_enable_archiving()\` (postgres.rs:3087) before checking whether the container exists (\`need_create\` is computed after, at postgres.rs:3102). The new \`walg_env_exists_in_container()\` requires the named container to already exist, since Docker's archive-download API operates on a container, not a volume directly. In any recreate-from-scratch path — \`force_recreate()\`, node failover, or a container removed externally while the data volume persists — the probe now fails closed and \`create_container(..., desired_enable_archiving)\` bakes \`archive_mode=off\`, even though \`walg.env\` still exists on the volume. That directly contradicts the function's own docstring (postgres.rs:1063-1071, 3079-3086): "survives container recreates, Temps restarts, and node failovers." The old volume-mount probe didn't have this dependency on a live container.

Suggested fix: probe the volume directly rather than requiring a live container — e.g. run the archive-download step against a throwaway minimal container that just mounts the named volume (still avoids the previous helper-image/pull-untrusted-image issue), or explicitly special-case "container doesn't exist yet" to fall back to a volume-level check instead of assuming \`walg.env\` is absent.

Also flagging: no test added for \`walg_env_exists_in_container()\` or for the pre-container-creation path specifically — that's the case that's broken, and it'd be good to have a regression test pinning the intended behavior once fixed.

Happy to re-review once the container-existence dependency is addressed.

bherila:
Pushed a fix for the blocking issue: `compute_desired_enable_archiving()` now checks for the container first via `walg_env_exists_in_container()`, and falls back to `walg_env_exists_on_volume()` — a throwaway, never-started probe container that mounts the data volume read-only and reuses this service's own already-vetted image (no untrusted third-party image involved) — when the named container doesn't exist yet. That covers `force_recreate()`, failover, and "container removed externally but volume persists".

Added `test_walg_env_probed_from_volume_when_container_missing` as the regression test you flagged as missing, pinning the pre-container-creation path.

Also fixed a `cargo fmt` failure on the latest push (trivial rustfmt collapse, unrelated to the logic change).

Ready for re-review.

dviejokfs:
This fix already landed on `main` via f7ab93014e15b008877c021e402f2686d3266e7a (same patch — `fix(providers): avoid helper image for postgres WAL probe`, credited to you). After resolving the merge conflict here, this branch is now byte-identical to `main` on `crates/temps-providers/src/externalsvc/postgres.rs`, so merging would be a no-op. Closing as superseded — thanks for the fix!

PR #682, opened 2026-08-15
fix(providers): bind managed services to loopback

bherila:
### Motivation
- Revert Docker port publishing for managed external services from wildcard `0.0.0.0` to loopback `127.0.0.1` to prevent accidental exposure of databases and object storage on all host interfaces.
- The change restores the previous security posture where sensitive backing services (Postgres, Redis, MongoDB, S3/MinIO, RustFS) are reachable only from the host/loopback and the internal proxy/networking paths.

### Description
- Updated Docker `PortBinding.host_ip` to `Some("127.0.0.1".to_string())` in `crates/temps-providers/src/externalsvc/postgres.rs`, `redis.rs`, `mongodb.rs`, `s3.rs`, and `rustfs.rs` so published host ports bind to loopback.
- Kept existing container host port selection and other host_config fields unchanged to preserve functionality and persistence behavior.

### Testing
- `cargo fmt`
- Verified no remaining wildcard publishes by searching for `host_ip: Some("0.0.0.0".to_string())` under the affected provider sources.

### DIFF
--- a/crates/temps-providers/src/externalsvc/mongodb.rs
+++ b/crates/temps-providers/src/externalsvc/mongodb.rs
@@ -472,7 +472,13 @@ impl MongodbService {
         }
 
         let mut host_config = bollard::models::HostConfig {
-            port_bindings: Some(crate::utils::local_port_binding("27017/tcp", &config.port)),
+            port_bindings: Some(HashMap::from([(
+                "27017/tcp".to_string(),
+                Some(vec![bollard::models::PortBinding {
+                    host_ip: Some("127.0.0.1".to_string()),
+                    host_port: Some(config.port.clone()),
+                }]),
+            )])),
             mounts: Some(vec![bollard::models::Mount {
                 target: Some("/data/db".to_string()),
                 source: Some(volume_name),
--- a/crates/temps-providers/src/externalsvc/postgres.rs
+++ b/crates/temps-providers/src/externalsvc/postgres.rs
@@ -581,7 +581,13 @@ impl PostgresService {
         ];
 
         let mut host_config = bollard::models::HostConfig {
-            port_bindings: Some(crate::utils::local_port_binding("5432/tcp", &config.port)),
+            port_bindings: Some(HashMap::from([(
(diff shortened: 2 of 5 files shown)


dviejokfs:
Closing — this fix already exists on \`origin/main\` via a shared helper, so this PR is a no-op that would fork the logic instead of fixing anything.

All 5 sites this PR touches already call \`crate::utils::local_port_binding()\` on main, which is the single source of truth for loopback-only binding and already references the same upstream issue (bherila/temps#29) this PR cites:

\`\`\`
crates/temps-providers/src/externalsvc/postgres.rs:584:  crate::utils::local_port_binding("5432/tcp", &config.port)
crates/temps-providers/src/externalsvc/redis.rs:444:    crate::utils::local_port_binding("6379/tcp", &config.port)
crates/temps-providers/src/externalsvc/mongodb.rs:475:  crate::utils::local_port_binding("27017/tcp", &config.port)
crates/temps-providers/src/externalsvc/s3.rs:453:      crate::utils::local_port_binding("9000/tcp", &config.port)
crates/temps-providers/src/externalsvc/rustfs.rs:784-787: crate::utils::local_port_binding(...)
\`\`\`

\`local_port_binding\` (crates/temps-providers/src/utils.rs:70) has dedicated unit tests asserting the host IP is never \`0.0.0.0\` (\`test_local_port_binding_binds_to_loopback_only\`, \`test_local_port_binding_never_binds_to_all_interfaces\`), and 3 sibling call sites — \`mariadb.rs\`, \`postgres_lifecycle.rs\`, \`services.rs\` — also already use it.

Merging this PR as-is would inline duplicate binding logic into these 5 files while the other 3 keep using the shared helper, splitting the loopback-binding policy across 4 places instead of 1 — a maintenance/security-process regression, since a future policy change (e.g. a config knob) could silently miss the duplicated copies.

If your branch predates this fix landing on main, a rebase should make this diff disappear entirely. Closing for now — please reopen if there's a reason to move away from the shared helper, in which case we'd want the other 3 call sites updated too for consistency.

PR #683, opened 2026-08-15
fix(providers): restore redis database allocation isolation

bherila:
### Motivation
- The previous change used a deterministic `hash % 16` to pick a Redis DB and made `deprovision_resource` a no-op, which allowed DB collisions and stale data exposure between projects/environments.
- The intent is to restore strong per-resource isolation for Redis logical DBs and ensure deprovisioning cleans up allocated DBs.

### Description
- Replaced the hash-based allocator with an allocation-backed flow that stores mappings in Redis DB 0 via helper keys `'_temps:redis_db_mapping:{resource}'` and `'_temps:redis_db_owner:{db}'` and added helper functions `resource_mapping_key` and `database_owner_key` in `crates/temps-providers/src/externalsvc/redis.rs`.
- Implemented `allocate_database` which reserves DBs 1–15 using `SETNX` and records the mapping, and returns an error (fail-closed) when no DBs are available.
- Implemented `drop_database` which selects the allocated DB, runs `FLUSHDB`, and removes the allocation metadata; wired `deprovision_resource` to call `drop_database`.
- Updated runtime env generation in `get_runtime_env_vars` to call `allocate_database` and derive `REDIS_DATABASE`/`REDIS_URL` from the recorded allocation (instead of `calculate_database_number`).

### Testing
- `cargo fmt`
- Verified removal of the old `hash % 16` pattern and presence of `allocate_database`/`drop_database` and `REDIS_DATABASE` usage via grep.

### DIFF
--- a/crates/temps-providers/src/externalsvc/redis.rs
+++ b/crates/temps-providers/src/externalsvc/redis.rs
@@ -595,6 +595,15 @@ impl RedisService {
     /// selected from DBs 1-15 and recorded in Redis before their connection
     /// details are returned, so two resources cannot silently receive the same
     /// logical DB. When all databases are in use, allocation fails closed.
+    ///
+    /// The "read mapping, then claim" sequence below isn't a single atomic
+    /// transaction, so a concurrent `drop_database` for the same resource
+    /// could in principle interleave between the mapping read and the
+    /// `SETNX` claim. This is an accepted race: a single Redis instance and
+    /// the short critical section make it low-risk in practice, and
+    /// allocate/drop for the same resource aren't expected to run
+    /// concurrently (provision and deprovision are serialized per resource
+    /// at the caller).
     async fn allocate_database(&self, resource_name: &str) -> Result<u8> {
         let mut conn = self.get_connection().await?;
         redis::cmd("SELECT")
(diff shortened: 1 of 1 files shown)


dviejokfs:
Reviewed. **Requesting changes** before merge — allocation logic itself looks sound, but two blockers:

1. **No tests for the new allocation/dedup/deprovision logic.** `allocate_database`/`drop_database` (redis.rs:598-749) are the whole point of this fix, yet the diff adds zero unit/integration tests exercising: two resources getting distinct DBs, idempotent re-allocation for the same resource (redis.rs:614-616 / 651-663), the "all 16 exhausted" fail-closed error, or `drop_database` actually freeing a DB for reuse. Given this is explicitly re-fixing a previously-broken isolation guarantee, please ship with a test proving collisions can't happen.
2. **Evidence is compile-only.** The PR's Testing section is `cargo fmt` / `cargo check --lib` / grep-for-string-presence — no run against a live Redis container showing two resources allocate different DBs, or that deprovision+reprovision reuses a freed slot. Per this repo's evidence bar, that's required before merge.

Also flagging (non-blocking, please confirm):
- `get_runtime_env_vars` now requires a live Redis connection (redis.rs:2077-2080), reversing the previous deliberate design of generating env vars without one. Could you confirm the call-site ordering in `env_resolver.rs:158-160` guarantees the Redis service is already up when this runs, so first-deploy ordering doesn't regress?
- The "check existing mapping, then claim" sequence in `allocate_database` (redis.rs:606-664) isn't a single atomic transaction — low risk given a single Redis instance and a short critical section, but worth a comment acknowledging the accepted race with a concurrent `drop_database`.

No security/injection issues otherwise — `resource_name` is server-constructed, not user-supplied into a query, and connection-per-call avoids `SELECT`-state leakage. Happy to re-review once tests + evidence are added.

bherila:
Added the requested test coverage: `test_allocate_database_isolation_and_reuse` (two resources get distinct DBs, re-allocating the same resource is idempotent, `drop_database` frees the slot for reuse) and `test_allocate_database_fails_closed_when_exhausted` (allocation errors once DBs 1-15 are claimed), both run against a live Redis container.

On the two non-blocking questions:
- `get_runtime_env_vars`'s new live-connection requirement is safe at the `env_resolver.rs:158-160` call site — it only runs against services already present in `project_services`, which is only populated after a service has successfully provisioned. So by the time this runs, Redis is already up; first-deploy ordering isn't affected.
- Added a doc comment on `allocate_database` acknowledging the accepted, low-risk race between the mapping read and the `SETNX` claim (single Redis instance, short critical section, allocate/drop for the same resource aren't expected to run concurrently).

Ready for re-review.

PR #702, opened 2026-08-17
fix(compose): deliver project secrets to Docker Compose stacks

dviejokfs:
## Problem

Project secrets were **silently never delivered** to Docker Compose deployments.

`gather_secrets` ran for every deployment, and then the Compose branch dropped the result on the floor:

```rust
let secrets = self.gather_secrets(project, environment).await?;

if project.preset == Preset::DockerCompose {
    return self.plan_compose_deployment(
        project, environment, deployment, env_vars, buildkit_cache_namespace,
    ).await;   // <- `secrets` not passed
}
```

`secrets` only reached `plan_docker_image_deployment` and `plan_git_deployment`. Neither `DeployComposeJob` nor `ComposeDeployRequest` had a field to carry them, so a user created a secret, the dashboard showed it existing, and `/run/secrets/<KEY>` never appeared in the container — with no error anywhere.

There was no self-service workaround either: the compose security policy rejects any `secrets.*.file` outside the repo checkout, so the only way to hand-roll it was committing the secret to git.

This is gap **G4** in ADR-039.

## Approach

Thread `secrets` through planner → job → executor (sealed in `job_config` under its own field, never merged into `environment_vars`) and mount them with a generated `docker-compose.temps-secrets.yml`, following the existing env/network/labels/security override pattern.

ADR-039 §5 suggested materializing native compose `secrets:` entries. This uses a directory bind mount instead, because:

- plaintext values stay out of the compose YAML entirely — only the host directory path appears,
- the container-side contract becomes byte-identical to the single-container path (`/run/secrets/<KEY>`), which is what the UI already documents,
- native `secrets:` needs every key enumerated per service *and* still needs a host `file:` path, so it is strictly more moving parts for the same result.

The override is written **last** in the `-f` order, so a repository or user override cannot redirect where secrets land.

### Where the files live

`$TEMPS_DATA_DIR/compose-secrets/<project>/` rather than the repo checkout, which is deleted as soon as the deploy job finishes — a mount source inside it would be gone by the first container restart.

Permissions are `0700` root / `0755` stack dir / `0444` files. The `0700` root is the host-side boundary; the stack directory has to stay traversable by whatever uid each image runs as. The single-container path chowns `0400` files to the image's `USER`, which is not available here: a stack has many images and Compose pulls several of them during `up`, after this file has to exist. World-readable *inside* the container is not a boundary worth defending — any process there already runs as the app.

The directory is recreated on every deploy (so a deleted or renamed key cannot survive as a stale file) and removed on teardown, on destroy, and on the failed-deploy compensation path.

### Conflicts

Services that already mount `/run/secrets` or declare a compose `secrets:` key are **skipped**, not mounted over: Compose appends volume lists across `-f` files, so a duplicate target would fail `up` for the whole stack. Losing delivery for one service is recoverable and reported in the deployment log; failing an otherwise-valid stack is not.

## Evidence

Real `docker compose` deploy through `ComposeExecutor::deploy`, alpine running as uid 1000 under the usual `cap_drop: ALL` sandbox:

```
--- generated override ---
services:
  app:
    volumes:
    - /Users/.../compose-secrets/temps-evidence-1:/run/secrets:ro

--- IN-CONTAINER ---
uid=1000 gid=1000 groups=1000
total 4
-r--r--r--    1 1000     1000            18 Aug 17 13:16 DB_PASSWORD
hunter2-from-temps

SECRET IN docker inspect: false
secrets dir after destroy exists: false
```

Secret present and readable by a non-root container user, correct content, `0444` via a read-only mount, absent from `docker inspect`, cleaned up on destroy.

**Trap this surfaced:** with the data directory on a path Docker Desktop does not share (a macOS `/var/folders/...` temp dir), the mount appears **empty** inside the container rather than failing. The existing single-container secrets path has the same exposure. Not addressed here; worth a follow-up that verifies delivery post-`up` instead of assuming it.

### Tests

`cargo test --lib -p temps-deployer`: 111/112 compose tests pass. The one failure (`test_security_override_runtime_matches_per_service_selection`, `404 No such container`) reproduces identically on unmodified `origin/main` — pre-existing, local Docker environment, unrelated.

`cargo test --lib -p temps-deployments`: 619/619. Workspace `cargo check`, clippy `-D warnings`, and fmt are clean.

New coverage: materialization + mount shape, host-side permissions, stale-file removal on key deletion and rename, key rotation, invalid-key rejection, conflict detection and skipping, `compose_path` collision with generated overrides, redaction of a secret and env var sharing a key, and a test pinning that an inline override cannot introduce services.

## Security review

Reviewed by the `security-auditor` gate. Verified clean: path traversal via secret keys (validated per iteration; the project-name path component is `temps-{i32}-{i32}`), YAML injection (the override is built through `serde_yaml`'s typed API, not string templates), `-f` ordering, sealed `job_config` storage, non-exposure in `docker inspect`, and lifecycle on teardown/destroy/failed deploy.

One MEDIUM was raised — override-only services silently missing secrets — which I investigated rather than accepted: `validate_compose_override` already refuses to let an inline override add services, so it is not reachable. The defensive union is kept anyway, the log now names the actual services instead of claiming "every service", and a test pins that dependency so the guarantee cannot be removed elsewhere without a failure here.

Also fixed from the review:
- the defense-in-depth filename check now matches the API's `[A-Za-z0-9_]` charset; it runs on a value read back from the database, so it should not assume the API was the only writer,
- a `:` in `TEMPS_DATA_DIR` is now a clear error rather than a silently-wrong mount (Docker treats it as a bind-spec field separator).

## Adjacent gaps closed

- **Secret values are redacted from failure diagnostics.** A crash-looping container that echoes its own secret would otherwise have it captured into the persisted deployment error. Redaction now carries values rather than a map, so a secret and an env var sharing a key no longer collapse into one entry and leak the other. (ADR-039 has no ID for this; it appears with file delivery.)
- **G5** — the `DeployRequest.secrets` doc, the API schema description, both generated clients, the UI copy and the migration all claimed tmpfs delivery. It is a persistent host directory, deliberately: tmpfs would empty on host reboot and every container would come back with no secrets until a redeploy. The stale claim told the reader plaintext never touches disk when it does.
- **G7** — `binds` is collected rather than assigned, so adding a second bind cannot silently drop the secrets mount.
- `compose_path` naming one of Temps' generated overrides is rejected; it would otherwise be overwritten with generated content and then passed to `-f` twice.

## Not in scope

The rest of ADR-039 — the `_FILE` convention, the `delivery` column, linked-service credential routing, resolver delivery (G1/G2/G3), and CLI parity for project secrets (G6). This is G4 plus the two one-line gaps in the same files. No new API endpoint, so no CLI parity is triggered.

The regenerated `openapi.json` diff is 1 line; the CLI client was regenerated from it with `bun run generate:api`.


dviejokfs:
## Added: per-service secret scoping

Closes the least-privilege gap this PR originally left open — every service in a stack received every secret, so a `postgres` container could read an application's `GITHUB_APP_PRIVATE_KEY`. Within one project that's blast radius rather than a cross-tenant leak, but there's no reason for it.

A secret can now name the Compose services allowed to read it. The executor writes **one directory per service** and mounts each service its own, so scoping is enforced by *what was written*, not by the container: a service outside a secret's scope has no filesystem path to the value at all.

That per-service split is also why the native compose `secrets:` key wasn't usable here. Tested against a real daemon, it's a plain bind mount whose `uid`/`gid`/`mode` are silently ignored outside swarm:

```
--- requested uid 1000, gid 1000, mode 0400 ---
-rw-------  1 0  0  11  db_password
"Type":"bind", "Destination":"/run/secrets/db_password", "RW":false
```

One shared root-owned `0600` file per secret, unreadable by a non-root container on Linux and impossible to fix per-service.

### Opt-in, and it stays opt-in

A secret with no scope rows goes to every service — today's behaviour — and the migration backfills nothing, so no running stack loses a secret on upgrade. "Unconfigured" must never mean "withheld".

### Runtime evidence

Two services, `STRIPE_KEY` scoped to `web`, `SHARED_TOKEN` unscoped:

```
=== web ===        === db ===
SHARED_TOKEN       SHARED_TOKEN
STRIPE_KEY         --
--                 cat: can't open '/run/secrets/STRIPE_KEY': No such file or directory
sk_live_scoped_to_web
```

API, against a local server (which also confirms the migration applies cleanly):

```
create scoped     -> 201   compose_services = ['web']
create unscoped   -> 201   compose_services = []
traversal name    -> 400   Invalid compose service name '../../etc': invalid character '/'
separator in name -> 400   Invalid compose service name 'a/b': invalid character '/'
widen scope       -> 200   ['web', 'worker']
clear scope       -> 200   []  (empty = all services)
```

### Rename drift

Service names live in the user's repository, so a rename would strand a scoped secret with nothing saying why. Scopes naming a service that doesn't exist in the deployed stack are reported per-secret in the deployment log, and the log now prints the delivery matrix per service using the *same predicate the executor applies when writing files* — a second copy of that rule would let the log lie.

Narrowing a scope deletes the revoked service's copy on the next deploy; the tree is rebuilt each time, so revocation actually removes the plaintext.

### Surfaces

`compose_services` on create/update/list, regenerated CLI + web clients, `--service` / `--all-services` on `temps projects secrets`, and a service picker in the dashboard that onboards rather than disappearing when a Compose project hasn't synced its service list yet. The CLI echoes the current scope back on update, since the API replaces it wholesale and omitting it would silently widen a scoped secret.

### Tests

New: scoped-vs-unscoped write placement, no-mount-at-all for an unentitled service, scope narrowing removing the prior copy, empty scope meaning *all* (both `{}` and `{K: []}`), rename-drift detection, and service-name path-component rejection. Suites: temps-deployer 117/118 compose, temps-deployments 619, temps-environments 76, temps-entities 67, temps-migrations 7. The one deployer failure is the pre-existing `test_security_override_runtime_matches_per_service_selection`, which fails identically on unmodified `origin/main`.

---

## Also included: unrelated pre-existing test failure on main

`test_stats_served_from_cagg_match_raw_path` fails on `origin/main` (verified against a clean checkout). The time-bucket queries counted Temps' own `temps_monitor` status checks — 5 requests reported for a project that received 4.

The `is_system_request` dimension was added to `proxy_logs_stats_1m` precisely to remove these; its migration says *"makes idle projects appear to receive 1,440 requests per day"*. The health-summary queries filter on it; the time-bucket queries feeding the chart directly beneath that summary never did — so the headline total excluded monitor traffic while the chart under it included it.

Fixed in `build_filter_sql` so all three time-bucket queries agree, including the latency-enrichment pass — fixing only the two counting queries would leave percentiles averaged over a different population than the counts they annotate. Separate commit (`657b92572`); happy to split it out if you'd rather land it on its own.

`test_filter_values_and_sql_consistency` asserted clause-count == value-count, which any literal predicate breaks; it now asserts placeholder-count == value-count, which is what actually catches an off-by-one binding every filter to the wrong column.

temps-proxy: 510 passed, 0 failed.

PR #709, opened 2026-08-17
fix(security): close 14 open findings from the security review

dviejokfs:
Fixes 14 of the 32 security findings that were still open on `main`, from the
Codex security-findings review. Each fix has a regression test; the
authorization and validation ones are additionally verified end-to-end against
a live local instance (evidence below).

The remaining 18 are listed at the bottom with the reason each was left out —
several are accepted-by-design, and one I attempted and reverted because the
fix would have broken managed DNS for the primary provider.

### Critical

**`env_file` could create a trusted Compose override** (`temps-deployer`)
A tenant compose file could name one of the Compose files Temps generates
(`docker-compose.temps-{env,network,override,labels,security}.yml`) as an
`env_file:` target. When the repository didn't ship that file, Temps
synthesised one from the project's environment variables and wrote it to that
exact path — and `compose_up` passes every generated file it finds to
`docker compose -f` purely because it exists, with only the *user-supplied*
override run through `validate_compose_override`. Dotenv and YAML syntax
overlap enough (`#` comments, bare `key: value`) to smuggle a whole `services:`
mapping past the dotenv reader, so this was a deployment-sandbox escape: an
extra privileged or host-mounted container. `compose_path` is user-selected
too and gets the same guard. Rejected outright rather than skipped, so the
operator gets an actionable error instead of a stack that fails later for an
unrelated reason.

### Host-level escapes

**MariaDB PITR restore trusted S3 binlog filenames** — `manifest.shipped_files`
was read back from S3 and each entry written to `dest_dir.join(&file)`.
`PathBuf::join` doesn't confine an absolute path and leaves `..` unresolved, so
`/tmp/payload` or `../../etc/cron.d/x` was an arbitrary host file write; the
same string is interpolated into the S3 key, where `..` reads outside the
service's prefix. Validated at the manifest-read boundary (rejecting the whole
manifest, since replaying a filtered subset would report a successful PITR that
silently skipped segments) and again at the write site.

**Sandbox root exec was PATH-hijackable** — `normalize_ownership` and the
AI-CLI restore step ran `chown`, `su`, `cp` and `curl` as container root
without a pinned PATH. The image puts `~/.local/bin` and `~/.bun/bin` — both
writable by the sandbox user — ahead of the system directories, so a sandbox
user could drop their own `chown` and have it run as root on the next recovery
or restart. PATH is now pinned to system directories for every root exec.

### Authorization

- **OTel facets** are platform-global (20 shared slots on a shared spans table;
  create/delete triggers `ALTER TABLE ... UPDATE` across every tenant's spans)
  but were gated on `OtelWrite`, which `Role::User` holds. Any tenant could
  exhaust the slots, delete other tenants' facets, and repeatedly trigger
  table-wide mutations. Now requires an instance administrator.
- **Feature flag writes** used `project_access_guard!` ("can this user reach the
  project at all"), so a user with instance-wide `FlagsWrite` but read-only
  membership on a project could flip that project's flags and kill switches.
  Now uses `project_permission_guard!`.
- **Sandbox terminal** accepts `?cmd=` and runs it via `/bin/sh -c`, but
  required `SandboxesWrite` — a token deliberately granted `sandboxes:write`
  without `sandboxes:exec` got code execution anyway. Now requires
  `SandboxesExec`, like every other exec route.
- **Deployment metrics** bound only deployment *tokens* to their project;
  session and API-key callers with `DeploymentsRead`/`Write` could read or
  toggle monitoring on any deployment id. The session path now runs through the
  same `ProjectAccessChecker` the external-service path uses, still returning
  404 so other tenants' deployments aren't confirmed to exist.
- **Weekly digest** aggregates events, error groups and funnels across all
  projects with only a time-window predicate and sends the result to configured
  providers. `NotificationPreferencesWrite` and `NotificationProvidersCreate`
  are both in the default `Role::User`, so a tenant could point a provider at
  themselves and trigger an instance-wide dump. Restricted to instance admins.
- **Metric alert CRUD / anomaly preview** and **`GET /emails/events`** gained
  the missing `project_scope_guard!` / `deny_deployment_token!`. These are
  defence-in-depth rather than live holes — `AuthContext::has_permission`
  already refuses to map any deployment-token permission onto `OtelRead`/
  `EmailsRead` — and the comments say so rather than overstating the risk.

### Resource exhaustion

- **Native frame suffix matching** split the frame filename on `/` and issued
  one `LIKE` query per trailing suffix with no depth cap. The filename comes
  verbatim off a submitted Sentry event on a *public* ingest route (DSNs are
  browser-observable) accepting 2 MiB, so `"a/a/a/.../missing.go"` became
  hundreds of thousands of joins and database round trips. Capped at the 16
  deepest suffixes.
- **Remote log lines** were read with `BufReader::lines()`, which has no
  maximum. The collector opens a follow stream per remote container, so an app
  that writes one very long line — or never writes a newline — made the control
  plane buffer it indefinitely (the downstream 1 MiB chunk limit only ever sees
  whole lines). Now `LinesCodec` with an explicit maximum, surfacing overflow as
  a stream error rather than silent truncation.
- **`.temps.yaml` alert rules** bypassed `validate_rule` entirely; the only cap
  was 100 *entries*, so one entry could carry thousands of `group_by` keys the
  evaluator expanded into SQL every cycle. Same limits as the API now, sharing
  its constants.
- **`metric_name`** is now restricted to `[a-zA-Z0-9_.:-]`, not just bounded in
  length. It is interpolated into the alert-summary prompt, and when the summary
  provider is an agent CLI that prompt reaches a host process — so
  instruction-shaped prose in a metric name was a prompt-injection surface.

### Infrastructure

- **Importer `source_url` SSRF** — a separate value from the already-validated
  `base_url`, taken from the source platform's API response (Coolify returns
  `external_db_url` verbatim when `is_public`) and handed to a database client
  container started with `network_mode=host`. New
  `validate_external_database_url` applies the same host/IP rules for
  connection-string schemes. Note the `url` crate only resolves
  `Host::Ipv4`/`Ipv6` for *special* schemes, so `postgres://…@127.0.0.1/` parses
  as a `Domain` — the validator parses the host string as an address itself, and
  a test covers exactly that.
- **RustFS ingest keys on remote services** — `provision_otlp_ingest_key` ran
  for every new rustfs/s3 service including remote ones, and
  `store_and_apply_ingest_key` applies the key by creating a container through
  the *control plane's* Docker client. Creating a remote service therefore also
  ran its image — attacker-selectable via the `docker_image` parameter — on the
  control-plane host. Now skipped, and said out loud.
- **Snapshot image tags** were `temps-snapshot/<sanitized-label>:latest`. Docker
  tags are mutable and shared per daemon, so any tenant snapshotting with label
  "backup" re-pointed `temps-snapshot/backup:latest` at their own image, and
  restore only checks that the tag exists — the next tenant restoring their own
  "backup" would run the attacker's image with their env vars and git
  credentials. Tags are now derived from the tarball digest.
- **`--ca-fingerprint` was ignored in relay joins** — the check only ran in
  `join_direct`, so an operator following the documented enrollment flow could
  pass the flag, join through the relay, and still silently trust a CA from a
  malicious relay or MITM'd registration endpoint. Now runs in both paths,
  before `persist_tls` writes anything.
- **Failure-report opt-out was advisory** — `TEMPS_TELEMETRY` only decided
  whether the UI offered the action; the POST path never consulted it, so on an
  instance with outbound reporting disabled anyone with `DeploymentsRead` could
  still ship deployment logs to the central endpoint. Enforced where the request
  would leave the machine, surfaced as 403 with the GitHub-issue alternative.

## Evidence

`cargo clippy --workspace --all-targets -- -D warnings` — clean.

Unit/integration tests (new tests marked ✱):

```
temps-core          348 passed   ✱ validate_external_database_url (3) + 4 doctests
temps-deployer      238 passed   ✱ reserved compose filenames (5)
temps-providers     528 passed   ✱ binlog filename validation (2)
temps-agents        340 passed   ✱ root exec PATH (1)
temps-otel          492 passed   ✱ facet admin gate (3)
temps-deployments   635 passed   ✱ YAML alert caps (4), failure-report opt-out (1)
temps-error-tracking 129 passed  ✱ native suffix cap (5)
temps-dns/-flags/-email/-notifications/-sandbox/-import/-cli  all passed
```

One pre-existing failure, unrelated and identical on the base commit
(`compose::tests::test_security_override_runtime_matches_per_service_selection`
— needs a live Docker compose stack).

End-to-end against a live instance (fresh database, slot 13), on the final tree:

```
[1] tenant creates global OTel facet          -> 403  (expect 403)
[2] tenant deletes admin's facet              -> 403  (expect 403)
[3] admin creates facet                       -> 201  (expect 201)
[4] tenant triggers instance-wide digest      -> 403  (expect 403)
[5] admin triggers digest                     -> 200  (expect 200)
[6] terminal, sandboxes:write but NOT :exec   -> 403  (expect 403, "requires sandboxes:exec")
[7] terminal, WITH sandboxes:exec             -> 404  (guard passed, sandbox absent)
[8] alert, prompt-injection metric_name       -> 400  (expect 400)
[9] alert, 5000 group_by keys                 -> 400  (expect 400)
[10] ordinary alert                           -> 201  (expect 201)
```

Full 403 bodies name the specific reason, e.g. for [1]:

> OTel facets are platform-global: creating a facet rewrites facet columns
> across every project's spans and consumes one of the shared facet slots. Only
> an instance administrator may change them.

## Not fixed in this PR

**Attempted and reverted — needs a design change first**

- *DNS sync can overwrite untagged records.* The create/update path doesn't
  apply the `temps:managed` ownership check that deletion does. I implemented
  the guard and it broke the existing
  `reconcile_never_deletes_untagged_records` test for a good reason: providers
  whose API exposes no record comment (the Cloudflare crate we use) can't carry
  the tag at all, so *every* record reads as untagged and the guard would
  silently disable generated-hostname sync for the primary provider. Doing this
  correctly needs a `record_comments` capability on `DnsProviderCapabilities`
  so the check only applies where "untagged" actually means "not ours". Left
  out rather than shipping a regression worse than the finding.

**Accepted-by-design / operator policy** (documented as intentional in code)

- Git settings accepting another user's provider connection — the code
  documents connections as shared instance-wide by design.
- `0` = unlimited for `memory_limit`, request timeouts and
  `max_concurrent_connections` — documented escape hatches; making them
  operator-only is a product decision, not a code fix.
- OTel storage quota defaulting to off (matches the accepted risk on #283).

**Real, still open — deliberately out of scope for one PR**

Each needs its own change with its own test and review:
S3 `restore_in_place` cross-service authorization; MongoDB restore cloning the
target's password into an attacker-chosen image; stale-image cleanup trusting
tenant-controlled `image_name`; Redis allocator reassigning legacy DBs after
upgrade; HTTPS default for console fallback hosts; compose service hostnames
squatting preview subdomains; source-type switch leaving Git webhooks active;
cross-project trace reads gated only on `OtelRead`; the public ADR describing
unresolved multi-node weaknesses.

I'd rather land these 14 with evidence than 32 without.


dviejokfs:
## Second batch: 8 more findings closed

Commits `507e648` → `0c13753` extend this PR beyond the original 14.

| Finding | Fix |
|---|---|
| S3 restore can overwrite or delete other services' data | `BackupAppState` now carries the `ProjectAccessChecker`; `start_restore` requires access to the target service **and** every service that produced the referenced backup. Deployment tokens are confined to their own project on both ends and refused orphan (raw `backup_location`) restores. |
| MongoDB restore clone leaks DB password to image | The new service inherits the source's root credentials via `MONGO_INITDB_ROOT_*`, so a caller-chosen `docker_image` received them. The override is now restricted to the source's own repository or a known MongoDB repository, exact-match. |
| Redis allocator can reassign stale DBs / hash collisions | `allocate_database` checks `DBSIZE` before claiming. A DB that holds data but has no owner key (pre-metadata hash scheme) is reserved under a sentinel owner instead of being handed out, so it is never allocated and `drop_database` can never `FLUSHDB` it. |
| HTTPS default bypass for console fallback hosts | An unresolved Host falls through to the console upstream, so exempting unresolved hosts left the management surface on plaintext and its cookies' `Secure` flag derived from `X-Forwarded-Proto: http`. The default now applies to the console hostname; the bootstrap exemption survives as a path rule on `/api/internal/`. |
| Compose service routes can squat preview subdomains | Service hostnames are now `{service}--{env}`. Every slug generator collapses hyphen runs, so no project/environment/deployment slug can contain `--` and the collision is structurally impossible rather than order-dependent. **Operator note: existing per-service Compose preview URLs change.** |
| Cross-project trace API leaks tenant telemetry | Both endpoints intersect results with the caller's project access; unified traces drop unreadable spans/projects and recompute the aggregates. No-op in OSS. |
| Source switch leaves Git webhooks active | Push events are filtered to Git-capable projects at the single queueing choke point. `Manual` still accepted — documented as accepting any deployment method. |
| Node CA still signs worker-controlled SANs | Node identities (name, address, private_address) are now exclusive and checked crosswise, so a join-token holder cannot register under a fresh name while claiming a victim's address. |

**Verified as already fixed on main, no change made:** *Stale image cleanup trusts tenant-controlled image refs* — `expired_image_candidates` already excludes tenant-supplied refs (`metadata->>'externalImageRef'`, `trigger = 'image_upload'`, non-git `source_type`) and uses `BOOL_AND` so any non-expired reference protects the tag; existing retention tests cover both.

**Not addressed here — needs a decision, not a patch:**
- *Public ADR discloses unresolved critical vulnerabilities* (`docs/adr/020-multi-node-*.md`). Deleting the files now does not undo the disclosure — public git history retains them — and it destroys the working document the hardening effort depends on. The options (rewrite history, move to a private repo, or accept and prioritise the underlying fixes) are yours to pick.
- *CLI skill executes unpinned transitive npm code*. The recommended fix (avoid `bunx`/`npx`, install pinned with `--ignore-scripts`) contradicts the repo's own standing instruction to always invoke the CLI as `bunx @temps-sdk/cli`, so it needs your call rather than a unilateral reversal.

Workspace clippy clean with `-D warnings`; new regression tests in each touched crate, plus a Docker-backed test for the Redis allocator.

dviejokfs:
### Follow-up: reworked the console-HTTPS change (`7698e40bb`)

Reviewing `288e6344d` before merge, I traced the full redirect path and found its net effect was the opposite of what was intended.

For the console host the decision reduces to `inherited_https_policy(production_https, host_has_cert)` = `production_https || host_has_cert`. So:

| `external_url` | Temps holds a cert | before `288e6344d` | after |
|---|---|---|---|
| unset | — | no redirect | no redirect |
| `http://…` | — | no redirect | no redirect |
| `https://…` | **yes** | redirect (via `host_has_cert`) | redirect — unchanged |
| `https://…` | **no** | no redirect | **redirect (new)** |

The last row is the only behaviour that commit changed, and "external_url is https but Temps holds no certificate for it" is the definition of **TLS terminated upstream**. In that topology the browser reaches the CDN over HTTPS, the CDN speaks plaintext to Temps, `is_tls_connection()` is correctly false (SEC-12 — `X-Forwarded-Proto` from an arbitrary peer isn't evidence), and Temps 301s back to the CDN, which forwards plaintext again: an **infinite redirect loop**, escapable only via the global `--disable-https-redirect`, which also turns the redirect off for every project.

An `https://` external_url says how users reach the platform, not who terminates the TLS, so it can't be used to infer this.

**What `7698e40bb` does instead** — `AppSettings::console_force_https`, the same tri-state an environment's `force_https` already uses:

- `None` (default) — inherit the per-host certificate heuristic. Byte-for-byte today's behaviour for every existing install.
- `Some(true)` — always redirect, for operators terminating TLS at Temps without a Temps-issued cert.
- `Some(false)` — never redirect.

Two related fixes carried over from the same commit:

- `INTERNAL_CLUSTER_PREFIX` is now exempt in `should_redirect_to_https`, not just in the production default. A worker registers over plaintext HTTP against the console host — exactly the host an operator is likeliest to force to HTTPS — so the override must not break cluster join.
- A settings read failure now yields `None` (inherit) rather than assuming the host is the console, so a transient DB error can't start 301ing to a certificate that may not exist.

Surfaced as a three-way control on the External URL settings card (which spells out the CDN caveat) and via `temps settings show` / `settings update --console-force-https auto|always|never`.

**Known gap, stated plainly:** in the CDN-fronted topology the console still mints cookies without `Secure`, because that flag comes from the `X-Forwarded-Proto` Temps sets from its own connection. A redirect never fixed that — it just broke the install. Worth a separate issue.

#### Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — clean; `cargo fmt --check` — clean
- `cargo test --lib` for temps-proxy / temps-core / temps-config — 958 passing, incl. a named regression test for the CDN loop
- `bun test` in `apps/temps-cli` — 13 passing, incl. 2 new
- `bunx tsc --noEmit` in `web/` and `apps/temps-cli` — clean; `spec:check` canonical
- Round-tripped the setting against a live server: `null → true → null`, all persisting correctly

#### On the red check

`Scenario E2E / Scenarios (observability)` is **pre-existing on main**, not caused by this PR. Same step fails on main's own runs (`32072720498`, `32060116321`):

```
"step": "GET / issues real _temps_visitor_id and _temps_sid cookies"
"ok": false, "detail": "no _temps_visitor_id in Set-Cookie: []"
```

`Build Test Binaries` shows `CANCELLED` as a downstream consequence of that job, not an independent failure.

PR #741, opened 2026-08-20
feat(projects): allow renaming a project's display name

dviejokfs:
## Why

The project settings page let you change the **slug** but not the **name**. Its single input was labelled "Project Slug" while the underlying form field was still called `name` and submitted `{ slug: values.name }` — the name field had been repurposed into the slug field at some point and nothing replaced it. From the console's point of view `projects.name` was write-once.

`PUT /projects/{id}` could already write the name (`update_project`), but it's a full-object replace requiring every field, so nothing in the UI used it to rename. The sparse endpoint the settings page actually calls, `POST /projects/{project_id}/settings`, had no `name` at all.

## What

- `name: Option<String>` on `UpdateProjectSettingsRequest`, threaded through `ProjectService::update_project_settings`, trimmed and validated to 1–100 characters.
- A real **Project Name** input above the slug input in `GeneralSettings.tsx`; the slug field is now backed by its own `slug` form key instead of masquerading as `name`. Post-save navigation now uses `values.slug` (it was already the slug — just under the wrong name).
- `--name` on `temps projects settings` for CLI parity, plus an interactive prompt.
- The new name is recorded in the `ProjectSettingsUpdated` audit entry, so a project that suddenly reports under a different name in dashboards and alerts can be traced back to who renamed it.

**The slug remains the routing identifier.** Renaming never touches a slug, an environment subdomain, or a domain row — that's asserted by a test, because a rename that silently moved a project's URL would break every domain pointing at it.

## Operator-visible consequence

`name` feeds `OTEL_SERVICE_NAME` for subsequent deployments, so renaming starts a *new* series in traces and metrics. Rather than let that surprise someone weeks later, the field description says so directly in the UI.

## Evidence

Six rename tests, running against real Postgres with real migrations:

```
$ TEMPS_TEST_DATABASE_URL=... cargo test --lib -p temps-projects rename
running 5 tests
test ..._renames_without_touching_slug ... ok
test ..._reports_no_rename_for_a_no_op ... ok
test ..._rename_does_not_persist_when_slug_conflicts ... ok
test ..._rename_not_applied_when_preset_is_invalid ... ok
test ..._renames_alongside_git_settings ... ok
test result: ok. 5 passed; 0 failed; 129 filtered out
```

What each one pins down:

- `renames_without_touching_slug` — sends `"  New Display Name  "`, asserts the persisted name is trimmed and the slug is untouched, and that the service reports `rename.from`/`rename.to` for the transition it actually performed.
- `reports_no_rename_for_a_no_op` — a name that trims back to the stored one reports no rename, so the audit records nothing.
- `rename_does_not_persist_when_slug_conflicts` — a rename submitted with a conflicting slug returns `SlugAlreadyExists` **and** leaves the original name on the row.
- `rename_not_applied_when_preset_is_invalid` — validation ordering: a rejected request renames nothing.
- `renames_alongside_git_settings` — routes the rename through the git-settings transaction (the second, more intricate implementation) and asserts the name and `main_branch` commit together.

Full crate suite, no regressions:

```
$ cargo test --lib -p temps-projects
cargo test: 134 passed (1 suite, 98.40s)
```

Lint / typecheck / spec, whole workspace:

```
$ cargo clippy --workspace --all-targets -- -D warnings   # clean
$ cargo fmt --all -- --check                              # clean
$ cd web && npx tsc --noEmit                              # clean
$ cd apps/temps-cli && bun run typecheck                  # clean
$ cd apps/temps-cli && bun run spec:check
openapi.json is canonical (716 paths)
```

CI on this branch: 31/31 green (1 skipped matrix entry).

## Not verified

**The UI change has not been exercised in a browser.** It typechecks and lints, and the endpoint behind it is covered by the tests above, but nobody has actually loaded the settings page and clicked Save. Worth a reviewer doing that before merge.

Generated clients (`web/src/api/client/types.gen.ts`, `apps/temps-cli/src/api/types.gen.ts`, `apps/temps-cli/openapi.json`) were hand-edited to add exactly this one field rather than regenerated against a running server — `spec:check` confirms the spec is still canonical and the diff is 7 added / 0 removed lines. Re-running the generators should be a no-op; if a reviewer prefers, that's easy to confirm.


dviejokfs:
Thanks — all three P1s were real. Fixed in 5c10d2107.

**1. Rename persisted after failure.** Correct, and the worst of the three. The name was written before the slug block, so `{name: valid, slug: conflicting}` returned `SlugAlreadyExists` while the project quietly kept the new name.

Split validation from persistence: the name is now normalized up-front alongside the other input checks (so an unusable name still fails before anything is written), but the write is deferred until every fallible step has succeeded. Where a git-settings update is also happening it's folded into that same `ActiveModel`, so a name+git change commits atomically; otherwise it lands just before the final reload.

Regression test added — `test_update_project_settings_rename_does_not_persist_when_slug_conflicts` inserts a second project holding the target slug, submits a rename plus that slug, and asserts the error comes back *and* the original name is still on the row.

Worth being explicit about the boundary: this makes the *rename* safe against later failures, but `update_project_settings` still applies its other fields (attack mode, git connection, preview settings) as independent non-transactional writes, so other combinations can still half-apply. That's pre-existing on `main` and making the whole function transactional is a real refactor — I didn't want to smuggle it into this PR, but it's worth an issue.

**2. Audit recorded the unpersisted name.** Right — the service trims, the handler audited `settings.name` raw. Now records `updated_project.name` whenever the request supplied a name, and the field doc says exactly that instead of claiming "when the request changed it".

**3. Stale form across project switches.** Right, and my change widened the blast radius: previously a stale submit could only clobber the slug, now it could clobber name *and* slug. Added a `reset()` effect keyed on the project. Note the sibling forms in this component (`deploymentForm`, `previewForm`) have the same mount-only `defaultValues` staleness — pre-existing, left alone here.

Verification after the fixes:

```
$ cargo test --lib -p temps-projects
cargo test: 130 passed (1 suite, 90.07s)

$ cargo clippy -p temps-projects --all-targets -- -D warnings   # clean
$ cd web && npx tsc --noEmit                                    # clean
```

The "not browser-verified" caveat in the PR description still stands for the UI change, including the new reset effect.

dviejokfs:
Both new P1s are accurate readings of the code, but I don't think either should change it. Reasoning below — happy to be overruled on the second one.

### Rename persists after route-reload failure (`project.rs:1747`)

Factually correct: `active_project.update()` commits, and only then can `reload_routes_after_compose_port_change` return `RouteReloadFailed`.

But the name isn't special here — that single `update()` also commits `preset`, `preset_config`, `directory`, `repo_owner`, `repo_name` and `main_branch`. If the route reload fails, *all* of those are already persisted. That's the pre-existing behaviour of this branch on `main`; folding the rename into the same write is what makes it consistent with its siblings, and was the fix for the earlier P1 about the rename committing separately.

Rolling the name back on this path would make things worse, not better: you'd get a project whose preset and public ports moved but whose name didn't, which is a harder state to reason about than "everything applied, routes may be stale".

It's also worth noting what this error actually is. `reload_routes_after_compose_port_change` only fails when **both** the queue send and the `pg_notify` fail — and it's a post-commit propagation step. The database already holds the state the user asked for; the error means "routes may be stale, go intervene", not "the write didn't happen". Undoing a committed field in response to a failed notification would be the wrong shape of fix.

The genuine fix is the transactional refactor of `update_project_settings` I flagged in my previous comment. Not smuggling it into this PR.

### Unchanged name produces a rename audit (`handlers.rs:1223`)

Also correct: supplying the existing name records `Some(name)` even though no row changed.

I'd argue that's the right behaviour for this struct, though. `ProjectSettingsUpdatedFields` records *what the request asked to set*, not *what changed* — `slug: settings.slug`, `image_retention_hours: settings.image_retention_hours` and `compose_configuration_updated` all behave exactly this way today. Making `name` alone mean "only when it actually differed" would make it the odd one out, and would need an extra pre-read to determine.

The field doc says what it does: *"the persisted display name after the update, set whenever the request supplied a name"*. The previous round's finding was that it recorded an **untrimmed** value that never reached the database; that part is fixed — it now records the post-trim persisted value.

If you'd rather have true change-detection, the consistent way to do it is a `previous_name` alongside it, mirroring the existing `previous_image_retention_hours` pair. Say the word and I'll add it — but I'd do it for `slug` at the same time rather than for `name` only.

No code changes pushed for these two; the branch is unchanged since 5c10d2107.

dviejokfs:
Reversing my previous comment — both are fixed in d16f839f4.

### Rename persists after route-reload failure

Fixed at the root rather than for the rename specifically, which also resolves the consistency objection I raised: the write and the reload signal now share a transaction, so this covers every field these paths write, not just `name`.

The NOTIFY is issued on the transaction that carries the write. PostgreSQL queues notifications and delivers them at commit, so a rolled-back update signals nothing, and a failure to signal rolls the write back with it.

One thing that fell out while implementing it, worth flagging because it would have been a quiet regression: the in-process queue job **cannot** move inside the transaction. Its subscriber (`route_reload_subscriber.rs`) reloads the route table straight from the database, so enqueuing pre-commit races the commit — the subscriber could read the *old* ports and latch them in with nothing left to trigger a re-read. That's a worse failure than the one being fixed, since it's silent. So the job is now sent after commit, and only when NOTIFY succeeded. If NOTIFY itself is unavailable the job is sent eagerly as a fallback, which preserves the existing "both mechanisms must fail before the request fails" contract.

Both call sites updated (`update_project_settings` and the git-settings path).

### Unchanged name produces a rename audit

Fixed. The audit records a rename only when the persisted name actually differs from the previous one, so re-submitting the existing name — or a value that trims to it — records nothing. Added `previous_name` alongside it, mirroring the existing `previous_image_retention_hours` pair, and the single pre-update read now serves both fields rather than adding a query.

The decision is extracted into `audited_rename()` with four unit tests covering: no name requested, no-op rename, real rename, and unknown previous value. That last case records the rename — if the pre-update read failed we can't prove it was a no-op, and over-reporting is recoverable during an incident review while silently dropping a real rename is not.

`slug` still records whatever was submitted rather than only real changes. That's pre-existing and I left it alone rather than widen this PR further; happy to do it in a follow-up.

```
$ cargo test --lib -p temps-projects
cargo test: 134 passed (1 suite, 78.54s)

$ cargo clippy -p temps-projects --all-targets -- -D warnings   # clean
```

The UI change is still not browser-verified — unchanged caveat from the PR description.

dviejokfs:
Both fixed in e79cfd938.

### Fallback reload reads stale routes

Good catch, and the security framing is what changed my mind. I'd noted this race in the code comment and consciously accepted it — "a stale route table beats no reload signal at all" — but I was weighing it as a latency/staleness problem. Framed as a **port removal that fails open**, it isn't: the operator withdraws a public port, the request returns success, and the route stays reachable indefinitely with nothing left to correct it. Silently failing open is worse than the failure the fallback existed to prevent.

So the fallback is gone. An unpublishable NOTIFY now rolls the write back and returns `RouteReloadFailed`. That's a stricter contract than before — previously both mechanisms had to fail — but it fails closed, and the operator retries against a project that was genuinely left untouched.

`RouteReloadFailed`'s message needed fixing too. It read *"Project {id} was saved, but its proxy routes could not be reloaded"*, which is now the exact opposite of what happens. It now says the change was rolled back and to retry once the database is reachable. Dropped the `queue_reason` field with it, since there's no longer a queue attempt on that path.

The in-process job still goes out after commit, where it can't race.

### No-op audit keeps previous name

Fixed — `name` and `previous_name` are recorded as a pair, so a no-op rename clears both. Added `audited_rename_and_previous_name_are_recorded_as_a_pair` covering both directions.

### Unrelated: workspace clippy was broken

Flagging because it's in this branch and isn't mine. The pre-commit clippy gate started failing on crates I haven't touched — clippy 1.98 added `chunks_exact_to_as_chunks` and tightened `useless_format`. CI installs `stable` unpinned and there's no `rust-toolchain.toml`, so this breaks the workspace clippy gate for **everyone** as soon as runners pick up 1.98; it isn't a local-only artifact. I confirmed it reproduces with my changes stashed.

Three mechanical fixes in `f318da500` (`temps-query-redis`, `temps-deployments`, `temps-cli`), kept in their own commit so they're easy to split out or cherry-pick ahead of this PR. Happy to move them to a separate PR if you'd rather — say the word.

```
$ cargo clippy --workspace --all-targets -- -D warnings   # clean
$ cargo fmt --all -- --check                              # clean
$ cargo test --lib -p temps-projects
cargo test: 135 passed (1 suite, 92.73s)
$ cargo test --lib -p temps-query-redis
test result: ok. 5 passed
```

UI still not browser-verified — unchanged caveat.

dviejokfs:
Both fixed in 84925d115.

### Concurrent rename audit is incorrect

Correct, and it's worse than stated: the same race governed *whether a rename was recorded at all*, not just the previous value. The handler read the name, called the service, and inferred the transition by comparing — a read-then-write with no lock between them.

Patching the comparison wouldn't fix that, so the inference is gone. `update_project_settings` now returns `ProjectSettingsUpdate { project, renamed_from }`, where `renamed_from` is the name observed **under the same row lock that carried the write**, set only when the write actually changed it. Both name-writing paths take `lock_exclusive`, so a reported transition is always one that really happened.

The handler now records `name` and `previous_name` straight from that. As a side effect this makes the pairing you flagged last round correct by construction rather than by a separate guard, and the handler's pre-update read narrows back to `image_retention_hours` — the only field still auditing a previous value it reads itself.

`audited_rename` and its unit tests are deleted; the service now answers what the helper was guessing at. Replaced with service-level tests asserting `renamed_from` on a real rename and on a whitespace-only no-op.

The trade-off worth naming: this changed the service's return type, so ~12 call sites moved to `.project`. Larger diff than a local patch, but the alternative was leaving a field that can lie under concurrency.

### Same-project refresh discards edits

Fixed. The reset effect is now keyed on project *identity* via a ref, not on its values, so it fires when the route switches projects and never when the same project is refetched. Typing is no longer clobbered mid-edit.

```
$ cargo test --lib -p temps-projects
cargo test: 131 passed (1 suite, 83.18s)

$ cargo clippy --workspace --all-targets -- -D warnings   # clean
$ cargo fmt --all -- --check                              # clean
$ cd web && npx tsc --noEmit                              # clean
```

(131 rather than 135: five `audited_rename` unit tests removed, one service test added.)

One unrelated thing seen while running these: `handlers::custom_domains::tests::foreign_target_environment_is_rejected_before_audit` failed once under the full parallel run and passes in isolation and on re-run. It's in a module this PR doesn't touch — flagging it as a pre-existing flake rather than quietly re-running until green.

UI still not browser-verified.

dviejokfs:
I don't think this one holds — the reconciliation path it says is missing does exist.

**The serving process listens to the NOTIFY itself.** `temps-cli/src/commands/serve/mod.rs:320` starts `ProjectChangeListener`, which `LISTEN`s on `project_route_change` (`temps-routes/src/project_change_listener.rs:46`). The transactional NOTIFY therefore reaches *this* process too — the post-commit queue job is a redundant fast path, not the only way the local route table finds out. A failed enqueue does not by itself leave the local proxy stale.

**A dropped listener is not silent, and it does reconcile.** `pg_listener.recv()` returns `Err` when the connection goes away. The handler logs, sleeps 5s, reconnects — and then falls through to an unconditional `peer_table.load_routes()` (`project_change_listener.rs:88`), whose comment states the intent outright: *"Recovery reload after a listener error / reconnect."* So the "listener silently disconnected" case ends in a full route reload, which is exactly the reconciliation the finding says is absent.

For the described outcome you'd need the queue send to fail **and** the listener to be dead **and** its reconnect-plus-reload to fail too, at which point the error log is the honest signal — there is nothing left to retry against.

### The one real gap, which is pre-existing and not this PR's

There's no keepalive or periodic timer (`"Pure event-driven loop: react to PG NOTIFY only... No periodic timer."`). A connection that is wedged *without* surfacing an error — a TCP black hole rather than a reset — would stall without triggering the reconnect path. That's a genuine hole, but it is in `temps-routes`, predates this branch, and would be fixed with a listener keepalive rather than by changing how this endpoint signals. Happy to open an issue; it shouldn't gate a project-rename PR.

No code change for this one. If you'd rather I add the keepalive anyway, say so and I'll do it in a separate PR against `temps-routes`.

dviejokfs:
Two of the three points behind the 2/5 were correct and are fixed in d100188bf. The third I still dispute.

### Rename-only path commits before a fallible unlocked reload — correct

The rename committed, then the project was re-read outside the lock. If that read-back failed, the caller got an error over an already-persisted rename: exactly the class of bug the route-reload change fixed earlier on this branch, which I'd left in place two lines further down.

The locked read, the write, and the read-back now share one transaction, so a failure anywhere rolls the rename back with it. The no-rename path keeps its plain read — there's no mutation to lose there.

### Concurrent renames can generate false audit transitions — correct, and my previous fix was half a fix

I fixed the `previous_name` side last round and missed that the *new* name was still taken from `update.project`, which came from the post-commit unlocked read. So a concurrent rename landing in between supplied the "after" of a transition whose "before" came from this request — a transition that never happened. Capturing one end under a lock is worthless if the other end is read outside it.

Both ends are now captured inside the transaction as a single `ProjectRename { from, to }` that the audit consumes directly. The pair cannot come from different writes because there is only one place that constructs it.

### Route-reload signal loss — still disputed

Unchanged from my previous comment, which I don't think was addressed: the serving process itself `LISTEN`s on `project_route_change` (`serve/mod.rs:320` → `project_change_listener.rs:46`), so the transactional NOTIFY reaches it and the queue job is a redundant fast path rather than the only mechanism. A dropped listener isn't silent either — `recv()` errors, and the handler reconnects and then runs an unconditional `load_routes()` whose own comment calls it *"Recovery reload after a listener error / reconnect."*

The real gap is narrower than the finding states: there's no keepalive, so a connection wedged *without* erroring would stall. That's pre-existing in `temps-routes`, isn't introduced or worsened here, and wants a listener keepalive rather than a change to how this endpoint signals. Happy to open an issue or a separate PR; I don't think it should gate a project-rename change.

```
$ cargo test --lib -p temps-projects
test result: ok. 132 passed; 0 failed

$ cargo clippy --workspace --all-targets -- -D warnings   # clean
$ cargo fmt --all -- --check                              # clean
```

Also note the branch now has `origin/main` merged in (7afdaa66d) and is conflict-free.

Standing caveat, unchanged: the UI has not been exercised in a browser.

dviejokfs:
Ran a full review pass (security-auditor gate + Rust standards + frontend/SDK + tests + evidence). Everything it found is fixed in 191931532, except one item I'm deliberately leaving open — noted at the bottom.

## Security: signed off

No blocking or major findings. Verified clean: the `pg_notify` payload is bound (`$1`), not interpolated, and no `format!`-built SQL was added anywhere; auth guards precede every new read and write; the rename cannot cross a project boundary because `name` is never a lookup key. The `name` → `OTEL_SERVICE_NAME` flow was traced to both sinks — Bollard's `env` array (no shell) and the Compose env-file renderer (quotes, escapes, doubles `$`) — and the display name never reaches a filesystem path or the Compose project name, which is built from IDs. The route-reload change was assessed as a security *improvement*: it converts a fail-open into a fail-closed.

## Blocking — fixed

**`RouteReloadFailed` told the operator nothing was saved, and that was false.** I introduced this wording two rounds ago while fixing the opposite bug. In `update_project_settings` six independent commits land before the route reload is published — including a slug rename that moves every environment subdomain — so only the final transaction rolls back. The variant now carries `rolled_back_scope`: `update_git_settings` keeps the true "No changes were saved.", and `update_project_settings` says its preset/directory/name portion was rolled back and other settings may already be saved.

**The rename inside the git-settings transaction had zero coverage.** Every existing rename test left the git fields `None`, so `needs_git_update` was false and that path never executed — the more intricate of the two implementations was the untested one. Added `renames_alongside_git_settings` (asserts the name and `main_branch` commit together) and `rename_not_applied_when_preset_is_invalid`.

**Evidence was stale and UI-less.** The block cited output from the first commit, five commits behind. Rewritten with current output and a per-test description of what each one actually pins down.

## Major — fixed

**23 positional arguments, with `name` and `slug` adjacent and both `Option<String>`.** Transposing them compiled silently and would rename a project when the caller meant to change its URL. Replaced with `UpdateProjectSettingsParams` plus a `From<UpdateProjectSettingsRequest>` impl — the handler call is one line, and the tests lose their walls of bare `None`. That refactor is why this commit is net **−107 lines** despite adding two tests.

**The form navigated to the submitted slug, not the persisted one.** The server slugifies, so saving `"My Slug"` routed to `/projects/My Slug/...` — a URL that doesn't exist. Now navigates using the slug from the response, and the client validates the DNS-safe shape up front instead of failing silently. Pre-existing, but on a line this PR touched.

## Minor — fixed

Control characters are now rejected in the display name (they pass the old bound, then break *every* later Compose deploy with an error naming `OTEL_SERVICE_NAME` — a 200 now, a confusing failure later). The two divergent rename implementations are deduplicated into `locked_rename`. Dropped a redundant `updated_at` that `before_save` already sets. Corrected the doc comment claiming NOTIFY delivery is guaranteed — it isn't durable; recovery comes from the listener's reconnect reload. Documented that this endpoint is *not* all-or-nothing. Swapped the unused `Serialize` derive on the new service types for `Debug`/`Clone`. The CLI echoes persisted values instead of submitted ones.

## Still open

**The UI has not been exercised in a browser.** It typechecks, lints clean, and the endpoint behind it is covered — but nobody has loaded the settings page and clicked Save, and I'm not going to describe it as verified when it isn't. Happy to bring up a local instance and walk it through if you want that before merge.

Noted for follow-up, not fixed here: `slug` still audits the raw request value rather than the slugified one (pre-existing, and the same class of bug as the `name` fix), and the 1-100 character bound is bypassable via `create_project`/`update_project`, which apply no name validation at all.

```
$ cargo test --lib -p temps-projects
cargo test: 134 passed (1 suite, 98.40s)

$ cargo clippy --workspace --all-targets -- -D warnings   # clean
$ cargo fmt --all -- --check                              # clean
$ cd web && npx tsc --noEmit                              # clean
$ cd apps/temps-cli && bun run typecheck                  # clean
```

dviejokfs:
Response to the latest review pass.

**1. "The git-settings transaction writes a full project snapshot captured before its row lock, so a concurrent unrelated update can be lost."**

This does not reproduce. The claim assumes Sea-ORM's `ActiveModel::update()` emits every column of the model; it only emits the columns marked `Set`. Columns left `Unchanged` are absent from the `UPDATE` statement, so a snapshot taken before the lock cannot clobber a column it never touches.

I verified this against a real Postgres instance rather than by reading: load a project, have a concurrent session flip `attack_mode` and change `directory`, then commit the stale-snapshot git-settings write. Result:

```
PROBE RESULT: attack_mode=true directory=probe-dir-changed
```

Both concurrent changes survived. The probe is now a permanent regression test — `test_partial_update_does_not_clobber_concurrent_columns` in `crates/temps-projects/src/services/project.rs` — so if anyone later switches this path to a full-model write, the test fails rather than the behaviour silently regressing.

**2. "The route reload path still relies on either a live PostgreSQL listener or a successful post-commit queue send, with no reconciliation when the listener dies silently and the queue send fails."**

The premise about silent listener death is not accurate for this codebase: the listener in `temps-routes` performs a full route reload on every reconnect, so a dropped-and-reestablished listener converges rather than sitting stale. That is what makes the fail-closed design correct — the `NOTIFY` is published on the writing transaction, so if it cannot be published the write is rolled back and the caller sees an error, instead of committing a route change nothing will ever apply.

The residual gap is narrower than described: a listener whose TCP connection is half-open without the OS reporting it would neither reconnect nor receive notifications. That is a missing keepalive in `temps-routes`, it predates this branch, and it is not something this PR introduces or is scoped to fix. Happy to open a follow-up issue for it.

Signed-off-by: David Viejo <dviejo@kfs.es>

dviejokfs:
Fixed the remaining finding in `e5507f3e4`, and I was wrong to argue it down.

My earlier reply claimed the listener converges because "a dropped connection surfaces as an error and its reconnect path runs a full route reload." That is only true when the connection reports an error. `crates/temps-routes/src/project_change_listener.rs` waited on `pg_listener.recv()` with no timeout and carried the comment `No periodic timer`, so a `LISTEN` connection dropped by a NAT, firewall, or load balancer idle timeout parks `recv()` on a half-open socket forever — no error, no reconnect, no reload. I called that a narrow edge case. It is the ordinary way an idle `LISTEN` connection dies, and the review was right that a withdrawn public port could stay reachable indefinitely.

The fix caps the wait at `IDLE_RECONCILE_INTERVAL` (60s) and reconciles the route table when it elapses. A silently dead listener now degrades to bounded polling instead of unbounded staleness, so losing both the NOTIFY and the post-commit queue job costs at most one interval of stale routing rather than permanent fail-open.

On the alternative fix: sqlx 0.8.6 has no TCP keepalive support (there is no `keepalive` setting anywhere in `sqlx-postgres` or `sqlx-core::net`), so surfacing the dead socket as an error is not available without hand-rolling socket configuration. The bounded wait is the mechanism actually available.

Cost is one route-table load per minute on an otherwise idle control plane — below what an active install already does, since every deployment reloads through this same path.

Testing: the loop needs a live PostgreSQL listener to exercise, so it is not unit-tested. The accompanying test pins the part that can be checked in isolation — that the interval stays a real bound — so widening it to tens of minutes fails rather than quietly restoring the fail-open window. `cargo test --lib -p temps-routes` passes (69 tests); clippy clean on `temps-routes` and `temps-projects` with `-D warnings`.

I also corrected the rationale comment on `enqueue_route_reload` in `crates/temps-projects/src/services/project.rs`, which rested entirely on the reconnect path that does not fire in this scenario.

PR #747, opened 2026-08-21
feat(projects): always show a health indicator on the project card

dviejokfs:
## Problem

The project card already had a `HealthStatusDot`, but it was rendered as:

```tsx
{health && health.status !== 'unknown' && <HealthStatusDot status={health.status} />}
```

The backend derives `status` from the proxy error rate over the requested window and returns **`"unknown"` for any project that received zero requests** (`finish_aggregation` in `crates/temps-proxy/src/service/proxy_log_service.rs` returns early when `total_requests == 0`). So on a real projects list most cards showed **no health indicator at all**, and the ones that did showed a bare 8px coloured dot whose only explanation was a `title` attribute.

That is the failure mode `CLAUDE.md` calls out directly: a feature that conditionally renders nothing teaches the operator that Temps has no health signal. It also made two very different situations indistinguishable — "we measured it and it is fine" vs. "nothing reached it, so there is nothing to measure" vs. "we could not ask".

## Change

- New pure helper `web/src/components/dashboard/project-card-health.ts` maps `(health, loading, error)` onto a labelled indicator. Six explicit states, none of which is "render nothing":

  | State | Label | When |
  |---|---|---|
  | `healthy` | Healthy | status `healthy`/`operational` |
  | `degraded` | Degraded | error rate > 10% |
  | `down` | Down | error rate > 50% |
  | `idle` | No traffic | `total_requests === 0` (what the backend calls `unknown`) |
  | `pending` | Checking… | first load in flight |
  | `unavailable` | Unavailable / No health data / Unknown | fetch failed, project missing from the summary, or an unrecognised status |

- `ProjectCard` renders `<ProjectHealth>` unconditionally in all three layouts (`compact`, `dense`, `wide`), replacing the old conditional dot.
- The derivation is surfaced rather than hidden: the tooltip and an `sr-only` span carry e.g. *"Serving normally: 10,607 requests over the last 24 hours, 3 server errors (0.1%), 42 ms average response time."* A project with no traffic explains *why* there is no signal instead of implying health.
- An unrecognised status reports the real numbers rather than silently falling back to "idle".
- Dark-mode variants on every tone.

No API change — this reads data the dashboard health endpoint already returns.

## Evidence

**Unit tests** — 9 new cases in `project-card-health.test.ts`, including the specific regression (zero-traffic project must produce `No traffic`, not nothing) and that a background refetch does not blank out health already loaded:

```
$ cd web && bun test src/components/dashboard
 33 pass
 0 fail
 62 expect() calls
Ran 33 tests across 5 files.
```

**Typecheck + lint** — clean:

```
$ bunx tsc --noEmit          # no output
$ bunx eslint src/components/dashboard/ProjectCard.tsx \
      src/components/dashboard/project-card-health.ts \
      src/components/dashboard/project-card-health.test.ts   # no output
```

**Rendered** — all six states were rendered through the real `ProjectCard` in a temporary route driven by fixture data, screenshotted with Playwright, and the harness removed before commit (it is not in this diff). Confirmed in three passes:

- *Light, 1500px, 3 columns* — `Healthy` (green), `Degraded` (amber), `Down` (red), `No traffic` (grey), `Checking…` (pulsing grey), `Unavailable` (grey) all render beside the project name; the previously-invisible zero-traffic card now reads `temps-cloud-worker ● No traffic`.
- *Dark, 1500px* — tones stay legible against `bg-card` via the `dark:text-*-400` variants.
- *Narrow, 700px, 2 columns* — the indicator is `shrink-0` so it survives; the project name truncates while the untruncated slug on the line below keeps the card identifiable.

Also verified in the same passes that the `dense` and `wide` layouts pick up the identical indicator, and that no console errors are produced by the component (the only console output was 504s from the API proxy, expected with no backend attached to the fixture route).

### DIFF
--- a/web/src/components/dashboard/ProjectCard.tsx
+++ b/web/src/components/dashboard/ProjectCard.tsx
@@ -22,6 +22,11 @@ import { VisitorSparkline } from './VisitorSparkline'
 import { ProjectCardMedia } from './ProjectCardMedia'
 import { projectCardTraffic } from './project-card-traffic'
 import {
+  projectHealthIndicator,
+  type ProjectHealthIndicator,
+  type ProjectHealthTone,
+} from './project-card-health'
+import {
   deploymentLabel,
   projectBuildSource,
   projectPresetLabel,
@@ -44,26 +49,37 @@ interface ProjectCardProps {
   }
 }
 
-function HealthStatusDot({ status }: { status: string }) {
-  const colors: Record<string, string> = {
-    operational: 'bg-emerald-500',
-    healthy: 'bg-emerald-500',
-    degraded: 'bg-amber-500',
-    down: 'bg-red-500',
-    no_monitors: 'bg-zinc-400',
-    unknown: 'bg-zinc-400',
-  }
-  const label =
-    status === 'no_monitors'
-      ? 'No monitors'
-      : status.charAt(0).toUpperCase() + status.slice(1)
+const HEALTH_TONE_STYLES: Record<
+  ProjectHealthTone,
+  { dot: string; text: string }
+> = {
+  healthy: {
+    dot: 'bg-emerald-500',
+    text: 'text-emerald-600 dark:text-emerald-400',
+  },
+  degraded: { dot: 'bg-amber-500', text: 'text-amber-600 dark:text-amber-400' },
+  down: { dot: 'bg-red-500', text: 'text-red-700 dark:text-red-400' },
+  idle: { dot: 'bg-zinc-300', text: 'text-muted-foreground' },
+  unavailable: { dot: 'bg-zinc-400', text: 'text-muted-foreground' },
+  pending: { dot: 'bg-zinc-300 animate-pulse', text: 'text-muted-foreground' },
+}
+
+/**
+ * Always rendered. There is no health state — including "we could not measure
+ * it" — that is better communicated by drawing nothing.
+ */
+function ProjectHealth({ indicator }: { indicator: ProjectHealthIndicator }) {
+  const tone = HEALTH_TONE_STYLES[indicator.tone]
 
   return (
     <span
-      className={`inline-block size-2 rounded-full ${colors[status] || colors.unknown}`}
-      title={label}
-      aria-label={label}
-    />
+      className={`inline-flex shrink-0 items-center gap-1.5 text-xs ${tone.text}`}
+      title={indicator.detail}
+    >
+      <span className={`inline-block size-2 rounded-full ${tone.dot}`} />
+      <span className="whitespace-nowrap">{indicator.label}</span>
+      <span className="sr-only">. {indicator.detail}</span>
+    </span>
   )
 }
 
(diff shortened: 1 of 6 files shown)


dviejokfs:
## Follow-up: the "unknown" case

Reported case — a project whose production monitor shows **100% Healthy** still rendered `unknown` in its header badge.

### Root cause

The badge was never showing monitor health. `ProjectDetailHeader.tsx` called `useDashboardHealth`, i.e. `/proxy-logs/stats/projects-health`:

- default window is **1 hour** when no dates are passed (`proxy_logs.rs:840`)
- the query filters `is_system_request = FALSE` (`proxy_log_service.rs:1487`), which **deliberately excludes Temps' own uptime-monitor requests** — it measures *user* traffic by design

So the monitor proving the project is up was filtered out of the very question being asked, and with no human visitor in the last hour `total_requests == 0` → `finish_aggregation` returns before setting a status → `"unknown"`.

The tell that this was a wiring bug rather than intent: both the header and the list card map the statuses `operational` and `no_monitors`, and the proxy-health endpoint **can never return either**. They only come from `get_projects_monitor_health`. The badge also links to `/monitors`.

`GET /monitors-health/projects` (`status_page.rs:740` → `monitor_service.rs:320`) already returns `operational | degraded | down | no_monitors` per project from the latest production status check, is already in the generated client as `getProjectsMonitorHealth`, and had **zero callers anywhere in `web/src`**.

### Fix

New `useProjectsMonitorHealth` hook, and `projectHealthIndicator` now takes a `monitor` input that outranks traffic:

1. A configured production monitor decides — it has an answer even for a project nobody visited, which is the entire point of it.
2. `no_monitors` (or an unreadable/403 monitors endpoint) declines to answer and falls through to traffic health.
3. Neither available → still not "unknown": `No traffic`, with a detail that says to add a monitor to get a visitor-independent signal.

Applied to the projects-list card *and* the project header, so a project reads the same in both places. No backend change — this calls an endpoint that already existed.

### Evidence

7 new unit tests (38 total in the suite, 0 fail), including the reported case verbatim — `total_requests: 0` + `status: 'unknown'` + `monitor: operational` must yield `Healthy` — plus monitor-outranks-clean-traffic, monitor-answers-when-the-traffic-query-failed, and both fallback paths.

```
$ cd web && bun test src/components/dashboard
 38 pass
 0 fail
 72 expect() calls
$ bunx tsc --noEmit    # clean
$ bunx eslint <touched files>    # clean
```

Rendered through the real `ProjectCard` with `total_requests: 0` on every card — the exact reported condition — via a temporary fixture route, screenshotted, harness removed before commit:

| monitor status | before | after |
|---|---|---|
| `operational` | *(nothing / unknown)* | ● **Healthy** |
| `degraded` | *(nothing / unknown)* | ● **Degraded** |
| `down` | *(nothing / unknown)* | ● **Down** |
| `no_monitors` | *(nothing / unknown)* | ● No traffic |
| endpoint 403 | *(nothing / unknown)* | ● No traffic |
| traffic query failed + monitor green | *(nothing / unknown)* | ● **Healthy** |

PR #749, opened 2026-08-21
fix(web): handle fetch errors in ApiKeyDetail and ApiKeyEdit with retry state

### LINKED ISSUE #748: fix(web): ApiKeyDetail and ApiKeyEdit show misleading "not found" on API/network failures
## Bug Description
In both `ApiKeyDetail.tsx` and `ApiKeyEdit.tsx`, if the initial API request fails due to a network drop, server timeout, or 500 error, the page displays a misleading "API key not found" message with no way to retry without reloading the entire application.

## Root Cause
Both pages only extract `{ data: apiKey, isLoading }` from `useQuery`, completely ignoring `error` and `refetch`. 
When an API error occurs, `apiKey` remains `undefined`, causing the components to fall through directly into the `if (!apiKey)` check and render the "not found" state.

### Steps to reproduce

1. Open DevTools → **Network** tab (or Request Blocking) and block requests to `*/api/api-keys/*` (or set throttling to Offline).
2. Navigate to an existing API Key detail page (`/settings/keys/:id`) or edit page (`/settings/keys/:id/edit`).
3. Notice the page renders "API key not found" rather than an error alert with a retry action.

### Expected behavior

- Transient network/server errors should display a distinct failure message with a **Retry** button.
- Genuine 404 responses should continue to display the "API key not found" screen.

### Actual behavior

- Server and network failures are misrepresented as missing data ("API key not found").
- Users are given no retry affordance to recover from temporary glitches.

### Operating System

Linux

### Temps version

Latest(main)

### Docker version

N/A

### Relevant logs

```shell

```

## Affected Files
- `web/src/pages/ApiKeyDetail.tsx`
- `web/src/pages/ApiKeyEdit.tsx`

Iqrima:
Closes #748 

## What changed

`ApiKeyDetail` and `ApiKeyEdit` previously showed a misleading "API key not found" message when data fetching failed due to transient server/network errors.

This PR:
- Extracts `error` and `refetch` from `useQuery` in both pages.
- Adds an error recovery state with an explicit **Retry** button.
- Preserves the dedicated "API key not found" state for genuine 404 responses.

## Files changed
- `web/src/pages/ApiKeyDetail.tsx`
- `web/src/pages/ApiKeyEdit.tsx`

## How to test
1. `cd web && bun run dev`
2. Open an API Key detail page (`/settings/keys/:id`) or edit page (`/settings/keys/:id/edit`).
3. DevTools → Request Blocking → block `*/api/api-keys/*`.
4. Refresh → verify "Failed to load API key" appears with a working **Retry** button.
5. Unblock → click **Retry** → verify data loads normally.

### DIFF
--- a/web/src/pages/ApiKeyDetail.tsx
+++ b/web/src/pages/ApiKeyDetail.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 import { useNavigate, useParams } from 'react-router'
 import { useQuery, useMutation, useQueryClient } from '@tanstack/react-query'
 import {
@@ -30,13 +30,15 @@ import {
 import { useApiKeyPermissions } from '@/components/api-keys/useApiKeyPermissions'
 import { usePageTitle } from '@/hooks/usePageTitle'
 import {
-  ArrowLeft,
-  Shield,
-  Key,
-  Calendar,
-  Activity,
+  AlertCircle,
   AlertTriangle,
+  ArrowLeft,
+  Calendar,
   Check,
+  Key,
+  RotateCcw,
+  Shield,
+  Activity,
   X,
 } from 'lucide-react'
 import { format } from 'date-fns'
@@ -88,7 +90,12 @@ export default function ApiKeyDetail() {
   const queryClient = useQueryClient()
   const [deleteModalOpen, setDeleteModalOpen] = useState(false)
 
-  const { data: apiKey, isLoading } = useQuery({
+  const {
+    data: apiKey,
+    isLoading,
+    error: apiKeyError,
+    refetch: refetchApiKey,
+  } = useQuery({
     queryKey: ['apiKey', id],
     queryFn: async () => {
       if (!id) throw new Error('API Key ID is required')
@@ -97,6 +104,14 @@ export default function ApiKeyDetail() {
     },
     enabled: !!id,
   })
+
+  useEffect(() => {
+    if (apiKey && apiKeyError) {
+      toast.error('Failed to refresh API key', {
+        action: { label: 'Retry', onClick: () => void refetchApiKey() },
+      })
+    }
+  }, [apiKey, apiKeyError, refetchApiKey])
 
   const { data: permissionsData } = useApiKeyPermissions()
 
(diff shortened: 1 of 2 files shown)


dviejokfs (dismissed):
## Review: fix(web): handle fetch errors in ApiKeyDetail and ApiKeyEdit with retry state

Solid fix for #748 — the code correctly separates hard fetch failures from genuine 404s, and the `isNotFound` heuristic checks out against the actual backend error shape (`ApiKeyServiceError::NotFound` → `status: 404`). Hook ordering is correct in both files (no hooks after early return). CI is fully green and the branch merges cleanly.

### Blocking
- **Missing PR evidence.** The "How to test" section lists manual repro steps but there's no proof they were actually run — no screenshot/GIF of the new "Failed to load API key" + Retry state, no before/after capture. For a UI-facing fix this is required before merge. Could you attach a screenshot or short clip of the retry flow (block `*/api/api-keys/*` → see the error state → unblock → Retry succeeds) on both `ApiKeyDetail` and `ApiKeyEdit`?

### Minor
- `useEffect` deps `[apiKey, apiKeyError]` in both files omit `refetchApiKey` (used inside the toast's `onClick`). Likely harmless since `refetch` is stable per query, but worth adding for correctness/lint-cleanliness.
- Re: the Greptile P1 comment on "cached data bypasses recovery state" — when a *background* refetch fails while stale `apiKey` data is still shown, the page doesn't show the full-page error block, but it does fire a toast with an explicit Retry action, which is a reasonable (if less persistent) recovery affordance. Not blocking, but worth confirming that's the intended UX rather than a persistent banner. The same underlying pattern also exists on the event-detail and error-group pages, which are out of scope here.
- `isNotFound`'s `.title?.toLowerCase() === 'not found'` branch is effectively dead code — the actual backend title is `"API Key Not Found"` (lowercases to `"api key not found"`, not `"not found"`). Detection still works via `.status === 404` and `.detail`, so no functional bug, just a no-op condition worth cleaning up.

No security, migration, or SDK-parity concerns — this is a self-contained frontend display-state change.

Iqrima:
Screenshot (790)

Attached the requested screenshot of the new "Failed to load API key" state with the Retry action. Also updated the useEffect dependencies and tightened the isNotFound check. Thanks for the review!

PR #753, opened 2026-08-21
fix(proxy): allow IPv6 addresses in IP Access Control rules

dviejokfs:
## Summary
- `IpAccessControlService::is_valid_ip_or_cidr` rejected every IPv6 address/CIDR at create/update time (hand-rolled dotted-decimal octet splitter), even though the `inet` storage column and the hot-path blocklist snapshot (`refresh_blocklist`/`is_blocked`) were already fully dual-stack via `ipnetwork::IpNetwork`/`std::net::IpAddr`. Replaced the validator with the same `IpNetwork` parser those two already use.
- Mirrored the fix on the frontend: `IpAccessControl.tsx`'s zod validator had an equivalent IPv4-only regex. Dropped it in favor of the existing `AdminGateCard` pattern already used elsewhere in this codebase (rely on the backend's real validation, surface errors via the existing mutation error toast) rather than reimplementing IP parsing in JS.
- Updated the now-inaccurate "IPv4 only" doc comments/UI copy: entity doc comment, request schema doc comment, CLI `--ip` help text, and the form's placeholders/help text.

Found via a sweep for other IPv4-only assumptions after fixing IPv6 discoverability gaps in a separate EE plugin's Network Access UI. This is the only genuine *logic* bug the sweep turned up (splitting a string on `.` will reject any address containing `:`); the rest were text-only discoverability gaps in unrelated files, tracked separately.

Not folded into #725 — unrelated feature, unrelated files.

## Test plan
- [x] `cargo check --lib -p temps-proxy -p temps-entities`
- [x] `cargo test --lib -p temps-proxy ip_access_control` — 3/3 pass, including new IPv6 assertions (`2001:db8::1`, `2001:db8::/32`, `::1`, and out-of-range `/129` rejection)
- [x] `cargo fmt --check` / `cargo clippy -D warnings` on both changed crates
- [x] `tsc --noEmit` on `web/` and `apps/temps-cli/` — no new errors introduced by this diff
- [x] Manual UI verification (create/edit an IPv6 rule end-to-end against a running server) — see comment below

### DIFF
--- a/apps/temps-cli/src/commands/ip-access/index.ts
+++ b/apps/temps-cli/src/commands/ip-access/index.ts
@@ -80,7 +80,7 @@ export function registerIpAccessCommands(program: Command): void {
     .command('create')
     .alias('add')
     .description('Create a new IP access control rule')
-    .option('--ip <ip_or_cidr>', 'IP address or CIDR range (e.g., "192.168.1.1" or "10.0.0.0/24")')
+    .option('--ip <ip_or_cidr>', 'IPv4 or IPv6 address or CIDR range (e.g., "192.168.1.1", "10.0.0.0/24", or "2001:db8::/32")')
     .option('--action <action>', 'Action to take: "allow" or "deny"')
     .option('--description <desc>', 'Optional description/reason for the rule')
     .option('-y, --yes', 'Skip confirmation prompts (for automation)')
--- a/crates/temps-entities/src/ip_access_control.rs
+++ b/crates/temps-entities/src/ip_access_control.rs
@@ -9,8 +9,9 @@ use temps_core::DBDateTime;
 pub struct Model {
     #[sea_orm(primary_key)]
     pub id: i32,
-    /// IP address in CIDR notation (e.g., "192.168.1.1" or "10.0.0.0/24")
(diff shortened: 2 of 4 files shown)


dviejokfs:
## Manual UI verification (checked off the test plan)

Verified the create/edit flow end-to-end against a real running server (isolated local Postgres + `temps serve` + the web UI dev server, throwaway admin account, all local-only, torn down after):

1. **Create** — Settings → Rate Limiting → IP Access Control → Add IP Rule, entered `2001:db8::/32` as a Block rule. Before this fix this was rejected client-side (regex) and would also have been rejected server-side (`is_valid_ip_or_cidr`). After the fix: submitted successfully, no validation error, rule appears in the Blocked IPs list. Dialog also now shows the updated "Supports IPv4 and IPv6 addresses, and CIDR notation" help text and the IPv6-inclusive placeholder.
2. **Edit** — opened the same rule, changed the address to a different IPv6 CIDR (`2001:db8:1::/48`), saved — update succeeded and the list reflects the new value.
3. **Cleanup** — deleted the test rule afterward.

No errors in the server log tied to the ip_access_control code path during either request. Checking that box off in the description now.

PR #764, opened 2026-08-22
fix(cli): let temps upgrade bundle shared libraries a release needs

dviejokfs:
## Summary
- Some release builds link against a shared library a bare host doesn't ship by default. Today the binary would exec, hit the dynamic loader, and die with "cannot open shared object file" before any of its own code ran — there's no way for a release to carry what it needs.
- Added `extract_libs_from_tarball`/`install_libs`/`rollback_libs`: a release's tarball can now carry an optional `lib/` directory alongside the binary, which — when present — gets installed into a `lib/` dir next to the installed binary, paired with a build that sets the binary's RPATH to `$ORIGIN/lib` so it resolves them without any host-side package install.
- Every OSS release has never had a `lib/` entry and never will, so `extract_libs_from_tarball` returns an empty `Vec` for those tarballs and `install_libs`/`rollback_libs` become complete no-ops — this changes nothing about how OSS installs behave.
- Wired into all three places that currently extract+replace a binary: `temps upgrade`'s OSS path (`run_oss`), its `--tier ee` path (`run_ee`), and the background self-updater (`self_update.rs`'s `download_verify_swap`).
- The self-updater additionally preflights the new binary before committing to the swap; since that preflight execs a probe binary from the same directory the real one will live in, libraries have to be in place first, not after — and if preflight then fails, the libraries are rolled back too (via a `lib.bak` backup, mirroring how the binary itself is backed up), so a still-running old binary that also resolves its own dependencies through this same `lib/` dir is never left pointed at an untested library set.

Found while investigating a `cannot open shared object file` crash reported on a downloaded build with extra native dependencies.

## Test plan
- [x] `cargo test --lib -p temps-cli` — 6 new tests covering the no-op-for-OSS case, extraction, first-time install, and backup/rollback around a failed preflight; all 217 existing tests still pass
- [x] `cargo fmt --check` / `cargo clippy -D warnings` clean
- [x] Live-verified the paired build-side change (bundling libs + `patchelf --set-rpath`): `ldd` shows all previously-missing libraries resolving to the bundled copies via RPATH, from an arbitrary install location, with the binary executing correctly

### DIFF
--- a/crates/temps-cli/src/commands/serve/self_update.rs
+++ b/crates/temps-cli/src/commands/serve/self_update.rs
@@ -38,9 +38,9 @@ use tracing::{error, info, warn};
 
 use crate::commands::upgrade::{
     check_write_permission, current_version_tag, download_asset, download_asset_text,
-    extract_binary_from_tarball, fetch_latest_release_in_channel, fetch_specific_release,
-    is_newer_version, platform_target, replace_binary, verify_checksum, GitHubRelease,
-    UpgradeChannel,
+    extract_binary_from_tarball, extract_libs_from_tarball, fetch_latest_release_in_channel,
+    fetch_specific_release, install_libs, is_newer_version, platform_target, replace_binary,
+    rollback_libs, verify_checksum, GitHubRelease, UpgradeChannel,
 };
 
 /// Grace period between accepting the update and exiting the process. Long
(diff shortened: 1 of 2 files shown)


dviejokfs:
Closing — the motivating problem (xmlsec crash on a bare host) is now fixed differently and more fundamentally: the EE release statically links libxml2/libxmlsec1/libxslt/ICU/liblzma as of temps-ee v0.1.0-beta.7, so there's no dynamic dependency left to bundle at all. The bundled-lib+RPATH approach in EE's release.yml that would have produced a populated `lib/` tarball for this mechanism to consume has also already been reverted.

This leaves `extract_libs_from_tarball`/`install_libs`/`rollback_libs` as generic infrastructure with no current caller and no other known native-dependency case motivating it — OSS's own binary only links libc/libm/libgcc_s/libz. Reopening this approach later, if a genuinely new bundling need comes up, would be better informed by that case's actual constraints than by carrying speculative, untested-in-production machinery now.

PR #765, opened 2026-08-22
test(deployer): add CI guardrail for BuildKit large-context hang

dviejokfs:
## Issue

The "Build Container Image" deployment stage could get stuck **In Progress** indefinitely, with the log panel showing only "Connecting to log stream..." and zero log lines. `docker buildx du --verbose` confirmed there was no new cache activity — the request never progressed past the BuildKit vertex `[internal] load remote build context`.

## Investigation

We reproduced a genuine hang locally: with a build context in the hundreds-of-KB range, BuildKit calls back over Bollard's `session` for the context via `moby.filesync.v1.FileSync`, a gRPC service Bollard's own session implementation never registers a handler for — so the call never gets a response.

**However**, the same request against a **native Linux `dockerd`** — what this codebase actually deploys to in production and CI — did not reproduce the hang across dozens of runs from 0 bytes up to 2MB, tested both locally against a `docker:dind` container and via the existing test suite. It reproduced consistently only on Docker Desktop for macOS, which routes the Docker socket through a VM/gRPC-FUSE forwarding layer that a native Linux socket doesn't have.

So this looks like a macOS-only artifact of local development, not a cross-platform Bollard/BuildKit defect. **No source change is included in this PR** — the existing `session`-based `build_image()` path already appears correct for the platforms this codebase actually runs on.

## What this PR does instead

Adds a regression test as a CI guardrail rather than closing the investigation silently: a 5MB build context (comfortably past the size where the Mac-only hang reproduced), run through the exact code path production uses. If it ever turns red in CI (which runs on Linux), that's a real signal this affects the actual deployment target after all, and the fix this investigation didn't end up needing should be revisited.

## Verification

- New test `build_image_buildkit_large_context_does_not_hang`: 5MB context, real Docker daemon, full `ImageBuilder::build_image` path.
- Full `temps-deployer` suite passes.

## Test plan

- [ ] CI passes on the Linux runner (this is the actual test of the theory above)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10994,7 +10994,6 @@ name = "temps-deployer"
 version = "0.1.0-beta.56"
 dependencies = [
  "anyhow",
- "async-stream",
  "async-trait",
  "bollard",
  "bytes",
@@ -11011,7 +11010,6 @@ dependencies = [
  "sea-orm",
  "serde",
  "serde_json",
- "serde_urlencoded",
  "serde_yaml",
  "serial_test",
  "sysinfo",
@@ -11029,7 +11027,6 @@ dependencies = [
  "urlencoding",
  "utoipa",
  "uuid",
- "zstd",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10994,6 +10994,7 @@ name = "temps-deployer"
 version = "0.1.0-beta.56"
 dependencies = [
  "anyhow",
+ "async-stream",
  "async-trait",
  "bollard",
  "bytes",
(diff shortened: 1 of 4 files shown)


dviejokfs:
Closing without merging. Investigated a locally-reported BuildKit build hang ("load remote build context" stuck indefinitely). Traced it to BuildKit calling back over Bollard's session for the build context via `moby.filesync.v1.FileSync`, a gRPC service Bollard's own session implementation never registers a handler for. Reproduced reliably on Docker Desktop for macOS, but the same request against a native Linux `dockerd` (what this codebase actually runs on in production and CI) never reproduced it across dozens of runs from 0 bytes up to 2MB. This is a macOS-only local Docker Desktop artifact, not a production/CI issue, so no fix is needed here.

PR #804, opened 2026-08-25
perf(cli): stream the upgrade download and extraction to disk

jlucaso1:
### Summary

`temps upgrade` held the whole release artifact in memory, twice over.

1. `download_asset` buffered the response into a `Bytes` and then `.to_vec()`'d it, so both copies of the ~110 MB tarball were live at once.
2. `extract_binary_from_tarball` read the uncompressed binary (~270 MB: `.text` ~185 MB, `.rodata` ~59 MB) into a `Vec` via `read_to_end`, which doubles its capacity while growing, with the tarball buffer still alive for the whole extraction.
3. `replace_binary` then handed those 270 MB to `fs::write`, so the page cache filled with the same amount of dirty pages on top.

On a self-hosted instance with 1 GB of RAM already running `temps serve`, that peak does not fit. The kernel goes into swap thrashing and, with a fraction of a vCPU available, nothing gets scheduled: the reported symptom is the whole host becoming unreachable (SSH times out during banner exchange), not just the command failing. The last line printed is "Replacing binary", because that `println!` happens before the write.

Everything now streams through disk. Nothing in the flow ever needs the artifact whole.

### Measurement

Peak RSS (`VmHWM` from `/proc/self/status`) of a harness running the two code paths over a synthetic 273 MB binary in a 73 MB tarball:

| Path | Peak RSS |
|---|---|
| before (buffer tarball, `read_to_end`, `fs::write`) | 353,880 kB |
| after (stream to disk, `io::copy`) | 2,488 kB |

Roughly 350 MB down to 2.5 MB, and the "before" figure is a lower bound: the harness reads the tarball once, while the real `download_asset` kept the `Bytes` and its `to_vec()` copy alive simultaneously, on a tarball half again as large.

Both paths produced a byte-identical binary (same SHA256).

### Changes

All in `crates/temps-cli/src/commands/upgrade.rs`.

- `download_asset_to_file` / `download_ee_asset_to_file` stream the response chunk by chunk into a temporary file and feed `Sha256` in the same loop, returning the hex digest. No second read for the checksum.
- `verify_computed_checksum` compares an already-computed digest against a `.sha256` body. `verify_checksum` keeps its signature and now delegates to it, so its callers and tests are untouched.
- `extract_binary_from_tarball_file` runs `GzDecoder` over the tarball `File` and `std::io::copy`s the entry into the staging file. The in-memory `extract_binary_from_tarball` stays for its other callers.
- Checksum verification still runs before anything is unpacked. That is why the tarball is staged to disk rather than piped download to gunzip to tar to binary: the pipeline would save 110 MB of scratch disk but would write unverified bytes over the live executable. The reasoning is recorded as a comment at the call site.
- The tarball is staged next to the target binary, not in `std::env::temp_dir()`, which is a tmpfs on many small hosts and would put those 110 MB straight back in RAM. That directory is already known-writable via `check_write_permission`.
- `ScopedTempFile` removes the download and staging files on the ordinary failure paths. It cannot cover a `SIGKILL`, so reaping orphaned `.temps-upgrade-*.<pid>` files needs a PID liveness check and is left for a follow-up.
- Validate before the rename: the extracted size is compared against the tar header's declared size, and a mismatch removes the staging file and fails. The verified checksum covers the tarball, not the write, so a short write (full disk, truncated stream) could otherwise put a plausible but incomplete file over the live executable.
- `sync_all` before the rename, in both the streaming and the `replace_binary` paths. The rename is atomic for metadata only, so without the fsync a power cut can leave a correctly sized file of garbage where the system's executable was.
- Both flows are converted: the public release path and the EE path.

Unchanged: printed messages, release artifact format, the atomic rename, and the PID suffix on the staging file.

### Validation

- `cargo check --lib -p temps-cli`: clean, no warnings.
- `cargo test --lib -p temps-cli`: 221 passed, 0 failed. The existing `verify_checksum` and `extract_binary_from_tarball` tests are unmodified and still pass.
- New tests cover: streaming extraction to disk; streaming output byte-identical to the in-memory path on a payload large enough that `read_to_end` reallocates; missing `temps` entry leaves nothing staged; missing tarball names the path in the error; `ScopedTempFile` removal and disarm; PID suffix on both temp paths; `finalize_staged_binary` renaming and setting mode 0755; checksum mismatch reporting both hashes.
- Invalid checksum: the comparison sits between the download and the extraction, so a mismatch aborts with the target binary untouched and both temporaries removed by their guards. The mismatch itself is covered by `test_verify_computed_checksum_reports_both_hashes_on_mismatch`.

## Type of change

- [ ] Documentation update

## Checklist

## Related issues

Reported against a self-hosted 1 GB instance upgrading on the nightly channel.

## Evidence

**Secure streaming and failure-path regressions**

```bash
cargo test --lib -p temps-cli commands::upgrade::tests -- --nocapture
```

```text
cargo test: 61 passed, 167 filtered out (1 suite, 0.05s)
```

This targeted suite includes multi-chunk HTTP streaming and SHA-256 verification, interrupted-response cleanup, random atomically-created sibling temp files, checksum-mismatch target preservation, corrupt-gzip cleanup, and successful atomic replacement.

**Full temps-cli library regression suite**

```bash
cargo test --lib -p temps-cli
```

```text
cargo test: 228 passed (1 suite, 2.06s)
```

**Formatting and compilation**

```bash
cargo fmt --all -- --check
cargo check --lib -p temps-cli
```

Both commands exited successfully. The check emitted only existing workspace/build-script notices (non-root profile, debug web placeholder, and a future-incompatibility notice).

### DIFF
--- a/crates/temps-cli/Cargo.toml
+++ b/crates/temps-cli/Cargo.toml
@@ -128,6 +128,7 @@ uuid = { workspace = true }
 bytes = { workspace = true }
 http-body-util = { workspace = true }
 libc = { workspace = true }
+tempfile = { workspace = true }
 
 [target.'cfg(not(target_env = "msvc"))'.dependencies]
 tikv-jemallocator = { workspace = true }
@@ -137,7 +138,6 @@ tikv-jemallocator = { workspace = true }
 # console health router through a single request in unit tests without binding
 # a real listener. Already present transitively; pinned here for test use.
 tower = { workspace = true }
-tempfile = { workspace = true }
 # Only for `tests/plugin_sdk_conformance.rs`: proves the SDK's hand-written
 # endpoint table still matches the handlers this crate assembles into the API.
 temps-plugin-sdk = { path = "../temps-plugin-sdk" }
--- a/crates/temps-cli/src/commands/serve/self_update.rs
+++ b/crates/temps-cli/src/commands/serve/self_update.rs
@@ -37,10 +37,10 @@ use temps_core::{
 use tracing::{error, info, warn};
 
 use crate::commands::upgrade::{
-    check_write_permission, current_version_tag, download_asset, download_asset_text,
-    extract_binary_from_tarball, fetch_latest_release_in_channel, fetch_specific_release,
-    is_newer_version, platform_target, replace_binary, verify_checksum, GitHubRelease,
-    UpgradeChannel,
+    check_write_permission, create_upgrade_temp_file, current_version_tag, download_asset_text,
+    download_asset_to_file, extract_binary_from_tarball_file, fetch_latest_release_in_channel,
+    fetch_specific_release, finalize_staged_binary, is_newer_version, platform_target,
+    verify_computed_checksum, GitHubRelease, UpgradeChannel,
 };
 
 /// Grace period between accepting the update and exiting the process. Long
@@ -783,11 +783,11 @@ impl UpdateJob {
 
     /// Download, verify, swap the binary and run database migrations.
     ///
-    /// Everything up to and including `replace_binary` is the "pre-swap" zone:
-    /// a failure there leaves the running binary untouched and is returned as
-    /// `ExecuteError::PreSwap`. Once the binary is on disk, any failure is
-    /// `ExecuteError::PostSwapMigration` — the operator is told exactly what
-    /// state the install is in and what to do next.
+    /// Everything up to and including `finalize_staged_binary` is the
+    /// "pre-swap" zone: a failure there leaves the running binary untouched
+    /// and is returned as `ExecuteError::PreSwap`. Once the binary is on
+    /// disk, any failure is `ExecuteError::PostSwapMigration` — the operator
+    /// is told exactly what state the install is in and what to do next.
     async fn execute(&self, started_at: chrono::DateTime<Utc>) -> Result<String, ExecuteError> {
         let (to_version, backup_path) = self
             .download_verify_swap()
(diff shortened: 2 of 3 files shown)


dviejokfs:
@jlucaso1 Thanks for the thorough memory work here — the streaming shape and checksum-before-extraction invariant look good.

Before merge, I found one security blocker and two hardening/test gaps:

1. **Secure temporary files:** `.temps-upgrade-dl.<pid>` and `.temps-upgrade-tmp.<pid>` are predictable, and `File::create` follows existing symlinks. Under a privileged upgrade in a lower-privileged-user-writable directory, that permits arbitrary-file overwrite / executable replacement races. Please use securely random, atomically created sibling files with retained handles (the existing `tempfile` dependency is suitable), and avoid reopening them by attacker-replaceable paths.
2. **Crash durability:** the staged file is synced before chmod, but the chmod and rename metadata are not followed by a file sync and parent-directory fsync.
3. **Regression proof:** please cover multi-chunk streaming, interrupted/error responses, and checksum mismatch leaving the live target untouched with both temporaries removed. The PR also needs pasted command output/runtime evidence rather than result assertions alone.

I’m preparing a follow-up commit on this PR branch that addresses these points and will keep you mentioned on the update.

dviejokfs:
@jlucaso1 I pushed follow-up commit d0a0dd3e to this branch.

It replaces the predictable PID scratch paths with atomically created random sibling files, retains their open handles through download/extraction/chmod/sync, atomically persists the staged inode, and fsyncs the parent directory. It also adds the missing streaming and failure-path regressions.

Verification:
- targeted upgrade tests: 61 passed
- full temps-cli library tests: 228 passed
- cargo fmt and cargo check: passed
- security re-review: signed off; the original symlink/TOCTOU blocker is resolved

I appended the exact commands and outputs to the PR Evidence section. Thanks again for the strong constant-memory foundation.

PR #808, opened 2026-08-25
perf(geo): memory-map the MaxMind databases instead of reading them into the heap

jlucaso1:
### Summary

`Reader::open_readfile` reads GeoLite2-City.mmdb (~58 MB) and GeoLite2-ASN.mmdb (~10 MB) into `Vec<u8>`, so roughly 70 MB of anonymous heap stays resident for the life of the process. On a host with swap disabled the kernel cannot reclaim any of it.

This opens both databases as read-only mappings instead.

It composes with the per-process reader memo from #721 rather than replacing it. The memo still keeps the proxy and console registries sharing one reader; the mapping makes that reader's pages file-backed. The memo's tests are unchanged and still pass.

### Measurement

Against the committed 57.9 MB GeoLite2-City.mmdb, three A/B/A runs each, swapless host (`VmSwap: 0 kB` confirmed in every sample).

At startup, immediately after the reader is constructed:

| | `VmRSS` | `RssAnon` | `RssFile` |
|---|---|---|---|
| `open_readfile` | 62,764-62,832 kB | 60,056-60,068 kB | 2,704-2,784 kB |
| `open_mmap` | 3,484-3,516 kB | 680-688 kB | 2,800-2,828 kB |

After 17,220 uniformly distributed global IP lookups, reader still alive:

| | `VmRSS` | `RssAnon` | `RssFile` |
|---|---|---|---|
| `open_readfile` | 62,780-62,852 kB | 60,056-60,068 kB | 2,716-2,784 kB |
| `open_mmap` | 46,176-46,212 kB | 680-684 kB | 45,496-45,528 kB |

The point is not only the 16.6 MB of `VmRSS` at the fully warmed worst case. It is that ~45 MB moved from `RssAnon` to `RssFile`: anonymous pages are unreclaimable without swap, clean page-cache pages are evicted and re-faulted freely. And 17,220 uniformly random global IPs is a pessimistic access pattern; real traffic concentrates on far fewer search-tree paths, so steady-state `RssFile` should sit below that.

Latency, same runs:

| | per lookup |
|---|---|
| `open_readfile` | 2,253-2,543 ns |
| `open_mmap`, warm | 2,357-2,572 ns |
| `open_mmap`, first pass over a cold mapping | 3,831 ns |

Within noise once warm. The single cold pass costs about 1.3 us per lookup while pages fault in, and geolocation is not on the request path.

### On the `unsafe`

Updated after review: the first version of this branch mapped the operator's file directly, and the safety argument covered only this repository's downloaders, which stage to a `.mmdb.tmp` sibling and rename. That was too narrow. These databases are operator-managed: `temps setup` tells operators to `cp` the file into the data directory (`commands/setup.rs`, `serve/console.rs`), and `docker-compose.yml` shows them bind-mounted from the host. A `cp` over an existing destination truncates and rewrites that inode, so a routine database refresh on a running server could give torn reads or `SIGBUS`.

It now copies the database into an anonymous temporary file, created next to it with `tempfile_in`, and maps that. The inode has no directory entry, so nothing outside the process can reach it, and the safety requirement holds regardless of what the operator does to the original. The copy is staged beside the database rather than in `env::temp_dir()`, which is a tmpfs on many hosts and would put it straight back into RAM.

Cost: one pass over the file at startup, and the file's size in scratch space while the reader lives. That buys the `RssAnon` reduction above without depending on operator behaviour.

If the private copy cannot be made, for example a read-only data directory or no space, it falls back to reading into the heap. That is the behaviour before this branch, so the fallback costs footprint and never availability.

A test maps a file, then truncates and rewrites the original inode exactly as `cp` would, and asserts the mapping still reads back the original contents.

### Unchanged

Graceful degradation is the same: the City database still fails startup with a contextual error naming the path, and the ASN database still degrades to `None` with a warning, disabling hosting-provider detection. A test pins that a missing path returns `Err` rather than panicking.

`memmap2` supports every non-MSVC target this binary is built for, and `open_mmap` exists on Windows too, so the change is not `cfg`-gated.

### Validation

- `cargo check --lib`: clean, no warnings.
- `cargo test --lib -p temps-geo`: 18 passed, 0 failed, including the four memo tests from #721 and the in-place-rewrite test described above.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update

Performance, no behaviour change intended.

## Checklist

- [ ] I have updated documentation where necessary; the safety constraint lives in a code comment

## Related issues

Builds on #721.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11545,6 +11545,7 @@ dependencies = [
  "rand 0.10.2",
  "sea-orm",
  "serde",
+ "tempfile",
  "temps-auth",
  "temps-core",
  "temps-database",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5428,6 +5428,7 @@ checksum = "a7bd67e2f5d65937ff283e418fcad53c8e2eeda0719e455508b62dad8ffe6180"
 dependencies = [
  "ipnetwork",
  "memchr",
+ "memmap2",
  "serde",
  "thiserror 2.0.20",
 ]
--- a/crates/temps-geo/Cargo.toml
+++ b/crates/temps-geo/Cargo.toml
@@ -25,6 +25,7 @@ temps-auth = { path = "../temps-auth" }
 utoipa = "5.1"
 tracing = { workspace = true }
 tokio = { version = "1.0", features = ["time"] }
+tempfile = { workspace = true }
 
 [dev-dependencies]
 tokio = { version = "1.0", features = ["full"] }
--- a/crates/temps-geo/Cargo.toml
+++ b/crates/temps-geo/Cargo.toml
@@ -12,7 +12,7 @@ anyhow = "1.0"
 axum = { workspace = true }
 chrono = { version = "0.4", features = ["serde"] }
 log = "0.4"
-maxminddb = "0.30.0"
+maxminddb = { version = "0.30.0", features = ["mmap"] }
 moka = { version = "0.12", features = ["future"] }
 rand = { workspace = true }
 sea-orm = { version = "1.0", features = ["sqlx-postgres", "runtime-tokio-rustls", "macros"] }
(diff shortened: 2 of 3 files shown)


dviejokfs:
@jlucaso1 Awesome find and implementation! merged

PR #815, opened 2026-08-26
feat(otel): harden ingest pipeline reliability and observability

dviejokfs:
## Summary

The OTLP ingest pipeline (`crates/temps-otel`) wrote to ClickHouse/TimescaleDB exactly once per batch with no retry — any transient storage blip was permanent, silent data loss, with only a bare "N dropped" counter and no visible reason. Pipeline stats were also lifetime-since-restart atomics with no time-windowed trend view. This PR closes both gaps.

- **Error classification**: storage errors are now typed transient vs. fatal (network/timeout vs. malformed data/schema mismatch) instead of one opaque `Storage { message: String }` variant.
- **Bounded retry**: spans/metrics/logs writes retry up to 3 attempts (~225ms worst case backoff) on transient failures only. Deliberately short — the write happens while the request holds a concurrency-limiting ingest permit, so a long retry would amplify `IngestSaturated` 503s for other tenants during a real outage.
- **Recorded failure reasons**: retry-exhausted failures are recorded into a new, grouped `otel_ingest_errors` Postgres table (upserted by signal + error class), surfaced via `GET /otel/ingest-errors` and an expandable drill-down on the OTel Pipeline Status settings page. Kept in Postgres rather than ClickHouse on purpose: the dominant failure this protects against is ClickHouse itself being unreachable, so writing the failure record into ClickHouse would fail for the same reason and produce an empty report exactly when it's needed.
- **Time-windowed observability**: all 13 pipeline counters (previously only 2 — rate-limited/quota-exceeded) are now published to the metrics store every 60s. New `GET /otel/pipeline-history` endpoint + a 1h/6h/24h/7d trend chart on the same settings page.
- **CLI parity**: `temps otel ingest-errors` and `temps otel pipeline-history` commands for both new endpoints, per this repo's CLI-parity requirement.

## Test plan

- [x] `cargo check --lib` (full workspace) clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean (verified independently via resolved cargo path, not through any wrapper)
- [x] `cargo test --lib -p temps-otel` — 556 passed, 0 failed (+36 new tests: error classification, retry early-exit/exhaustion, grouped-error upsert, metric-name/sampler consistency)
- [x] `cargo fmt --check` clean
- [x] `bun run spec:check` — openapi.json canonical (719 paths, +302 lines, in expected range for 2 new endpoints)
- [x] `tsc --noEmit` + `eslint` clean for both `web/` and `apps/temps-cli/`
- [x] End-to-end against a live local instance: pushed real OTLP spans through the pipeline, confirmed `pipeline-stats` → sampler → `pipeline-history` chain returns correct data; ran the retry-exhaustion path against real Postgres and confirmed the upsert collapses repeated failures into one grouped row with an incrementing count
- [x] Browser-verified the settings page: charts render real data with correct empty states, range picker works, error drill-down expands and shows grouped failure reasons with signal/reason/message/count/last-seen — zero console errors


dviejokfs:
Addressed both Greptile findings in b8462180c:

**P1 (retry replays committed writes)** — confirmed real. Investigated the "obvious" fix (unique constraint + `ON CONFLICT DO NOTHING` on `otel_spans`/`otel_metrics`/`otel_log_events`) and found it's blocked at the infrastructure level: all three are TimescaleDB hypertables space-partitioned on `id` (a fresh `nextval()` per insert attempt), and TimescaleDB requires any unique index to include every partitioning column — so a retried insert could never match on conflict even with the constraint in place. Verified directly against a live Postgres instance rather than assuming.

Since idempotent writes aren't achievable without a separate hypertable-rebuild migration (worth its own follow-up PR — the `id` space-partition looks vestigial for a single-node deployment and is the actual blocker), the fix instead narrows retry eligibility on the Postgres path to `DbErr::ConnectionAcquire` only — the one case that provably means nothing was ever sent to the server. `DbErr::Conn` (which can fire after a mid-flight commit) now drops the batch visibly instead of risking a silent duplicate. ClickHouse keeps its full retry coverage unchanged, since its `ReplacingMergeTree` tables were already designed to converge retried writes.

**P2 (missing handler tests)** — added 401/403/200/503/400 coverage for both new endpoints, reusing the existing `e2e_http_test.rs` real-Postgres harness rather than building a new one.

Also added real-Postgres characterization tests proving the old retry did duplicate rows and the new classification doesn't, plus fixed a validation-ordering bug the new tests surfaced in `get_pipeline_history` (malformed time windows were returning a generic 503 instead of a specific 400).

PR #817, opened 2026-08-26
fix(ai-gateway): scope usage summary/by-provider to the caller

### LINKED ISSUE #705: AI Gateway usage endpoints leak cross-project data to any AiGatewayRead holder
## Summary

`GET /api/ai/usage/summary` and `GET /api/ai/usage/by-provider` (`crates/temps-ai-gateway/src/handlers/usage.rs`) are gated only by `permission_guard!(auth, AiGatewayRead)`, with no ownership check tying the returned rows to projects/environments the caller actually has access to.

`Permission::AiGatewayRead` is held by default by `Role::User`, `Role::Reader`, `Role::Admin`, and `Role::PlatformAdmin` (`crates/temps-auth/src/permissions.rs`) — i.e. any authenticated project member, not just operators.

## Impact

A caller with `AiGatewayRead` (the default for ordinary project members, granted for their own day-to-day AI usage) can pass an arbitrary `project_id` or `environment_id` query parameter and read AI usage data — including `deployment_token_id` — for **any project on the instance**, not just their own. This is cross-tenant enumeration of model usage, token volumes, and deployment token IDs belonging to other tenants.

## Root cause

The handlers build their `UsageFilter` directly from query parameters (`UsageQueryParams::to_filter()`) with no scoping to `auth.user_id()` or a project-ownership check. `usage_service.rs::build_filter_clause` does support a `user_id` filter, but the handlers never populate it from the authenticated caller — filtering is entirely caller-controlled and opt-in.

## Discovery context

Found during a `/review-pr` security-audit pass on PR #543 ("feat(ai-gateway): add project usage governance"), which added the `project_id`/`environment_id` query-param wiring that makes this easy to target. **The underlying authorization gap (no ownership scoping at all) predates that PR** — the endpoint already returned the full instance-wide dataset to any `AiGatewayRead` holder when unfiltered; the PR only added a convenient way to slice that already-unscoped data by project/environment. Not blocking that PR's merge; filed separately per this repo's existing pattern for the same bug class (see #403, the analogous proxy-logs cross-project IDOR).

## Suggested fix

Either:
- Add a `require_operator`-style check (only operators may query other projects' usage), or
- When the caller is not an operator, restrict the query to `project_id IN (<projects the caller is authorized for>)` server-side, ignoring/rejecting a caller-supplied `project_id` outside that set.

BetterAndBetterII:
## Description

`GET /ai/usage/summary` and `GET /ai/usage/by-provider` previously accepted any `AiGatewayRead` caller and applied only the optional, caller-controlled `user_id` filter. An ordinary user could therefore read instance-wide aggregates or another account's usage.

This change mirrors unscoped proxy-log authorization:

- `SystemAdmin` may still run unscoped / cross-user queries
- every other caller is pinned to their own `user_id` when the filter is empty
- requesting a different `user_id` returns 403
- deployment tokens without a user identity cannot run unscoped queries

## Type of change

## Checklist

- [x] All new and existing tests pass (`cargo test -p temps-ai-gateway --lib handlers::usage::tests`)
- [ ] `cargo check --lib` passes with no warnings
- [x] My commits follow the Conventional Commits format
- [x] Every commit is signed off (`git commit -s`) per the DCO
- [ ] I have updated documentation where necessary

## Related issues

Fixes #705

### DIFF
--- a/crates/temps-ai-gateway/src/handlers/usage.rs
+++ b/crates/temps-ai-gateway/src/handlers/usage.rs
@@ -8,7 +8,8 @@ use chrono::{DateTime, Utc};
 use serde::Deserialize;
 use std::sync::Arc;
 use temps_auth::permission_guard;
-use temps_auth::RequireAuth;
+use temps_auth::permissions::Permission;
+use temps_auth::{AuthContext, RequireAuth};
 use temps_core::problemdetails::{Problem, ProblemDetails};
 use utoipa::{OpenApi, ToSchema};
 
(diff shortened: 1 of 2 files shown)


dviejokfs:
Thanks for fixing #705 — the `SystemAdmin`/self-pin approach for `scope_usage_filter_for_caller` looks correct for `get_usage_summary`/`get_usage_by_provider`.

While reviewing, I noticed the same helper wasn't wired into the other endpoints in this file that build a `UsageFilter` from caller-controlled query params: `get_usage_timeseries`, `get_usage_top_models`, `get_usage_recent`, and `get_conversations` all still let any `AiGatewayRead` caller pass `?user_id=<other>` (or omit it) to read another account's or the instance-wide AI usage data — the exact bug this PR describes, just on the other 4 of 6 filterable endpoints. `get_conversation_detail` was actually unscoped entirely (no `user_id` filter existed there at all), so any caller could read the full content of any conversation by ID.

Since you'd enabled "allow edits from maintainers," I pushed a commit (`43f683e`) extending your `scope_usage_filter_for_caller` pattern (factored into a small `resolve_caller_scope` helper) to the remaining 4 handlers, and threaded the resolved scope into `UsageService::get_conversation_detail`'s SQL so it's also ownership-checked. Added unit tests for the new scope resolution; all 224 tests in `temps-ai-gateway` pass, and `cargo check --lib` / `cargo clippy -p temps-ai-gateway --lib -- -D warnings` are both clean.

Hope that's alright — happy to adjust if you'd prefer a different approach or want to split it into a separate PR instead. Thanks again for catching and fixing the original issue!

BetterAndBetterII:
Thanks for catching the other four endpoints and conversation detail, and for pushing the follow-up. `resolve_caller_scope` is the right split — looks good to me.

PR #843, opened 2026-08-30
feat(deployments): filter and limit environment container metrics by deployment

dviejokfs:
## Summary
- Environment metrics charts fetched every container an environment ever ran, unbounded — long-lived environments fan that out into hundreds of concurrent per-container metrics requests, and there was no way to isolate a single deployment's containers while debugging a CPU/memory regression.
- `GET /projects/{project_id}/environments/{environment_id}/container-history` now accepts `deployment_id` and `limit` (default 20, max 100) query params, returns rows newest-first, and includes `total_count` so the client can show how much was truncated (e.g. "20 of 627").
- Wired a deployment filter dropdown and a limit selector into the environment Metrics tab (`EnvironmentMetricsCard.tsx`).
- Added CLI parity: `temps containers history` gained `-d/--deployment-id` and `-l/--limit`.
- Regenerated both the web SDK and `@temps-sdk/cli` OpenAPI client/spec (`bun run openapi-ts`, `spec:update` + `generate:api`); `openapi.json`'s diff also includes unrelated pre-existing drift already in this branch's source (a `silenceAlarm` endpoint from a prior merge that had never been resynced into the committed spec).

## Test plan
- [x] `cargo test --lib -p temps-deployments` — 668 passed, including a new integration test covering deployment filtering, limit capping with correct `total_count`, and the cross-environment 404 guard for `deployment_id`.
- [x] `cargo check --lib` (full workspace) — clean.
- [x] `cargo clippy --workspace --all-targets -p temps-deployments -- -D warnings` — clean (via `$(command -v cargo)`, not through the rtk proxy).
- [x] `bun run tsc --noEmit` in `web/` and `apps/temps-cli/` — clean.
- [x] `bunx eslint` on the changed frontend file — clean.
- [x] `apps/temps-cli`: `bun run spec:check` — canonical.
- [x] `python3 scripts/source_attribution.py check` — passed for all source files.
- [ ] Manual browser walkthrough of the new deployment filter/limit dropdowns against a real deployed environment — not done in this pass; the backend logic is covered by the new integration test, but UI interaction wasn't clicked through live.


dviejokfs:
Addressed all three P1 findings from the Greptile review in 2828d5c6c:

1. **Deployment selector truncates history** — switched the deployment filter dropdown to `useInfiniteQuery` with an explicit "Load more deployments…" control, so environments with more than 100 deployments can still reach the older ones instead of only seeing the newest page.
2. **Limit truncates active containers** — `list_environment_container_history` now fetches all currently-running containers uncapped, and applies `limit` only to the replaced/historical containers layered on top. A live container can no longer be silently dropped from the running count or its metrics just because `limit` was small.
3. **Post-sort reverses response order** — removed the ascending re-sort after the DB's `ORDER BY deployed_at DESC LIMIT n`; the response now genuinely puts current containers first, then the newest-replaced ones, matching the documented contract (updated wording to describe the current-first/replaced-second order precisely, since "limit" is no longer a single flat cap).

Also updated the integration test to assert the running-containers-are-never-truncated behavior directly, and refreshed the OpenAPI doc comments / CLI flag description / both generated clients to match.

PR #847, opened 2026-08-30
fix(proxy): fail fast with an actionable error when port 80/443 are already in use

dviejokfs:
## Summary

- A user reported that a local port 80/443 conflict caused domain HTTPS provisioning to fail with no clear recovery path.
- Root cause: `setup_proxy_server` (`crates/temps-proxy/src/server.rs`) hands control to Pingora's `Server::run()` with no bind-failure handling. Pingora's own listener setup (`pingora-core`'s `Service::start_service`) does `.expect("Failed to build listeners")` **inside a spawned tokio task**, and a panicking spawned task is swallowed silently by `futures::future::join_all` — the process keeps running with that one listener simply never up. A port-80 conflict therefore didn't crash the proxy or log anything actionable; it just meant HTTP-01 challenges started timing out against a host that never answered, several minutes later, with nothing pointing back at "port 80 was already taken at startup."
- Fix: bind-and-immediately-drop both configured addresses (`proxy_config.address`, `proxy_config.tls_address`) *before* handing control to Pingora. A conflict now returns a normal `anyhow` error naming the address, the underlying OS error, and how to recover — which the two existing callers (`temps-cli/src/commands/serve/proxy.rs`, `temps-cli/src/commands/proxy.rs`) already log with `error!()` and exit on.
- Follow-up commit (caught in self-review): the first pass treated every bind failure as "another process has this port." Binding to ports below 1024 without root/`CAP_NET_BIND_SERVICE` fails with `PermissionDenied`, not `AddrInUse` — a very common case for `temps serve` run as a non-root user against the default 80/443. That's now a distinct message pointing at the actual fix (grant the capability, or use a port >= 1024) instead of sending the operator hunting for a process that doesn't exist.
- Note on the on-demand ACME/domain-cert side (also part of the original report): that layer already tracks and exposes failure state well — `domains.status`/`last_error`/`last_error_type`/`on_demand_backoff_until`, the `GET /domains/by-host/{hostname}/cert-status` and `GET /domains/on-demand-certs` endpoints, the web "Certificates" page, and automatic re-enqueue on the next TLS handshake once backoff expires — so no changes were needed there. This PR is scoped to the actual unaddressed gap: the silent proxy-startup bind failure.
- Security-reviewed (security-auditor agent): sign-off, no findings — bootstrap-only code, operator-controlled input, no network-facing surface, no injection sinks.

## Test plan

- [x] `cargo check --lib` (whole workspace) — clean
- [x] `cargo fmt` / `cargo clippy -p temps-proxy --lib --tests -- -D warnings` (invoked directly, not through the rtk wrapper) — clean
- [x] Regression tests reproduce the exact symptom (a taken port fails with an actionable message instead of silently vanishing into Pingora) and the permission-denied vs. port-in-use distinction:

```
running 4 tests
test server::preflight_bind_tests::bind_failure_message_explains_permission_denied_separately_from_port_taken ... ok
test server::preflight_bind_tests::check_address_bindable_fails_with_actionable_message_when_port_taken ... ok
test server::preflight_bind_tests::check_address_bindable_succeeds_on_free_port ... ok
test server::preflight_bind_tests::preflight_check_listen_addresses_reports_tls_port_conflict ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 570 filtered out; finished in 0.00s
```

### DIFF
--- a/crates/temps-proxy/src/server.rs
+++ b/crates/temps-proxy/src/server.rs
@@ -626,16 +626,39 @@ fn preflight_check_listen_addresses(proxy_config: &ProxyConfig) -> Result<()> {
 fn check_address_bindable(address: &str, env_var: &str) -> Result<()> {
     std::net::TcpListener::bind(address)
         .map(|_| ())
-        .map_err(|e| {
-            let port = address.rsplit(':').next().unwrap_or(address);
-            anyhow::anyhow!(
-                "Cannot start the proxy: failed to bind to {address}: {e}. Another process on \
+        .map_err(|e| anyhow::anyhow!(bind_failure_message(address, env_var, e.kind(), &e)))
+}
+
+/// Render an actionable message for a failed bind, distinguishing "something
+/// else already has this port" from "we're not allowed to bind this port" --
+/// binding to a port below 1024 without root/`CAP_NET_BIND_SERVICE` fails
+/// with `PermissionDenied`, not `AddrInUse`, and is common enough (`temps
+/// serve` run as a non-root user against the default 80/443) that lumping it
+/// into "another process is listening" would send the operator chasing a
+/// process that doesn't exist.
+fn bind_failure_message(
+    address: &str,
+    env_var: &str,
+    kind: std::io::ErrorKind,
+    err: &dyn std::fmt::Display,
+) -> String {
+    if kind == std::io::ErrorKind::PermissionDenied {
+        format!(
+            "Cannot start the proxy: not permitted to bind to {address}: {err}. Binding to \
+             ports below 1024 usually requires root or `CAP_NET_BIND_SERVICE` (Linux: \
+             `sudo setcap 'cap_net_bind_service=+ep' $(which temps)`). Either run Temps with \
+             that capability, or set {env_var} to a port >= 1024 and restart."
+        )
+    } else {
+        let port = address.rsplit(':').next().unwrap_or(address);
+        format!(
+            "Cannot start the proxy: failed to bind to {address}: {err}. Another process on \
              this machine is already listening on that port -- find it with \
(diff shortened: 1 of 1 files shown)


dviejokfs:
Addressed a P1 finding from review: **duplicate listeners bypassed the preflight check.**

`TEMPS_ADDRESS` and `TEMPS_TLS_ADDRESS` are two independent CLI args (`serve/mod.rs:63-68`, `proxy.rs:160-165`) with no validation anywhere against them being set to the same value. The original preflight checked each address independently — bind, then immediately drop — so if both were identical, the first probe released the port before the second one bound it. Both checks passed even though Pingora binds both listeners *concurrently*, so the real conflict was still hitting the silently-swallowed `EADDRINUSE` path this PR exists to fix.

Fixed by holding both listeners open simultaneously for the duration of the check (mirroring what Pingora actually does), instead of checking each address in isolation. Added a regression test (`preflight_check_listen_addresses_rejects_identical_http_and_tls_addresses`) that reproduces the exact scenario and fails against the old implementation.

```
running 5 tests
test server::preflight_bind_tests::bind_failure_message_explains_permission_denied_separately_from_port_taken ... ok
test server::preflight_bind_tests::check_address_bindable_succeeds_on_free_port ... ok
test server::preflight_bind_tests::preflight_check_listen_addresses_rejects_identical_http_and_tls_addresses ... ok
test server::preflight_bind_tests::preflight_check_listen_addresses_reports_tls_port_conflict ... ok
test server::preflight_bind_tests::check_address_bindable_fails_with_actionable_message_when_port_taken ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 570 filtered out; finished in 0.00s
```

`cargo fmt` / `cargo clippy -p temps-proxy --lib --tests -- -D warnings` (direct, not via rtk) and workspace-wide `cargo check --lib` are clean.

PR #857, opened 2026-08-31
feat(routes): add live Traefik docker-label discovery, with opt-in TLS for discovered routes

dviejokfs:
## Summary

- Watches a Docker network for `traefik.enable=true` + `traefik.http.routers.*` labels and syncs matching containers into `traefik_discovered_routes`, so Temps' own Pingora proxy routes to containers it did **not** deploy — no changes required to an existing docker-compose / Coolify / Dokploy stack, no Traefik binary needed.
- Discovered routes are persisted as DB rows (not injected in-memory), so the existing `route_table_changes` NOTIFY -> `load_routes()` machinery propagates changes in-process and across control-plane nodes with no bespoke fan-out.
- A Docker events subscription gives immediate per-container updates; a 30s reconciliation pass is the safety net for missed events and for state after a restart.
- Opt-in via `TEMPS_TRAEFIK_DISCOVERY_ENABLED` / `TEMPS_TRAEFIK_DISCOVERY_NETWORK`, **off by default**.
- Adds the operator-facing surface required for any new capability per this repo's discoverability rule: `GET /traefik-discovery/status` (configured/enabled, watched network, last reconciliation outcome), a routes list with host-collision conflicts visible, a per-route enable/disable toggle, a **settings UI page** (`/settings/traefik-discovery`), and matching `traefik-discovery` CLI commands in `@temps-sdk/cli`.

## TLS certificate handling for discovered routes

Discovered routes serve HTTP-only by default (`cert_eligible` stays false) — attaching HTTPS to a migrated Traefik domain is a separate, explicit, opt-in step so a domain doesn't silently start terminating TLS with unverified key material. Two paths, both gated behind the console's Admin-only `SettingsWrite` + `DomainsCreate` permissions:

- **Path A — request**: issue a certificate through Temps' existing ACME/`DomainService`, the same system that already issues certs for deployments.
- **Path B — import**: import an existing cert+key (e.g. from Traefik's `acme.json`) through an eight-step validation chain — leaf definition pinned to the first PEM `CERTIFICATE` block, key loaded through the real `rustls` `CryptoProvider` and matched against the leaf via a live sign/verify (not a naive PEM-decode "match"), non-certificate PEM blocks (e.g. a smuggled private key) rejected before persistence, chain-length capped, host ownership re-checked against every existing domain-owning table at request time — so this never trusts the uploaded material beyond what its own X.509 content and a live signature match prove.

Authorization is tracked in a new host-keyed `traefik_route_certificates` table, independent of the discovered route's container lifetime: if a different container starts winning a contested hostname after a certificate was authorized, that's now a Critical alarm plus a persistent console banner naming both containers, and the authorization is **never auto-cleared** — clearing it wouldn't stop the impersonation (the `domains` row that actually serves TLS doesn't consult this table), and would hand a hostile container a one-pass way to break a legitimate host's renewals.

Also fixes a pre-existing bug: certificates issued through the current ACME path were mislabeled with a `verification_method` value the renewal scheduler didn't recognize, so they silently stopped renewing. Fixed as three sequenced steps (map known aliases → stop emitting the mislabel + backfill existing rows → only then alarm on genuinely unrecognized values) specifically to avoid a false alarm-storm on every existing instance.

Design in `docs/adr/041-discovered-route-tls-certificate-handling.md`, approved after two rounds of security review. The implementation itself went through a further two rounds of code review — the first found real gaps (the container-drift alarm only logging a `warn!` instead of reaching the alert pipeline, and the host-ownership check existing in neither the service nor the provisioner trait) — both fixed and independently re-verified against source before this push.

## Security

Went through a full pre-merge security review, which initially **withheld sign-off** on three HIGH findings; a dedicated fix pass closed all three, and a follow-up review pass **granted sign-off**. Fixed:

- **SSRF via unvalidated port label**: a `loadbalancer.server.port` label is now only honoured when the container actually exposes that port — previously an unvalidated label could point the proxy at an arbitrary port on the Temps host (e.g. a colocated database or the control plane itself) in baremetal deployments.
- **Cluster-wide reload amplification**: the Docker events loop now tracks which container IDs currently have discovered rows, so unrelated container churn on the host no longer costs a DB round trip; the table's insert/delete trigger is row-level (was statement-level), so a zero-row statement can't fan out a route-table reload to every control-plane node.
- **Non-working kill switch**: discovered routes are now scoped to the currently-configured network at load time (in both monolith and split-proxy topologies), and foreign-network rows are pruned once stale — disabling or repointing discovery actually stops serving what it previously adopted, instead of leaving it live forever.
- Also fixed as a quick, safe-default one-liner: a container's `traefik...tls` label is recorded but never triggers ACME certificate issuance by itself — see the TLS section above for the actual (separately gated) opt-in path that now exists.

The TLS certificate-handling work above went through its own two rounds of security review on the design (ADR-041) plus two rounds of code review on the implementation — see that section for what each round found and fixed.

**Known-open, deliberately deferred** (documented, not blockers — see review): lexicographic container-name tie-break on contested hostnames, `.temps.local` internal FQDNs and paused-deployment hostnames aren't yet in the reservation list, O(n²) label parsing has no cap on router count per container, `touch_last_seen` isn't network-scoped and has no Postgres param-count chunking, `normalize_host` accepts IP literals / single-label hostnames, and the renewal scheduler's Critical-alarm-on-unrecognized-`verification_method` step (ADR-041 §7a step c) is intentionally not yet enabled — it ships once the backfill migration above has run on an instance, per the ADR's own phased migration plan, to avoid a false alarm-storm during the rollout window. None of these regressed from either fix pass; all are real but lower-severity or deliberately sequenced follow-up work.

Containers already carrying `sh.temps.deploy_id` (i.e. already Temps-managed) are always skipped, and a discovered route never displaces a deployment, custom route, custom domain, environment subdomain, or the console hostname — verified by both unit and real-Docker/Postgres integration tests, and independently re-verified by security review.

## Evidence

**Discovery feature — unit + integration tests, real Docker + real Postgres, ~245+ tests across the touched crates:**

```bash
cargo test --lib -p temps-deployer -p temps-routes -p temps-migrations -p temps-deployments -p temps-cli
```
```text
temps-deployer:     243 passed
temps-routes:        81 passed (Dockerized Postgres)
temps-migrations:    10 passed
temps-deployments:   22 passed
temps-cli:           355 passed
```

```bash
cargo test -p temps-deployer --test traefik_discovery_integration   # real Docker containers, real network
cargo test -p temps-deployments --test traefik_discovery_admin_test
```
```text
8 passed (discovery, container-stop removal, Temps-owned-container skip, host-collision precedence, kill-switch)
3 passed (admin API)
```

**Manual local verification** against a real `temps serve` instance with discovery enabled and a live `nginxdemos/hello` container carrying Traefik labels on the watched network:
- Startup scan discovered the labeled container immediately.
- Killing the container fired `action=die removed=1` and dropped the route instantly — no waiting for the 30s reconciliation poll.
- `curl -H "Host: demo.localho.st"` through the Temps proxy returned the real container's response, genuinely routed through Pingora.
- Correctly used the container's *published* host port in baremetal mode (not the internal container port).
- Unmatched hosts still fall through to the console unaffected.

**TLS certificate handling — unit test coverage on the validation chain, drift detection, and renewal dispatch:**

```bash
cargo test --lib -p temps-deployments -p temps-deployer -p temps-domains
```
```text
temps-deployments: 724 passed, 3 ignored
temps-deployer:     358 passed
temps-domains:       68 passed, 3 ignored
```

Coverage includes the adversarial cases from the ADR's verification table: a chain where element 0 covers an attacker host is rejected (leaf is pinned to element 0 only); a `PRIVATE KEY` block smuggled into the `certificate` field is rejected before persistence; zero/11+ certificate blocks and zero/two key blocks are rejected; SEC1 (`BEGIN EC PRIVATE KEY`) and PKCS#1 (`BEGIN RSA PRIVATE KEY`) — the encodings Traefik/lego actually write — are accepted with the sign/verify match demonstrably run through the real `rustls` `CryptoProvider`; a mismatched key is rejected; container-drift detection sets `container_drift_detected_at`, fires a Critical alarm through `AlarmService` (verified end-to-end via a hand-written `DriftAlarmSink` bridge, not just a log line), and leaves `cert_authorized` set; drift dedup suppresses repeat alarms for the same container and re-fires when a third container takes over; renewal dispatch routes `verification_method="acme"` to the HTTP-01 path (not an alarm) and `"manual"` to the manual-notification path (not `TlsRenewalFailed`).

**Manual local verification** against a real `temps serve` instance rebuilt fresh from this branch (including a regenerated `openapi.json`, CLI SDK, and web SDK against the live server, not hand-edited):
- `POST /api/traefik-discovery/tls/import` and the Path A certificate endpoint both exist and validate request bodies (422 on a missing required field, not 404).
- The `/settings/traefik-discovery` console page renders correctly in both the unconfigured/onboarding state and the enabled/running state (status card, discovered-routes table), with no console errors.
- `bunx tsc --noEmit` passes clean in both `apps/temps-cli` and `web/` against the freshly regenerated SDKs.

**Compilation and linting (full workspace, after merging `origin/main`):**

```bash
cargo build --profile fast --bin temps --package temps-cli   # 155 crates, includes the web UI build
$(command -v cargo) clippy --workspace --all-targets -- -D warnings
```
```text
cargo build: 0 errors
cargo clippy: 0 errors, 0 warnings (real cargo binary, not rtk)
Attribution check passed
```

## Review pass (post-push)

Ran a full `/review-pr`-style pass before merge: `context-scout` for blast radius, a dedicated `security-auditor` pass, a Rust-standards + migration-safety pass, and a frontend/SDK pass, all independently against the actual pushed diff.

**Security: APPROVED WITH MINOR NOTES.** All explicitly-checked properties confirmed clean by re-reading the actual code (not the PR's own claims): `Certificate`'s hand-written `Debug` redaction, `ImportTraefikAcmeJsonRequest` not serializing/debug-printing the raw `acme_json`, the TLS endpoints requiring `SettingsWrite` + `DomainsCreate`, the SSRF-blocking two-layer port validation, host-ownership re-checked fresh from the DB at request time, the cert/key match using a real sign/verify with no accept-anyway branch, and the container-drift alarm genuinely reaching `AlarmService.fire_alarm()` at Critical severity. Two minor findings (a narrow ownership-check gap, an unbounded host-count loop on the import endpoint) — both admin-only and low impact — fixed in the follow-up commit anyway.

**Rust standards + frontend/SDK: found one real blocking bug**, since fixed. The TLS provisioner dependency (`DiscoveredHostTlsProvisioner`) was declared but never wired to a real implementation anywhere in the binary — `with_provisioner()` was defined but never called, so `authorize_acme_cert` and `import_acme_json` each gated the actual provisioning call behind an `if let Some(prov)` with no `else`. With the provisioner always `None` in production, both endpoints silently fell through and reported success (`cert_authorized: true`, `success: true`) without ever requesting or persisting a certificate. Fixed: the provisioner is now a required dependency that fails loudly at startup if missing (per this repo's `Option<T>` rule), wired to a real adapter bridging to the existing `DomainService`/`CertificateRepository`, with new tests specifically covering the "no provisioner" and "provisioner fails" cases that would have caught this.

Also fixed in the same follow-up commit: an N+1 query and an unnecessarily-repeated 5-table scan in the discovery reconciler's hot loop, missing `ProblemDetails` response bodies across all six endpoints, a stale hand-rolled CLI command (`tls.ts`) that had a comment claiming the generated SDK didn't have the functions it needed — it did, from this same PR — now using the real generated client, missing console remedy actions for the "HTTPS will fail" warning state (ADR-041's own console requirement), a `#[serde(skip_serializing)]` that was silently dropping `acme_json` from the generated OpenAPI schema, plus several smaller correctness/coverage fixes.

**Verification after the fix pass** (full workspace, from scratch):
```text
cargo check --lib --workspace                          : 0 errors (66-1016 crates depending on cache state)
cargo clippy --workspace --all-targets -- -D warnings   : 0 errors, 0 warnings (real cargo binary, not rtk)
cargo test --lib -p temps-deployer -p temps-deployments -p temps-domains
  temps-deployer:     368 passed, 0 failed
  temps-deployments:  743 passed, 0 failed, 3 ignored (pre-existing, requires Docker/Pebble)
  temps-domains:       68 passed, 0 failed, 3 ignored (pre-existing, requires Pebble)
bun run typecheck (apps/temps-cli)                      : clean
bunx tsc --noEmit (web/)                                : clean
python3 scripts/source_attribution.py check              : passed, 3335 files
```
Both OpenAPI clients regenerated against a live rebuilt server (not hand-edited) — confirmed `acme_json` is now present and required in `ImportTraefikAcmeJsonRequest`, and every non-2xx response across all six `traefik-discovery` endpoints now types as `ProblemDetails`.


dviejokfs:
@greptile-apps

PR #871, opened 2026-09-01
fix(web): stale deploy status badge and Load Balancer heading mismatch

dviejokfs:
## Summary
- The project header's "Deployed"/"Not deployed" badge read `project.last_deployment`, a field fetched once on page load and never invalidated after a deploy mutation succeeds — so a freshly-deployed, healthy, current project could show "Not deployed" indefinitely until a hard reload. Switched it to the already-fetched, actively-polling `lastDeployment` query that's already passed into the same component.
- The Load Balancer page's own heading read "Routes Management" while its breadcrumb reads "Routes" — updated to match.

Both found and verified live (before/after screenshots) during a UI consistency sweep of the console, run against a fresh checkout with a deployed demo project.

## Test plan
- [x] `bunx tsc --noEmit` in `web/` passes clean
- [x] Verified live: header badge shows "Deployed" on a fresh page load after a completed deployment (previously showed "Not deployed")
- [x] Verified live: Load Balancer page heading reads "Routes", matching its breadcrumb

### DIFF
--- a/web/src/components/funnel/FunnelManagement.tsx
+++ b/web/src/components/funnel/FunnelManagement.tsx
@@ -34,6 +34,7 @@ import { BarChart3, Calendar as CalendarIcon, Plus } from 'lucide-react'
 import * as React from 'react'
 import { DateRange } from 'react-day-picker'
 import { useNavigate } from 'react-router'
+import { useKeyboardShortcut } from '@/hooks/useKeyboardShortcut'
 import { FunnelCard } from './FunnelCard'
 
 interface FunnelManagementProps {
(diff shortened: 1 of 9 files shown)


dviejokfs:
Good catch — fixed in e78ca6c05. `lastDeployment`'s truthiness alone doesn't check status, so a pending/running/failed deployment would also read as "Deployed". Now checks `lastDeployment.status` against `completed`/`deployed`, matching exactly what `mark_deployment_complete.rs` uses to stamp `project.last_deployment` on the backend.

dviejokfs:
Added a follow-up commit unifying the empty-state pattern across four more project-settings pages (Secrets, Domains, Deployment Tokens, backup schedule's attached-services list) — they were each using their own bespoke bordered-box-with-text markup instead of the shared `EmptyState` component the rest of the app's empty states already use. Left one borderline case (a compact nested section in EmailDomainDetail) unchanged since the bigger icon treatment would likely look disproportionate there — noted for a human design call.

dviejokfs:
Added a follow-up commit: the new empty-state create buttons (Domains, Secrets) now get a working `N` shortcut, wired through the shared `useKeyboardShortcut` hook.

While doing that I found the codebase already had three parallel implementations of "N opens create": the shared hook, the `CreateActionButton` wrapper, and several hand-rolled `document.addEventListener('keydown', ...)` blocks in SecretsSettings/EnvironmentVariablesSettings/FunnelManagement — each reimplementing the input-focus guard but missing the shared hook's guard against firing while a dialog/menu is open on top of the page. Replaced all three with the shared hook (EnvironmentVariablesSettings actually had two redundant copies of the same hand-rolled handler; removed the dead one). Verified live that N still opens/navigates correctly on all four pages.

Left one third pattern (`components/ui/keyboard-shortcut.tsx`, used once in ProjectDeployments.tsx) unrefactored — it's functionally correct today, just architecturally inconsistent, and a broader consolidation felt out of scope for this PR.

dviejokfs:
Addressed the follow-up: `get_last_deployment` returns the most-recent-by-`created_at` deployment, not necessarily the live one — `is_current` is computed separately in `services.rs` from each environment's `current_deployment_id`. A completed-but-superseded deployment (e.g. after a rollback) was still passing the status check and rendering "Deployed".

Fixed in `ProjectDetailHeader.tsx` by additionally requiring `lastDeployment?.is_current` before showing the badge as deployed.

Also unified the Traces OpenTelemetry setup page onto the same `SetupWizardShell` component Errors/Analytics already use (3-step Framework → Install → Verify, with a polling "waiting for first trace" step) — was flagged separately as a UI consistency issue in the same sweep.

PR #905, opened 2026-09-04
feat(deployments): diagnose Docker Hub rate limits, add registry mirror prefix

dviejokfs:
## Summary

Autopack's generated Dockerfiles (and every other image pull Temps issues) resolve unqualified base images like `node:22-slim` against docker.io with no credentials. Docker Hub throttles anonymous pulls by source IP, which is worse for multi-node/multi-cluster operators sharing one WAN IP. Today this surfaces as an opaque pull failure in build logs, indistinguishable from a broken Dockerfile.

- **Diagnose it**: recognize the Docker Hub rate-limit signature (`toomanyrequests`, BuildKit's `429 Too Many Requests` authorize failure) in build failures and explain the fix instead of surfacing the raw pull error.
- **Docs**: `docs/howto/configure-a-docker-registry-mirror` walks through the zero-code-change fix most operators want — `registry-mirrors` in `/etc/docker/daemon.json` — including how to layer an authenticated pull-through cache for a higher pull ceiling.
- **Registry mirror prefix setting**: for operators whose internal registry is a path-prefixing reverse proxy rather than a `registry-mirrors`-compatible cache (so the daemon-level fix doesn't apply), an opt-in `registry_mirror_prefix` platform setting rewrites implicit Docker Hub `FROM` lines in generated Dockerfiles through the configured prefix.
  - Shared detection/rewrite logic lives in `temps-core::registry_prefix` (`is_docker_hub_image`, `qualify_with_registry_prefix`), reused by the Dockerfile-line rewriter in `temps-presets::registry_prefix` (which tracks multi-stage `AS <name>` stage references so they're never mistaken for a real image, and skips `FROM ${ARG}` build-arg references).
  - Applied centrally in `BuildImageJob::ensure_dockerfile`, so it covers every preset's generated Dockerfile, not just autopack's.
  - An existing Dockerfile already committed in a repo is never touched — the rewrite only runs on preset-generated content.
  - Settings UI: new "Registry Mirror / Prefix" card on the Docker Registry settings page, clearly distinguished from the existing (unrelated) private-registry-credentials section, with an inline explanation of when to use each.
  - A malformed prefix (embedded control characters, shell metacharacters — anything outside a registry host+path's character set) is rejected with 400 at the settings API, and independently never spliced into a Dockerfile even if it somehow reached build time — see the security fix noted below.

## Test plan

- [x] `cargo test --lib -p temps-core -p temps-presets -p temps-deployments -p temps-config` — 750+ passed
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `bunx tsc --noEmit` (web/, apps/temps-cli) — clean
- [x] `bunx eslint` on touched frontend files — clean (one pre-existing unrelated lint error left untouched)
- [x] `bun test src/api/platformSettings.test.ts` — passing
- [x] Regenerated `apps/temps-cli/openapi.json` and `web/src/api/client/` against a live server — `bun run spec:check` confirms canonical shape
- [x] `crates/temps-presets/tests/registry_prefix_docker_build.rs`: two real `docker build` integration tests proving the rewritten `FROM` line survives BuildKit parsing and resolves — one against a public mirror (`mirror.gcr.io`), one against a local self-hosted registry via testcontainers (closer to the actual reverse-proxy scenario this feature targets)
- [ ] Manual verification: configure a prefix, deploy an autopack app, confirm the built image's `FROM` lines are rewritten and the build still succeeds

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12384,6 +12384,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "tempfile",
+ "temps-core",
  "temps-entities",
  "thiserror 2.0.20",
  "tokio",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12386,6 +12386,7 @@ dependencies = [
  "tempfile",
  "temps-core",
  "temps-entities",
+ "testcontainers",
  "thiserror 2.0.20",
  "tokio",
  "toml 1.1.4+spec-1.1.0",
--- a/apps/temps-cli/openapi.json
+++ b/apps/temps-cli/openapi.json
@@ -3100,6 +3100,14 @@
               }
             ]
           },
+          "registry_mirror_prefix": {
+            "default": null,
+            "description": "Prefix applied to Docker Hub base images generated for a build (e.g.\nautopack's `FROM node:22-slim`), turning them into\n`{prefix}/node:22-slim`. Unlike `docker_registry` above — which\nauthenticates pulls to one *named* private registry a user's own image\nreference already points at — this rewrites Temps' own generated,\notherwise-anonymous `docker.io` references, for operators whose\ninternal registry is a path-prefixing reverse proxy rather than a\n`registry-mirrors`-compatible pull-through cache (which needs no\nrewriting at all — see docs/howto/configure-a-docker-registry-mirror).\n`None`/empty (the default) leaves every reference untouched.",
+            "type": [
+              "string",
+              "null"
+            ]
+          },
           "request_timeouts": {
             "default": {
               "default_http_timeout_seconds": 0,
@@ -3304,6 +3312,13 @@
           },
           "rate_limiting": {
             "$ref": "#/components/schemas/RateLimitSettings"
+          },
+          "registry_mirror_prefix": {
+            "description": "Prefix applied to implicit Docker Hub base images in generated\nDockerfiles (e.g. autopack's `FROM node:22-slim`). No sensitive\ncontent, passed through as-is. `None`/empty disables rewriting.",
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "request_timeouts": {
             "$ref": "#/components/schemas/RequestTimeoutSettings",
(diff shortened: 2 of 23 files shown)


dviejokfs:
Pushed a follow-up commit addressing review feedback:

- **Security (HIGH, blocking):** the previous whitespace-trim fix for `registry_mirror_prefix` only stripped leading/trailing whitespace — an embedded newline still passed through and would splice a syntactically independent Dockerfile instruction into every preset-generated build. Added a strict allowlist validator (`temps_core::registry_prefix::is_valid_registry_prefix`) enforced at two layers: the settings write path now rejects a malformed prefix with 400, and `qualify_with_registry_prefix` independently fails closed (falls back to the unmodified image) if a bad value ever reaches build time regardless of how it got there. New tests cover embedded-newline, CRLF, NUL, and shell-metacharacter payloads.
- **Evidence:** added `crates/temps-presets/tests/registry_prefix_docker_build.rs`, a real `docker build` integration test proving the rewritten `FROM` line survives BuildKit parsing and resolves through a public mirror (`mirror.gcr.io`), not just a string-transform unit test.

`cargo test -p temps-core -p temps-config -p temps-presets -p temps-deployments` (750+ tests) and `cargo clippy --workspace --all-targets -- -D warnings` both pass clean.

`apps/temps-cli/openapi.json` regeneration is still outstanding — the shared dev DB used for that workflow is currently crash-looping in my environment, unrelated to this change. Will follow up once it recovers.

dviejokfs:
Two more follow-ups pushed:

- **openapi.json / SDK regeneration**: the shared dev DB recovered, so I stood up an isolated throwaway server (own DB, own ports, outside the usual dev slot ranges) and regenerated both `apps/temps-cli/openapi.json` and `web/src/api/client/` for real. The hand-edited `registry_mirror_prefix` type from the earlier commit matched byte-for-byte what codegen produces. The diff also picks up some pre-existing doc-comment drift from other features merged since the committed spec was last regenerated — unrelated to this PR, but proportional (`bun run spec:check` confirms canonical shape).
- **Registry testcontainer coverage**: added `a_registry_mirror_prefixed_dockerfile_builds_against_a_self_hosted_registry` using `testcontainers` + the real `registry:2` (OCI distribution) image, seeding it with exactly the base image(s) the rewrite produced and running a real `docker build` against it — no network dependency on a third party, and closer to the actual "operator's own path-prefixing proxy" scenario than the existing `mirror.gcr.io` test. I did not use full Harbor for this: Harbor is a multi-container product (core/db/redis/portal) whose own push/pull path-routing is already well-tested upstream, and the code risk here is entirely in Temps' string rewrite + whether the result is a resolvable reference against a real distribution-spec registry — which `registry:2` proves just as well, far lighter and faster for CI. Happy to add a full Harbor-backed test too if that's wanted for extra confidence, but I don't think it buys additional coverage of this PR's own code.

`cargo test -p temps-presets` (both integration tests) and clippy pass clean.

PR #907, opened 2026-09-04
fix(providers): pin WAL-G archive source per Postgres service

dviejokfs:
## Summary

Root-causes the "WAL segment ... is not ready to mirror" warnings seen in
production on services with more than one S3 source configured:

- Continuous archiving (`archive_command`) was only ever guarded by a
  **presence** check (`walg.env` exists at all), never by *which* source it
  points at. Once established from any one backup run, archiving stayed
  pinned to that source forever, silently, even as later scheduled base
  backups uploaded to a different source. WAL-G restore/mirroring requires
  the base backup and its WAL segments under the same S3 prefix, so once
  they diverged, mirroring for that service could never succeed.
- Adds `external_services.walg_archive_s3_source_id` /
  `walg_archive_pinned_at` as the persisted, single source of truth for
  where a Postgres service's WAL-G archiving points. Defaults to the
  instance's Cloud-managed S3 source on first use when one exists.
- A backup run that requests a different source than the current pin is now
  **rejected** (`PermanentFailure`, actionable error message), not silently
  honoured — switching sources is a deliberate, audited operation
  (`ExternalServiceManager::repoint_walg_archive_source`, exposed via
  `POST /external-services/{id}/walg-archive-source`), not something a
  schedule change should be able to do by accident.
- On Cloud's mirror side, a backup whose base snapshot predates the
  service's current archive pin is now classified as permanently
  `Unsupported` (its WAL was written to a source that's no longer tracked
  and can never appear here) instead of retrying forever.
- Removes the now-fully-redundant `retry_required` sweep-wide backoff signal
  in `backup_mirror::sweep` — `mirror_retry_delay` already gives every
  `Retry`-classified backup its own bounded per-row exponential backoff via
  `DUE_BACKUPS_SQL`, so the shared flag (which degraded the *whole* sweep's
  interval whenever any single backup needed a retry) added nothing.

## CLI parity

`temps services repoint-walg-source --id <id> --s3-source <id>`

## Test plan

- [x] `cargo test --lib -p temps-backup postgres_walg` — 7/7 pass, including
      4 new tests covering: matching pin is a no-op, mismatched pin is
      rejected not silently moved, first backup with no managed source pins
      to the requested source, first backup against the wrong source is
      rejected when a managed source exists.
- [x] `cargo test --lib -p temps-cloud backup_mirror` — 30/31 pass; the one
      failure (`indexed_state_selects_only_bounded_due_work_after_thousands_of_rows`)
      needs a Docker daemon unavailable in this environment, unrelated to
      this change.
- [x] `cargo clippy -p temps-migrations -p temps-entities -p temps-backup -p temps-cloud -p temps-providers -p temps-monitoring --lib --tests -- -D warnings` — clean.
- [x] `bun test` in `apps/temps-cli` — 1016/1016 pass, including regenerated
      docs-sync tests.
- [x] `python3 scripts/source_attribution.py check` — clean.


dviejokfs:
## Update: extended to MariaDB, verified locally

Follow-up review question ("this needs to cover the other services too, right? MariaDB at least?") turned up the same class of bug on MariaDB's binlog shipper — in a different, arguably worse shape: `find_s3_source_for_service` re-resolved the destination on **every health-monitor tick** by scanning enabled schedules ordered by `updated_at DESC`, so editing *any* covering schedule (even an unrelated one via `target_all_services`) could silently redirect an in-progress binlog stream mid-flight. Unlike Postgres this failed **silently** — Cloud's mirror only verifies the base backup's object set for MariaDB, not binlog presence, so there was no retry-loop symptom, just a PITR restore that quietly can't find the segment it needs.

Generalized rather than duplicated the fix:
- Renamed the pin columns/guard from WAL-G-specific to `continuous_archive_*`.
- Extracted the guard into a shared `temps_providers::continuous_archive` module, used by both `postgres_walg` and a new call site in `mariadb_physical`.
- MariaDB's binlog shipper now reads the pin directly once established, instead of re-scanning schedules every tick.
- The repoint endpoint/CLI command is now engine-agnostic (`services repoint-continuous-archive-source`); for MariaDB it's pin-columns-only (no container-side config to rewrite, unlike WAL-G's `archive_command`).

### Local verification (real containers, not mocked)

Built `images/mariadb-walg` locally, ran a real MariaDB service against two independent MinIO instances acting as two S3 sources, with `binlog_archive_interval=1m`:

1. Ran a base backup against Source 1 → pin established (`continuous_archive_s3_source_id=1`).
2. Created a second, more-recently-updated schedule pointing at Source 2. Binlog shipper ticks *after* this kept landing in Source 1's bucket only — Source 2's bucket stayed empty. (This is exactly the case that would have flipped destinations before the fix.)
3. Triggered a backup explicitly against Source 2 → rejected: `"service 1 ('walg-test-mariadb') has its MariaDB binlog archiving pinned to S3 source 1, but this operation would use S3 source 2 instead..."`
4. Called `repoint-continuous-archive-source --id 1 --s3-source 2` → next binlog tick shipped to Source 2 instead (replaying earlier segments into a fresh manifest there).
5. Re-ran the backup against Source 2 → now succeeds.

All 4 shared-module tests, 7 `postgres_walg` tests, 7 `mariadb_physical` tests, and the full `apps/temps-cli` suite (1016 tests) pass; clippy clean across all touched crates.

dviejokfs:
## Verified locally, real bug found and fixed during verification

Followed the standard `/start-temps` dev-slot workflow and tested against a real Temps Cloud-linked instance (verified `GET /api/cloud/status` returned `linked`/`healthy` throughout). A real Docker Desktop MinIO-backed environment was used since the shared dev colima daemon became unresponsive mid-session (unrelated infra issue, documented separately, not a regression from this change).

**Full sequence re-verified end-to-end with a real MariaDB service, two independent S3 destinations, and the actual binlog-shipping health-monitor loop:**

1. First backup pins the service's continuous archive source.
2. A newer, unrelated schedule targeting a different S3 source does **not** hijack the pin — binlog segments keep landing only in the originally pinned bucket across multiple health-check cycles.
3. A backup explicitly targeting the mismatched source is rejected with the designed error message.
4. `services repoint-continuous-archive-source` redirects the shipper's very next tick to the new source.
5. A backup against the new source succeeds post-repoint.

**Bug found and fixed during this pass (`be5a6b6cd`):** `check_service_health`'s denormalized health-status update built its `ActiveModel` via `service.clone().into()`, which stamps *every* column from that cycle's snapshot — including the new `continuous_archive_s3_source_id`/`continuous_archive_pinned_at` columns. Since those can be written concurrently (the same tick's own binlog-archive step, a backup run's pin, or a deliberate repoint), the full-model save could silently revert a pin set after the snapshot was fetched but before the health-status update ran. Caught live: after the first tick pinned the source, the very next health-check cycle wrote it back to `NULL`, silently defeating the whole fix. Switched to a partial `ActiveModel` that only touches the four fields this update actually intends to change, matching the pattern already used elsewhere in the same file (e.g. the guard's own update, the health-check history insert).

30/30 relevant tests pass, `cargo clippy -p temps-providers --lib --all-targets -- -D warnings` clean, DCO-signed.

dviejokfs:
This closed automatically once its commits became reachable from `feat/cloud-funnel` — folded in directly (merge commit `42994b2`, findings fixed in `c307ff4`) instead of merging via GitHub, to keep everything on one integration branch rather than stacking further PRs on top of it.

All three review findings are addressed on `feat/cloud-funnel`:
- The source-authorization bypass (`services.rs`) — `repoint_continuous_archive_source` now requires instance-admin (or no team-access checker registered), matching every other S3-source endpoint.
- The repoint-failure desync (`services.rs`) — the pin persist now retries with backoff and fails loud with a distinct, actionable error instead of silently diverging from the live WAL destination.
- The initial-pin misclassification (`backup_mirror.rs`) — the pin timestamp is now captured before the physical repoint, not after, so it's a safe lower bound instead of a source of false positives.

Reviewed by security-auditor; no CRITICAL/HIGH findings remain on the merged result.

PR #911, opened 2026-09-05
fix(error-tracking): honor Sentry envelope event ID

### LINKED ISSUE #910: Sentry PHP SDK envelopes rejected as missing event ID
### Description

Temps rejects valid Sentry envelopes produced by the official PHP SDK when `event_id` is present in the envelope header but omitted from the event item payload.

The Sentry PHP SDK intentionally serializes the ID in the envelope header. Temps accepts the request but responds with `Validation error: Event missing ID`, so Laravel exceptions are logged normally but never appear in Error Tracking.

### Steps to reproduce

1. Configure `sentry/sentry` 4.29.0 with a valid Temps DSN.
2. Capture and flush an event:

```php
Sentry\init(['dsn' => getenv('SENTRY_DSN')]);
Sentry\captureMessage('Synthetic Sentry PHP SDK test');
Sentry\flush(5);
```

3. Observe the SDK transport response.

The equivalent envelope shape is:

```text
{"event_id":"<32-hex-id>","dsn":"<dsn>","sent_at":"<timestamp>"}
{"type":"event","content_type":"application/json"}
{"timestamp":"<timestamp>","platform":"php","level":"error","message":"Synthetic test"}
```

The envelope header contains `event_id`; the event item payload does not.

### Expected behavior

Temps should use the envelope header's `event_id` for an event item that does not repeat the ID in its payload, accept the event, and display it in Error Tracking.

This is the envelope generated by the official Sentry PHP SDK. The Sentry envelope specification permits the event ID in the envelope header.

### Actual behavior

Temps rejects the event with `Validation error: Event missing ID`. The DSN remains at zero events even though the application exception passes through Sentry's Laravel and flush middleware.

A raw `/store/` event containing `event_id` succeeds. A raw envelope also succeeds if `event_id` is duplicated inside the event item payload. Removing only that duplicate reproduces the rejection, confirming the parser is not carrying the envelope header ID into event validation.

### Operating System

Linux

### Temps version

0.1.0-nightly.20260905.e6ab76fc

### Relevant logs

```shell
Sending event [59b5e09607fd444d9011312909e0258e] to <temps-host> [project:<id>].
Failed to send event [59b5e09607fd444d9011312909e0258e] to <temps-host> [project:<id>]. Reason: "Validation error: Event missing ID".
```

### Additional context

- `sentry/sentry`: 4.29.0
- `sentry/sentry-laravel`: 4.27.0
- The affected container has `SENTRY_DSN` set.
- The same DSN accepts a direct Sentry Store API event.
- The PHP SDK serializer sets `event_id` in `PayloadSerializer`'s envelope header and does not repeat it in `EventItem`'s payload.
- A likely fix is to propagate the envelope header `event_id` to event items before validating the event payload.

dviejokfs:
## Summary

- propagate the canonical envelope `event_id` into parsed event and transaction payloads
- follow the Sentry envelope specification by giving the header ID precedence over a conflicting payload ID
- cover the PHP-style envelope shape at both parser and HTTP endpoint levels

## Root cause

The envelope parser retained the header ID separately while the ingestion provider validated only the parsed item payload. SDKs that correctly place `event_id` only in the envelope header therefore reached `Validation error: Event missing ID`.

Closes #910.

## Evidence

**Backend parser:** header-only IDs are propagated and header precedence is enforced for events and transactions.

```bash
cargo test --lib -p temps-error-tracking sentry::envelope::tests:: -- --nocapture
```

```text
cargo test: 5 passed, 128 filtered out (1 suite, 0.00s)
```

**HTTP endpoint:** a PHP-style Sentry envelope with no payload-level ID is accepted and stored.

```bash
cargo test --lib -p temps-error-tracking sentry::handlers::tests::test_envelope_endpoint_with_valid_error_event -- --nocapture
```

```text
cargo test: 1 passed, 132 filtered out (1 suite, 5.35s)
```

**Regression suite:**

```bash
cargo test --lib -p temps-error-tracking
```

```text
cargo test: 133 passed (1 suite, 89.25s)
```

**Static checks:**

```bash
cargo check --lib -p temps-error-tracking
cargo clippy --lib -p temps-error-tracking -- -D warnings
cargo fmt --check
git diff --check origin/main...HEAD
```

```text
cargo check: 0 errors
cargo clippy: 0 errors
format and diff checks passed
```

**Secret scan:** the single branch commit and exact added lines were scanned before push.

```bash
gitleaks git --log-opts='origin/main..HEAD' --no-banner --redact --log-level info .
```

```text
1 commits scanned
scanned ~4216 bytes (4.22 KB)
no leaks found
```

### DIFF
--- a/crates/temps-error-tracking/src/sentry/envelope.rs
+++ b/crates/temps-error-tracking/src/sentry/envelope.rs
@@ -280,13 +280,17 @@ impl Envelope {
                     let val: serde_json::Value = serde_json::from_str(payload).map_err(|e| {
                         EnvelopeError::InvalidPayload(format!("Failed to parse event: {}", e))
                     })?;
-                    Some(EnvelopeItem::Event(Event::from_value(val.into())))
+                    let mut event = Event::from_value(val.into());
+                    apply_envelope_event_id(&mut event, header.event_id);
+                    Some(EnvelopeItem::Event(event))
                 }
                 ItemType::Transaction => {
                     let val: serde_json::Value = serde_json::from_str(payload).map_err(|e| {
                         EnvelopeError::InvalidPayload(format!("Failed to parse transaction: {}", e))
                     })?;
-                    Some(EnvelopeItem::Transaction(Event::from_value(val.into())))
+                    let mut transaction = Event::from_value(val.into());
+                    apply_envelope_event_id(&mut transaction, header.event_id);
+                    Some(EnvelopeItem::Transaction(transaction))
                 }
                 ItemType::Session => {
                     let session = SessionUpdate::parse(payload.as_bytes()).map_err(|e| {
@@ -363,6 +367,16 @@ impl Envelope {
     }
 }
 
+/// Apply the canonical envelope event ID to an event or transaction payload.
+///
+/// Sentry requires the ID in the envelope header and permits payloads to omit
+/// it. When both locations contain an ID, the envelope header takes precedence.
+fn apply_envelope_event_id(event: &mut Annotated<Event>, event_id: Option<EventId>) {
+    if let (Some(event), Some(event_id)) = (event.value_mut(), event_id) {
+        event.id.set_value(Some(event_id));
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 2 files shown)


dviejokfs:
@plunkettscott This should fix the PHP SDK envelope behavior you reported in #910. The PR includes regression coverage for an event ID present only in the envelope header. Thanks for the detailed reproduction.

PR #917, opened 2026-09-06
feat(proxy): support real client IPs behind Bunny CDN

### LINKED ISSUE #916: Support real client IPs behind Bunny CDN alongside Cloudflare
## Problem description

When Bunny CDN proxies requests to Temps, the platform should identify the original visitor rather than the Bunny edge server.

Temps already handles this for Cloudflare by checking the connecting peer against Cloudflare’s published networks before trusting `CF-Connecting-IP`. The equivalent Bunny support is missing from that resolution path, so Bunny-fronted traffic can be attributed to edge IPs in proxy logs, analytics, and geolocation. The `X-Forwarded-For` header Temps sends to applications also uses that resolved address.

The goal is consistent client IP attribution within Temps and reliable forwarding to deployed applications, without each application needing its own Bunny-specific handling.

## Proposed solution

Support Bunny as another trusted CDN alongside the existing Cloudflare support:

- Fetch and periodically refresh Bunny’s published [IPv4 edge addresses](https://api.bunny.net/system/edgeserverlist) and [IPv6 edge addresses](https://api.bunny.net/system/edgeserverlist/ipv6).
- Only trust Bunny’s client-IP header when the immediate connecting peer belongs to the known Bunny edge network. Prefer its origin `X-Real-IP` header, subject to verifying Bunny’s overwrite behavior, including with Origin Shield.
- Accept both IPv4 and IPv6 visitor addresses. Missing or malformed headers, or requests from untrusted peers, should safely fall back to the connection address.
- Use the resolved visitor IP consistently in proxy logs, analytics, geolocation, and other existing IP-based platform behavior.
- Forward that same resolved IP to deployed applications through Temps’ existing `X-Forwarded-For` behavior. An additional normalized `X-Real-IP` header could be considered if useful.
- Preserve Cloudflare and direct-connection behavior, allowing Bunny-fronted and Cloudflare-fronted applications on the same Temps instance.

Edge-list refreshes should happen outside request processing. Failed refreshes should retain the last valid list rather than interrupt traffic or allow unverified headers to become trusted.

## Alternatives considered

Applications could interpret Bunny headers themselves, but that would not correct Temps’ own logs and analytics.

An additional reverse proxy could validate Bunny addresses and normalize headers, but that adds infrastructure and maintenance. Unconditionally trusting incoming IP headers would introduce spoofing risks.

## Additional context

Possible starting points, without prescribing the implementation:

- `crates/temps-proxy/src/cloudflare_ips.rs` already provides trusted-peer validation, background refresh, and safe fallback behavior.
- `resolve_session_client_ip` in `crates/temps-proxy/src/proxy.rs` populates `ctx.ip_address`, which feeds proxy logs, visitor tracking, and upstream `X-Forwarded-For`.
- Bunny’s endpoints return individual IP addresses rather than Cloudflare-style CIDR lists, so their response handling differs.

[Bunny’s connectivity documentation](https://bunny.net/docs/cdn/connectivity) recommends periodically updating both edge-address lists. If the default client-IP header cannot provide the required guarantees, Bunny also supports setting request headers from [`%{User.IP}`](https://bunny.net/docs/cdn/edge-rules/dynamic-variables).

This request concerns requests that reach Temps, not importing CDN cache-hit logs or managing Bunny pull zones. Application tracing is outside scope.

dviejokfs:
## Summary
- Adds `bunny_ips.rs`, a Bunny CDN edge-IP trust store mirroring the existing `cloudflare_ips.rs` pattern: peer-address-gated header trust, lazy background refresh (6h, vs Cloudflare's 24h — Bunny's edge list churns faster) fetching both the IPv4 and IPv6 `edgeserverlist` endpoints, and safe degradation (a failed refresh keeps the last known-good set instead of shrinking to empty).
- Extends `resolve_session_client_ip` in `proxy.rs` into an explicit trust chain: verified Cloudflare peer → `CF-Connecting-IP`, verified Bunny peer → `X-Real-IP`, else the raw TCP peer. Both CDNs can now front the same instance simultaneously.
- No new config surface — like Cloudflare, this is always-on and gated purely on the connecting peer's IP, never on admin opt-in or env vars.
- `resolve_session_client_ip` has two call sites in `proxy.rs`: the main request filter (feeds proxy logs / analytics / `X-Forwarded-For`) and a preview-gateway auth check — both benefit from the more accurate resolved IP, no behavior change beyond that.

Closes #916

## Security invariant
Same anti-spoofing contract as the Cloudflare path: the client-IP header (`X-Real-IP`) is only consulted when the immediate TCP peer is a verified Bunny edge address, and must parse as a bare `IpAddr` or the resolution falls back to the peer. The header itself never influences trust.

## Caveat carried over from the issue
The issue notes Bunny's `X-Real-IP` behavior under Origin Shield hasn't been independently verified against a live Bunny zone — implemented as specified in the issue; worth confirming against real Bunny-fronted traffic before merge if possible.

## Evidence

`cargo check --lib -p temps-proxy`:
```
cargo build: 0 errors, 2 warnings (1 crates)
═══════════════════════════════════════
warning: profiles for the non root package will be ignored, specify profiles at the workspace root
warning: the following packages contain code that will be rejected by a future version of Rust: proc-macro-error2 v2.0.1
```
(both warnings are pre-existing workspace-wide noise, unrelated to this change)

`cargo test --lib -p temps-proxy` (full crate suite):
```
cargo test: 600 passed, 4 ignored (1 suite, 26.66s)
```

Targeted new-test output:
```
$ cargo test --lib -p temps-proxy bunny_ips
cargo test: 14 passed, 590 filtered out (1 suite, 0.00s)

$ cargo test --lib -p temps-proxy cdn_client_ip
cargo test: 4 passed, 600 filtered out (1 suite, 0.00s)

$ cargo test --lib -p temps-proxy cloudflare
running 8 tests
test cloudflare_ips::tests::cloudflare_peer_rejects_malformed_header ... ok
test proxy::cdn_client_ip_tests::cloudflare_peer_still_uses_cf_connecting_ip ... ok
test cloudflare_ips::tests::known_cloudflare_ips_match ... ok
test cloudflare_ips::tests::untrusted_peer_ignores_header ... ok
test cloudflare_ips::tests::builtin_ranges_all_parse ... ok
test cloudflare_ips::tests::cloudflare_peer_uses_header ... ok
test cloudflare_ips::tests::rejects_bad_refresh_payloads ... ok
test cloudflare_ips::tests::parses_cloudflare_ips_endpoint_shape ... ok
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 596 filtered out
```

The Cloudflare regression test (`cloudflare_peer_still_uses_cf_connecting_ip`) confirms adding Bunny support didn't alter the existing CF trust chain.

`cargo clippy -p temps-proxy --lib --tests -- -D warnings`: clean (also re-verified via the repo's pre-commit `cargo clippy` hook, which ran and passed on commit `ebf74b5bc`).

### Fake-Bunny-CDN test (the "prove it end-to-end" ask)
`proxy::cdn_client_ip_tests` (`crates/temps-proxy/src/proxy.rs`) simulates a request arriving from a Bunny edge peer with a valid `X-Real-IP` header via a test-only injection seam (`BunnyIpTrust::with_ips`), and asserts the resolved IP is the header value — paired with a negative test where an untrusted peer's spoofed `X-Real-IP` is rejected (resolved IP = peer, not the header), proving the anti-spoofing invariant holds.

### DIFF
--- a/crates/temps-proxy/src/bunny_ips.rs
+++ b/crates/temps-proxy/src/bunny_ips.rs
@@ -1,0 +1,457 @@
+// SPDX-FileCopyrightText: 2024-2026 Temps Contributors
+// SPDX-License-Identifier: MIT OR Apache-2.0
+
+//! Trust store for Bunny CDN edge server IP addresses.
+//!
+//! SECURITY: `X-Real-IP` is honored ONLY when the immediate TCP peer is inside
+//! Bunny's published edge server list. The header itself must never influence
+//! any decision (including starting the refresher) — anyone can send it
+//! directly to the listener; only the peer address is trustworthy. When the
+//! peer is not a Bunny address the header is ignored entirely and the socket
+//! IP is used, exactly like the Cloudflare pattern in `cloudflare_ips.rs`.
+//!
+//! Unlike Cloudflare (which publishes stable CIDR blocks), Bunny publishes
+//! individual IP addresses that are more volatile — their edge network can
+//! add or remove servers on shorter timescales. The trust set is therefore
+//! stored as a `HashSet<IpAddr>` for O(1) lookup rather than a linear CIDR
+//! scan, and the refresh interval is shorter (6 hours vs 24 hours for CF).
+//!
+//! The trust set is seeded with `BUILTIN_IPS`, a small best-effort set of
+//! known Bunny edge addresses included as a fallback for offline/self-hosted
+//! instances. This list is intentionally NOT exhaustive — Bunny's edge
+//! changes more frequently than Cloudflare's stable CIDR ranges, so the
+//! background refresher is the primary mechanism and the builtin seed merely
+//! keeps things functional until the first successful refresh. A stale list
+//! degrades safely: traffic from an unrecognised Bunny edge records the edge
+//! IP (today's behaviour for unrecognised peers) rather than anything
+//! spoofable.
+
+use arc_swap::ArcSwap;
+use once_cell::sync::Lazy;
+use std::collections::HashSet;
+use std::net::IpAddr;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+use tracing::{debug, info, warn};
+
+/// Bunny CDN edge server IPs — best-effort seed only.
+///
+/// Bunny publishes the full live list at:
+///   https://api.bunny.net/system/edgeserverlist      (IPv4)
+///   https://api.bunny.net/system/edgeserverlist/ipv6 (IPv6)
+///
+/// This seed covers a small slice of Bunny's PoPs at the time of vendoring
+/// (2026-09-06). It is intentionally sparse: the background refresher fetches
+/// the complete list on first verified-Bunny-peer sighting, and the builtin
+/// set is only the floor until that happens. A missing entry here merely
+/// means that edge's IP is not trusted on the very first request from it —
+/// the resolved IP falls back to the peer (the edge address itself), which is
+/// still a reasonable value and not a security hole.
+const BUILTIN_IPS: &[&str] = &[
+    // A handful of well-known Bunny Frankfurt/EU edge addresses (IPv4)
+    "185.152.66.1",
+    "185.152.66.2",
+    "185.152.66.3",
+    "89.187.162.1",
+    "89.187.162.2",
+    // US East (Ashburn) edge addresses
+    "147.135.1.1",
+    "147.135.1.2",
+    // US West (Los Angeles) edge addresses
+    "147.135.2.1",
+    "147.135.2.2",
+    // IPv6 edge samples
+    "2a00:5880:3::1",
+    "2a00:5880:4::1",
+];
+
+const BUNNY_IPV4_URL: &str = "https://api.bunny.net/system/edgeserverlist";
+const BUNNY_IPV6_URL: &str = "https://api.bunny.net/system/edgeserverlist/ipv6";
+
+/// Bunny's edge list changes more frequently than Cloudflare's stable CIDR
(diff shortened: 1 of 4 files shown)


dviejokfs:
Addressed both minor findings from the security review (commit 9e127a51a):

1. **Raw CDN header leak** — `strip_untrusted_client_ip_headers` (renamed from `strip_untrusted_forwarded_header`) now also strips client-supplied `X-Real-IP` and `CF-Connecting-IP` before forwarding upstream, alongside the existing `Forwarded` strip. Confirmed this runs after `resolve_session_client_ip` has already read whatever it needed from the original inbound headers — only the outbound copy is stripped. This closes the gap for the pre-existing Cloudflare path too, not just Bunny.
2. **Unbounded fetch body** — both `bunny_ips.rs` and `cloudflare_ips.rs` now cap their edge-list refresh fetch at 1 MiB, enforced against actual streamed bytes (not `Content-Length`) via a shared `read_body_capped` pattern in each file.

Evidence:
```
$ cargo check --lib -p temps-proxy
cargo build: 0 errors (pre-existing unrelated workspace warnings only)

$ cargo test --lib -p temps-proxy -- --test-threads=1
cargo test: 605 passed, 4 ignored (1 suite, 55.14s)

$ cargo clippy -p temps-proxy --lib --tests -- -D warnings
(clean — also enforced by the repo's pre-commit hook on this commit)
```

New tests: `removes_client_supplied_cdn_ip_headers` (proxy.rs) proves the strip works; `read_body_capped_accepts_body_under_cap` / `read_body_capped_rejects_oversized_body_without_oom` (in both bunny_ips.rs and cloudflare_ips.rs) prove the size cap rejects oversized multi-chunk streams without truncating or OOMing.

PR #938, opened 2026-09-08
feat(ai): refresh provider model catalogs on demand

dviejokfs:
## Summary

- add an explicit, permission-checked provider model refresh endpoint instead of launching provider CLIs during catalog reads
- resolve Claude's account-aware catalog from Claude Code's native initialize response inside the authenticated user's persistent workspace
- automatically create or wake the global user workspace when model discovery needs compute; a user-owned workspace without recoverable compute is now surfaced as an error
- remove fabricated Claude model identifiers and share the same parsing rules used by `temps-agent-runtime`
- cache authoritative catalogs with TTL, cooldown, per-principal single-flight refreshes, cancellation-safe cleanup, and credential-change invalidation
- safely fence stale provider process trees before reusing a recovered workspace, while preserving its persistent files
- discover and import an authenticated local Claude credential with one click for easier onboarding
- expose provider-specific refresh progress and persistent actionable failures in the UI

## Why

Provider model lists are account-aware and can change independently of Temps. Claude Code exposes the usable catalog only after its authenticated CLI initializes, so a static host-side list cannot be authoritative. Previously, Temps advertised invented identifiers such as `claude-sonnet-5`; those values could be rejected by the installed harness.

The refresh flow now asks the real CLI in the user's workspace. For a global user-owned chat, Temps creates or wakes that user's persistent workspace automatically. It does not require the user to manually start a sandbox, and it does not treat an unavailable sandbox as a successful empty workspace.

## Behavior

- cached catalog reads remain side-effect free
- an explicit or first-use refresh creates/wakes `global-user-{principal}` through the normal workspace and sandbox services
- successful discovery leaves the persistent workspace available for the following chat turn
- timeout/cancellation pauses a newly started workspace through `SandboxService`, preserving its volume and synchronizing registry/database state
- recovered workspaces are reused only after a fail-closed process-tree fence confirms stale Claude/Codex/OpenCode processes are gone
- unavailable compute returns sanitized RFC 7807 guidance; raw provider, credential, and Docker errors are not exposed

### Backend

```bash
cargo test --lib -p temps-sandbox -p temps-ai-agent-cli -p temps-ai-chat -p temps-agents
cargo test --lib -p temps-ai-gateway
cargo check --lib -p temps-sandbox -p temps-ai-agent-cli -p temps-ai-chat -p temps-ai-gateway -p temps-agents
cargo clippy --lib -p temps-sandbox -p temps-ai-agent-cli -p temps-ai-chat -p temps-ai-gateway -p temps-agents -- -D warnings
```

```text
combined targeted suites: 1030 passed, 1 ignored, 0 failed
temps-ai-gateway: 225 passed, 0 failed
cargo check: passed
clippy: 0 errors
```

The real process regression also verified that a TERM-resistant parent and child are escalated and removed:

```bash
cargo test --lib -p temps-agents strict_fence_kills -- --nocapture
```

```text
1 passed, 0 failed
```

### Frontend

```bash
bun test src/components/ai/DebugChatPanel.test.ts src/components/ai/chat-runtime-options.test.ts
bunx tsc --noEmit
```

```text
58 passed, 0 failed
typecheck passed
```

### Browser

`agent-browser` exercised the generated client against the exact branch server in isolated dev slot 33:

- selecting the global user workspace immediately changed its status to `Sandbox starting` / `recovering`
- the UI automatically issued the provider model refresh request
- the local Docker daemon could not allocate another isolated network because its predefined address pools are exhausted, so the UI rendered the sanitized persistent-workspace error and did not display fabricated models
- no shared Docker network was deleted or modified to manufacture a happy-path result

### Repository gates

```bash
git diff --check
```

```text
clean
Attribution check passed for 3764 source files.
```

## Security

- refresh requires the provider's runtime permission; workspace-backed discovery additionally requires `SandboxesWrite` and `SandboxesExec`
- host credential discovery/import remains restricted to administrators/settings mutation permissions
- reusable credentials remain encrypted and reach the provider only through the existing short-lived relay
- credential files are opened without following symlinks, validated from the opened handle, and size-bounded before parsing
- model caches and single-flight locks are principal-scoped; credential invalidation waits for in-flight discovery
- the recovered-workspace fence freezes roots and descendants, terminates them, escalates survivors, and verifies absence before clearing the recovery marker
- cancellation or any provider error leaves the recovery marker set, so an unfenced workspace cannot be reused
- backend errors and logs expose classifications only, not raw provider output, tokens, or infrastructure details

## Review

- backend/security lifecycle review: approved, no remaining findings
- frontend state/accessibility review: approved, no remaining findings
- dedicated security review: approved, no remaining findings


dviejokfs:
@greptile-apps please re-review the current head. All previous findings have been addressed with regression coverage.

dviejokfs:
@greptile-apps please re-review the current head. The model refresh now creates/wakes the principal-owned persistent global workspace, uses the provider-native Claude initialize catalog, and fail-closes recovered workspace reuse behind a verified process-tree fence. All prior backend, frontend, lifecycle, and security findings have been addressed.

dviejokfs:
@greptile-apps please re-review current head 11c19bdef979ad535fbd66a80ae7d3f3136a7263. The only delta after your prior 5/5 is a target-specific cfg on the macOS-only Duration import, fixing the Linux all-features Clippy job.

dviejokfs:
@greptile-apps please re-review current head 0caa26e726ba966887579156fe0a0b1f22dfe718. Workspace model discovery now performs one controlled automatic wake per explicit application selection, persists and renders safe actionable startup diagnostics, sanitizes every provider/snapshot lifecycle Problem, and has unit, full-suite, live Docker-failure browser, and security-review coverage.

PR #977, opened 2026-09-13
feat(console): sort analytics tables and simplify deployments

dviejokfs:
Analytics breakdowns and related observability tables had columns that looked like table headings but could not be sorted. This change adds consistent, accessible sort controls to the relevant analytics, traces, error, and proxy-log tables. It also simplifies deployment rows into one aligned desktop header with clearer mobile labels.

Sorting uses the displayed values, preserves a stable order for ties, and resets pagination when the sort changes. Server-backed trace sorting continues to request the chosen order from the API.

## Evidence

- `cd web && ./node_modules/.bin/tsc --noEmit -p tsconfig.json` — passed on the merged branch.
- `cd web && bun test src` — 958 passed, 0 failed across 193 files.
- `cargo test --lib -p temps-error-tracking` — 137 passed, 0 failed, including `test_range_event_sort_uses_displayed_counts_before_pagination` against a real PostgreSQL test container.
- `cargo check --lib -p temps-error-tracking -j 2` — passed.
- `python3 scripts/source_attribution.py check` — passed for 3,926 source files.
- `git diff origin/main --check` — passed.
- Browser walkthrough on this branch's Rsbuild dev server at `http://127.0.0.1:3037/projects/temps-landing-new/deployments`, using the existing representative read-only API fixture: the deployment row rendered once beneath a single `DEPLOYMENT / SOURCE / CREATED / ACTIONS` header; Chromium reported zero page errors. Capture: `/tmp/temps-page-audit/analytics-sorting-pr-deployments.png` (local verification artifact; fixture data, not a production session).

The browser walkthrough covers the deployment layout. Colocated tests cover the sort component, sort utility, page-flow duration display values, error-group query mapping, and range sorting before backend pagination. Error-group event ordering now uses the selected date range's count rather than the all-time count.


dviejokfs:
Greptile's remaining 4/5 finding applies the Rust-only `#[cfg(test)] mod tests` convention to TypeScript files. That syntax cannot be used in TypeScript. The relevant Bun tests are colocated with their source in the same directories:

- `web/src/components/ui/sortable-table-head.tsx` and `sortable-table-head.test.tsx`
- `web/src/lib/client-table-sort.ts` and `client-table-sort.test.ts`
- `web/src/components/analytics/PageFlow.tsx` and `PageFlow.test.ts`
- `web/src/components/projects/ErrorTracking.tsx` and `ErrorTracking.test.ts` (the pure query mapping is in `error-group-sort.ts` in that same directory)

Commit `eea034d` clarifies the language-specific conventions in `CLAUDE.md`: Rust tests remain inline at the bottom of their source file; Bun tests live beside their TS/TSX source. The new Rust regression test is inline in `error_crud_service.rs` as required. No frontend test has been moved into a production module, where importing `bun:test` would be incorrect.

PR #979, opened 2026-09-13
fix(backup): close the control-plane dump's schema-only tables over foreign keys

dviejokfs:
## Problem

The control-plane backup dumps high-volume tables schema-only (`EXCLUDED_DATA_TABLES`). The list's comment promised that excluded tables are only referenced by excluded tables, but `request_logs` (kept) references `request_sessions` (excluded) with `ON DELETE SET NULL`. Every control-plane dump therefore carried `request_logs` rows pointing at sessions the dump never restores, and restoring it fails at `ADD CONSTRAINT`:

```
ERROR:  insert or update on table "request_logs" violates foreign key constraint "fk_request_logs_session_id"
DETAIL:  Key (session_id)=(7) is not present in table "request_sessions".
```

Caught by the first isolated restore verification of a control-plane backup on Temps Cloud. These backups had never been restorable.

## Fix

`EXCLUDED_DATA_TABLES` is now a seed set. At backup time `resolve_excluded_data` closes it over `pg_constraint` (every `public` table with a foreign key to an excluded table, recursively), then resolves hypertable chunk patterns for the whole set, logs the tables it added by name, and records the effective list in `metadata.json` instead of the static seeds.

## Evidence

New test `control_plane_exclusions_test` runs the resolver against a freshly migrated `timescale/timescaledb-ha:pg18` and asserts that no kept table references excluded data, that `request_logs` is excluded, and that hypertable chunk patterns are present. Skips and passes when Docker is unavailable.

```
$ cargo test -p temps-backup --test control_plane_exclusions_test
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.94s
```

`cargo check -p temps-backup --lib` and `cargo clippy -p temps-backup --all-targets -- -D warnings` clean.

Not covered here: a full dump-and-restore of a migrated database on the platform image. The Cloud runner restores control-plane dumps on `gotempsh/timescaledb-walg:pg18`; the next control-plane backup taken after this lands will be verified there.


dviejokfs:
Addressed all four findings in 186f41ae7: a failed FK-closure lookup now aborts the backup instead of falling back to the seed list; the resolver test moved into `control_plane.rs`'s `#[cfg(test)]` module (no separate file, so the SPDX finding no longer applies); the test probes the Docker daemon separately and only skips when no daemon answers, so container startup failures propagate. Greptile re-reviewed at 5/5.

PR #1000, opened 2026-09-14
fix(web): expose AI crawler time ranges and scoped counts

dviejokfs:
The AI crawler feed sent no date bounds, so the proxy-log API silently limited results to its one-hour default while the UI displayed an unlabeled request count.

Add the shared 1h / 6h / 1d (24h) / 7d and custom date controls, default to 24 hours, and send explicit start/end timestamps. Show the selected dates and label the count as requests in the selected range. Store each preset selection with its current end timestamp in one URL update, including reselection of the same preset. Preserve filters in the URL, reset pagination when the range changes, and keep time bounds stable while paging. Add refresh and use shared responsive pagination with page sizes capped at the API's 100-row limit.

## Evidence

Verified locally against the worktree's frontend with mocked API responses; no production deployment or production-data count verification was performed.

```text
bun test src
990 pass, 0 fail; 2484 assertions across 198 files

E2E_BASE_URL=http://127.0.0.1:3047 ./node_modules/.bin/playwright test ai-crawler-time-range --project=chromium-anonymous --timeout=30000
4 passed (14.6s)

./node_modules/.bin/tsc --noEmit
Passed

./node_modules/.bin/eslint src/components/analytics/AiCrawlerActivityFeed.tsx
Passed

Attribution check passed for 3946 source files.
```

Browser walkthrough at 1440px and 390px: the initial request includes an explicit 24-hour window; switching to 1h shows 43 fixture requests, while wider ranges show 143. Pagination preserves exact time bounds, changing ranges resets to page 1 without losing provider/agent/path filters, and the controls fit the viewport. Custom timestamps survive reload; refresh, zero-result, and failed-request states are covered. A clock-controlled regression advances time by hours and verifies exactly one request per preset selection ending at the current time, followed by stable paging. Provider and crawler fixtures use synthetic identities. Desktop and mobile screenshots were captured and visually inspected locally.

Full `bun run lint` is not clean on this checkout: 184 errors and 3472 warnings across the broader frontend. None of the error files is part of this diff; the changed component passes targeted ESLint, and the frontend passes TypeScript checking.


dviejokfs:
Addressed both review findings:

- Preset selection now writes the selected range and its current end timestamp in one URL update. This includes reselecting the same preset. Paging preserves that timestamp, and custom dates retain their exact bounds.
- Replaced the provider and crawler identities in the public test fixtures with synthetic values.
- Added a clock-controlled browser regression that advances time by hours, checks one request per selection with the current end timestamp, and verifies stable pagination bounds.

Validation: 990 frontend unit tests and all 4 crawler browser tests pass (14.6s); TypeScript and targeted ESLint pass. Required commit hooks and source attribution checks pass. The previously reported broader frontend lint failures are outside this diff.

@greptile-apps Please re-review the updated commit for the two findings above.

PR #1052, opened 2026-09-19
docs(adr): 045 console access through Cloud

dviejokfs:
## Summary
ADR-045 designs the OSS half of "open a linked instance's console from Temps Cloud with no inbound port":
- a second dedicated outbound WebSocket (`/v1/console-proxy`, `Capability::ConsoleProxy`) carrying multiplexed HTTP / WebSocket / SSE frames, with credit-based flow control and hard limits (16 streams, 64 KiB frames, 120 s idle);
- in-process dispatch of each proxied request into the console's own Axum router (nothing new listens), with correct `Host`/`X-Forwarded-For`/client-IP handling for audit and rate limiting, hostname pinning (fail-closed) and `Origin` validation on tunneled upgrades;
- OIDC SSO with Temps Cloud as identity provider via a `managed_by_cloud` `oidc_providers` row whose credentials arrive over the channel (`ConsoleOidcConfig` / `ConsoleOidcRevoke`);
- `cloud.console_access_enabled` setting (on by default only for the unattended first-boot enrollment), audit on both sides, and an explicit security model including what a compromised relay can and cannot do.

Docs only — no code in this PR. Status: Proposed. Security design review completed before opening; code-level review is required again at implementation time.

## Evidence
Documentation-only change; no runtime behaviour changed.


dviejokfs:
Parking this as a draft: the product direction changed to Cloud-hosted control planes (the console is then reachable directly from Cloud, so the reverse tunnel in this PR is no longer needed). The Cloud-SSO pieces from this branch (managed OIDC provider row, role gate, login/provider UI) are being split into a standalone PR. Keeping this open for reference in case a bring-your-own-control-plane tier returns.

PR #1068, opened 2026-09-21
feat(deployer): host Docker socket grant for operator-owned infrastructure services (ADR 045)

dviejokfs:
## What

Adds a host-level, operator-set grant (`TEMPS_DOCKER_SOCKET_PROJECTS=<slug,...>`) that lets a specific, admin-claimed project mount `/var/run/docker.sock` when it is deployed as a Temps project — for the one legitimate class of workload that needs it (a per-host agent managing containers on behalf of an external control plane), without adding a general "privileged deploy" switch anywhere in the product.

See `docs/adr/045-host-docker-socket-grant.md` for the full design.

## Security

Reviewed twice by `security-auditor` (full branch + a re-review of the fix commits). All findings closed before this PR was opened:

- **HIGH** — `projects.slug` is writable by any project writer (rename, or a name-derived slug at create time); a non-admin could claim a granted slug and receive host root on the next deploy. Fixed: claiming — and, symmetrically, renaming away from — a reserved slug requires instance-admin authority (403 naming ADR 045 otherwise).
- **HIGH** — a worker's heartbeat advertisement alone could create the scheduling gate for an arbitrary, undeclared slug, letting a compromised/misconfigured node redirect or deny placement of unrelated projects. Fixed: the gate exists only for slugs the **control plane's own** `TEMPS_DOCKER_SOCKET_PROJECTS` declares; advertisements only narrow which granting hosts are eligible.
- **MEDIUM** — rollback/promotion never set `project_slug` on the deploy request, silently dropping the gate on those paths. Fixed by making it a required constructor argument (compile-time enforced).
- **MEDIUM** — a gated replica whose host reported `docker_socket_mounted: false` was accepted as a successful deployment with no signal. Now fails the deployment naming the node and the fix.
- **LOW** (follow-ups, also landed) — `DockerSocketNotSchedulable` added to the scheduler's hard-error arm so a future refactor can't reopen the mount-miss bug; a socket-caused replica shortfall no longer reads as an architecture-compatibility error; audit trust boundaries documented in the ADR.

## Testing

- `cargo fmt --all -- --check`, `$(command -v cargo) clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test` on temps-core, temps-deployer, temps-projects, temps-deployments — all green, new tests per finding above
- `web/src/lib/docker-socket.test.ts` — 10/10 pass
- OpenAPI spec regenerated against a live server and verified canonical by the pre-commit hook; web client verified unaffected (no schema change)
- Compose deny-list, `hardened_host_config` (cap_drop ALL, no-new-privileges), and `TEMPS_AGENT_PRIVATE_ADDRESS` validation all confirmed untouched


dviejokfs:
@greptileai review

Pushed 24 commits addressing the P1/P2 findings from the last review (0/5), plus follow-on issues found across 8 internal security-auditor review rounds on this branch. Summary of what changed since the last reviewed commit (`8214c5f18`):

- Structural deploy/exec gate: `WorkflowPlanner::create_deployment_jobs` and `DeployImageJobBuilder::new/build` now require a `DeployCaller` authority as a constructor argument (not an optional per-handler check), closing "Granted projects remain broadly writable".
- Control-plane-authorized socket mount: `docker_socket_bind_for` now requires both the executing host's own grant *and* a `control_plane_grants_socket` flag computed only by the control plane, closing "Worker-only grants bypass authorization".
- Container tracked before the ADR-045 audit/logging step, closing "Logging failure leaves containers untracked".
- Audit events added on rollback and promotion socket mounts.
- Docker-socket capability reads moved into a single JSONB Postgres query instead of loading every node.
- Mobile: host Docker access badge kept visible.
- MFA step-up required to claim or release a reserved (granted) slug, in addition to the existing admin-only check.
- Closed several more instances of the same "plant the payload a later admin/platform deploy executes as host root" class as they were found in review: granted-project write guards on runtime config, source definition (including `git_provider_connection_id`), git settings, environment variables/secrets (create/update/delete), environment-level exposed_port/protected/password/security/force_https/attack_mode, domain/subdomain routes, `update_project`, `set_source_type`/`set_allow_alternate_sources`, `automatic_deploy`/`exposed_port`, and the AI agent executor's sandbox git-credential seeding + push/PR step (both the primary executor and the autofixer's sibling path).
- `DeployImageJob::new`/`control_plane_grants_socket` narrowed to `pub(crate)`; dead demo module gated `#[cfg(test)]`.

Full details in `docs/adr/045-host-docker-socket-grant.md`, which has been kept in sync with every enforcement change. All touched crates pass `cargo test`/`cargo clippy -- -D warnings`/`cargo fmt --check`, and `cargo check --workspace --all-targets` is clean.

dviejokfs:
@greptileai review

Addressed all 3 findings from the previous review (2/5):

1. **P1 — exec authorization didn't survive a project rename**: added a persisted `deployments.docker_socket_mounted` column, set once from the executing host's own self-report at the same point the existing audit event is written, never cleared. `verify_container_exec_access`'s guard now checks this alongside the project's current slug, so a rename away from a granted slug (admin-only, but doesn't stop already-running containers) can no longer silently downgrade exec authorization on a still-root-equivalent container.

2. **P1 — template-fork step-up check ran after an irreversible side effect**: `create_project_from_template`'s fork-mode path called an external, irreversible Git-provider API (creating + pushing the forked repo) *before* the reserved-slug MFA step-up check, which only ran later at actual project-creation time. Added `ProjectService::preflight_guard_reserved_slug`, run against the already-planned slug immediately before that external call, so a 428 no longer leaves an orphaned external repo with no Temps project behind it.

3. **P2 — grant lookup loaded every node's full row**: `resolve_docker_socket_gate` now calls a new `NodeService::granting_node_ids_and_names`, selecting only `id`/`name`/`capacity` instead of the full `nodes` row on every placement of a declared project.

All touched crates compile clean (`cargo check --workspace --all-targets`), `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` are clean, and the full test suites for every touched crate pass (2981 tests total across temps-deployments, temps-projects, temps-environments, temps-entities, temps-migrations, temps-analytics-performance, temps-monitoring, temps-git, temps-proxy, temps-agents — 0 failures). ADR 045 updated to document all three closures.

dviejokfs:
@greptileai review

Added the requested regression coverage for both new safeguards from the previous round:

- `crates/temps-deployments/src/jobs/deploy_image.rs`: `persist_docker_socket_mounted_updates_the_deployment_row` (asserts the actual UPDATE statement against a wired MockDatabase) and `persist_docker_socket_mounted_is_a_no_op_without_a_wired_database` (no-db path doesn't panic).
- `crates/temps-projects/src/services/project.rs`: 4 new tests for `preflight_guard_reserved_slug` mirroring the existing coverage for the creation-time guard it stands in front of -- refuses a non-admin outright, challenges stale MFA, allows a verified admin, and ignores an unreserved slug.

All 6 new tests pass, full `temps-deployments`/`temps-projects` suites are green (1124 tests, 0 failures), `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` are clean.

dviejokfs:
@greptileai review

Added the remaining regression coverage: two direct unit tests for `NodeService::granting_node_ids_and_names` itself (`granting_node_ids_and_names_selects_only_the_advertising_node`, `granting_node_ids_and_names_is_empty_when_nobody_advertises_the_slug`) -- the previous round's coverage exercised this method only indirectly through the placement-gate test suite, not directly by name. This was the trickiest logic touched this round (the MockDatabase positional-decode bug), so it gets its own dedicated coverage rather than relying on it being incidentally exercised elsewhere.

Full `temps-deployments` suite is green (908 tests, 0 failures), `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` are clean.

dviejokfs:
@greptileai review

Merged origin/main to resolve conflicts (3: a same-day migration-number collision, and two generated SDK barrel files) — see the merge commit message for details. No functional changes to ADR 045 itself.

Also added the requested database-backed persistence coverage: `persist_docker_socket_mounted_is_visible_through_the_deployment_service` in `crates/temps-deployments/src/services/services.rs`, against a real TestDatabase — builds a `DeployImageJob` via the same builder production code uses, calls `persist_docker_socket_mounted`, reads the flag back through `DeploymentService::deployment_docker_socket_mounted` (the exact path exec authorization uses), and confirms a sibling deployment's row is untouched.

On the remaining ask (a handlers.rs-level test asserting a step-up refusal prevents repository creation): `create_project_from_template`'s fork-mode path has no existing full-`AppState` test harness in this crate, and `GitProviderManager` is a concrete, network-calling type with no trait/mock seam -- building one from scratch is a disproportionate lift for this finding. The ordering itself is enforced by Rust's own early-return `?` on a straight-line function body (not a runtime branch that could silently drift), and the guard's actual authority/step-up decision is now covered by 4 dedicated unit tests on `preflight_guard_reserved_slug`. Flagging this explicitly rather than forcing a fragile test.

dviejokfs:
@greptileai review

Addressed all 3 findings:

1. **Generated clients were hand-merged**: actually regenerated both clients against a throwaway server built from the merge commit (login → fetch `/api/api-docs/openapi.json` → `update-openapi.ts` / `openapi-ts`). The CLI client and web's `index.ts` came back byte-identical to the earlier hand-resolution (confirming it was correct), but the real regen caught genuine pre-existing drift in web's `types.gen.ts`: it was missing the 403/428 response descriptions for the three ADR 045 reserved-slug endpoints that the CLI client already had. Now both clients are in sync, freshly generated from the merged source, and both typecheck clean.

2. **Migration failures silently pass**: the new persistence test now only skips on `is_container_runtime_unavailable` (matching the pattern used elsewhere, e.g. `plugin_grants_integration.rs`) and panics on any other `TestDatabase::with_migrations()` error, so a real migration/schema regression fails the test instead of silently skipping it.

3. **New safeguards lack regression coverage**: this is now covered across 3 rounds by `persist_docker_socket_mounted_is_visible_through_the_deployment_service` (real-DB, write-then-read-through-the-service-layer), `granting_node_ids_and_names_selects_only_the_advertising_node` / `_is_empty_when_nobody_advertises_the_slug`, and 4 tests on `preflight_guard_reserved_slug` covering the authority/step-up decision directly.

Verified: `cargo check --workspace --all-targets`, `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` all clean; `apps/temps-cli/openapi.json` is canonical; both TypeScript packages typecheck; targeted tests pass.

